PR #363, opened 2015-01-02
A class Mappable that makes it possible to apply operator() to Pair.

pleroy:


base/mappable.hpp
```diff
@@ -0,0 +1,28 @@
+#pragma once
+
+#include <functional>
```
eggrobin: d2d

base/mappable.hpp
```diff
@@ -0,0 +1,28 @@
...
+#include <utility>
+
+namespace principia {
+namespace base {
+
+// This class helps in declaring that a type is "mappable", i.e., that the maps
+// declared in principia::geometry can be act on it through the operator().  To
```
eggrobin: groan

base/mappable.hpp
pleroy: Done.

PR #375, opened 2015-01-10
Fix performance issues and update benchmark numbers.

pleroy:


base/not_null_body.hpp
```diff
@@ -5,6 +5,7 @@
 #include <algorithm>
 #include <utility>
 
+#include "base/macros.hpp"
```
eggrobin: dd

PR #380, opened 2015-01-13
5th phase of the not_null<> migration: add an implicit constructor and remove check_not_null

pleroy:


base/not_null.hpp
```diff
@@ -217,13 +221,19 @@ class not_null {
   bool operator>(not_null const other) const;
 
  private:
+  struct unchecked_tag {
+    inline unchecked_tag() {}
```
eggrobin: `inline` is redundant if you write the definition inline, it's only useful for separate definitions.

pleroy: OBE.

PR #386, opened 2015-01-18
Minimal usage of protocol buffers in Quantity, just to put in place all the mechanisms.

pleroy:


quantities/quantities.hpp
```diff
@@ -133,6 +136,8 @@ class Quantity {
   bool operator==(Quantity const& right) const;
   bool operator!=(Quantity const& right) const;
 
+  void SerializeTo(serialization::Quantity* quantity) const;
```
eggrobin: `not_null`

PR #390, opened 2015-01-21
Cauchy distribution

eggrobin:
Fix the "crashes crash" bug (due to adding 0-part vessels to the physics bubble), the bug where the plugin would still manage the physics bubble at ground level on bodies without an atmosphere, and the bug where a crashed vessel would keep being simulated.
Improved performance by not calling `FindObjectOfType` at every frame.

### DIFF
--- a/ksp_plugin/plugin.hpp
+++ b/ksp_plugin/plugin.hpp
@@ -210,7 +210,10 @@ class Plugin {
   // already in |next_physics_bubble_->parts|.
   virtual void AddVesselToNextPhysicsBubble(GUID const& vessel_guid,
                                             std::vector<IdAndOwnedPart> parts);
+
+  // Returns |bubble_.empty()|.
   virtual bool PhysicsBubbleIsEmpty() const;
+
   // Computes and returns |current_physics_bubble_->displacement_correction|.
   // This is the |World| shift to be applied to the physics bubble in order for
   // it to be in the correct position.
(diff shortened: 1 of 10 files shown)


ksp_plugin/plugin.hpp
```diff
@@ -210,6 +210,7 @@ class Plugin {
   // already in |next_physics_bubble_->parts|.
   virtual void AddVesselToNextPhysicsBubble(GUID const& vessel_guid,
                                             std::vector<IdAndOwnedPart> parts);
+  virtual bool PhysicsBubbleIsEmpty() const;
```
pleroy: Comment, unglue.

ksp_plugin_test/interface_test.cpp
```diff
@@ -386,6 +389,16 @@ TEST_F(InterfaceTest, PhysicsBubble) {
   XYZ const velocity =
       principia__BubbleVelocityCorrection(plugin_.get(), kParentIndex);
   EXPECT_THAT(velocity, Eq(XYZ{66, 55, 44}));
+
+  EXPECT_CALL(*plugin_, PhysicsBubbleIsEmpty()).WillOnce(Return(true));
+  bool const empty = principia__PhysicsBubbleIsEmpty(plugin_.get());
+  EXPECT_THAT(empty, Eq(true));
```
pleroy: `EXPECT_TRUE`

ksp_plugin/plugin.hpp

ksp_plugin_test/interface_test.cpp
eggrobin: Done.

PR #411, opened 2015-01-31
Movable trajectory

### LINKED ISSUE #410: Trajectory should be movable

eggrobin:
Fix #410.

### DIFF
--- a/physics/trajectory_body.hpp
+++ b/physics/trajectory_body.hpp
@@ -170,12 +170,10 @@ template<typename Frame>
 not_null<Trajectory<Frame>*> Trajectory<Frame>::Fork(Instant const& time) {
   auto fork_it = timeline_.find(time);
   CHECK(fork_it != timeline_.end()) << "Fork at nonexistent time";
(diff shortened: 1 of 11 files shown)


physics/trajectory_body.hpp
```diff
@@ -24,6 +25,16 @@ Trajectory<Frame>::Trajectory(not_null<Body const*> const body)
 }
 
 template<typename Frame>
+Trajectory<Frame>::Trajectory(Trajectory&& other)  // NOLINT(build/c++11)
+  : body_(std::move(other.body_)),
+    parent_(std::move(other.parent_)),
+    fork_(std::move(other.fork_)),
+    intrinsic_acceleration_(std::move(other.intrinsic_acceleration_)) {
+  std::swap(children_, other.children_);
```
pleroy: The standard way to do this is:

```
#include <utility>
using std::swap;
swap(a, b);
```
You might think that it's the same thing, but it's not.  See http://en.cppreference.com/w/cpp/concept/Swappable for details.

Issue #409, opened 2015-01-30
geometry::Frame needs a unit test


pleroy:

pleroy:
Fixed by #412.

PR #412, opened 2015-01-31
A unit test for Frame and serialization of frames in the linear and affine maps

pleroy:


pleroy:
Fix #409.

geometry/linear_map.hpp
```diff
@@ -36,7 +36,11 @@ class LinearMap {
 //  template<typename T>
 //  typename base::Mappable<LinearMap, T>::type operator()(T const& t) const;
 //
-// protected:
+ protected:
+  // Serialization of the frames.
+  static void WriteToMessage(not_null<serialization::LinearMap*> const message);
+  static void ReadFromMessage(serialization::LinearMap const& message);
```
eggrobin: Clarify that this does not dispatch.

pleroy: Done.

PR #414, opened 2015-02-01
Serialization of native trajectory iterator

pleroy:
Somewhat preliminary in the sense that the serialization functions can only be called with parameters that verify tight constraints.

### DIFF
--- a/physics/trajectory.hpp
+++ b/physics/trajectory.hpp
@@ -202,11 +202,13 @@ class Trajectory {
     typename Timeline::const_iterator current() const;
     not_null<Trajectory const*> trajectory() const;
 
-    // Helper functions for serialization of the subclasses.
+    // Helper functions for serialization of the subclasses.  The |trajectory|
+    // passed to |ReadFromMessage| must match the one that was denoted by the
+    // iterator when |WriteToMessage| was called.  Most of the time, this means
+    // that |WriteToMessage| should be called by an iterator returned by |first|
+    // and |ReadFromMessage| should be called with a root trajectory.
     void WriteToMessage(
        not_null<serialization::Trajectory::Iterator*> const message) const;
-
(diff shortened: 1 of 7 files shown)


physics/trajectory.hpp
```diff
@@ -190,10 +202,25 @@ class Trajectory {
...
+    static void ReadFromMessage(
+        serialization::Trajectory::Iterator const& message,
+        not_null<Trajectory const*> const trajectory,
+        not_null<Iterator*> const iterator);
+
    private:
+    // |ancestry_| has one more element than |forks_|.  The first element in
+    // |ancestry_| is the root.  There is no element in |forks_| for the root.
+    // It is therefore empty for a root trajectory.
     typename Timeline::const_iterator current_;
```
eggrobin: comment that `current_` is never past the end

eggrobin: or maybe it can be

serialization/physics.proto
```diff
@@ -29,14 +29,19 @@ message OblateBody {
...
-    required Point fork_time = 1;
-    repeated Trajectory trajectories = 2;
-  }
   message InstantaneousDegreesOfFreedom {
     required Point instant = 1;
     required Pair degrees_of_freedom = 2;
   }
+  message Iterator {
+    repeated int32 children_distance = 1;
+    repeated int32 timeline_distance = 2;
```
eggrobin: use a repeated foork

physics/trajectory.hpp
pleroy: Yeah, it can be when the iterator is `at_end()`.

serialization/physics.proto
pleroy: Done.

pleroy:
Renamed `oork -> ork` and `Fork -> NewFork`.

PR #420, opened 2015-02-05
Vessel proto

eggrobin:
`Vessel` serialization, and a big proto file outlining the way forward.

I'm not sure making these big types movable (we started doing that with `Trajectory`) is a good idea. We don't make them move-assignable (otherwise we can't have `const` member variables), so we almost always have to put them in `not_null<std::unique_ptr<T>>`s, which means that factories that return a `T` directly are always used with a `make_not_null_unique<T>`. Moreover, it's easy to break invariants when writing a move constructor: we had to use `swap`s to make sure `Trajectory` was safe (it's still unsafe for past-the end `Iterator`s, although that's a minor issue), here I spent an hour tracking a bug due to a missing `prolongation_(std::move(other.prolongation_)),` which could easily have slipped past testing altogether were `Vessel` not riddled with `CHECK` macros. Note that we can't use default move constructors anyway since MSVC doesn't do those.

Note that we already have `ReadFromMessage`s that return a `not_null<std::unique_ptr<T>>` for `Body` and subclasses (for good reason), so doing so for other high-level types wouldn't introduce style inconsistencies.

Relevant bit of styleguide:
>Make your type copyable/movable if it will be useful, and if it makes sense in the context of the rest of the API. As a rule of thumb, if the behavior (including computational complexity) of a copy isn't immediately obvious to users of your type, your type shouldn't be copyable. If you choose to make it copyable, define both copy operations (constructor and assignment). If your type is copyable and a move operation is more efficient than a copy, define both move operations (constructor and assignment). If your type is not copyable, but the correctness of a move is obvious to users of the type and its fields support it, you may make the type move-only by defining both of the move operations.

>Prefer to define copy and move operations with = default. Defining non-default move operations currently requires a style exception. Remember to review the correctness of any defaulted operations as you would any other code.

>Due to the risk of slicing, avoid providing an assignment operator or public copy/move constructor for a class that's intended to be derived from (and avoid deriving from a class with such members). If your base class needs to be copyable, provide a public virtual Clone() method, and a protected copy constructor that derived classes can use to implement it.

>If you do not want to support copy/move operations on your type, explicitly disable them using = delete or whatever other mechanism your project uses.

### DIFF
--- a/ksp_plugin/vessel.hpp
+++ b/ksp_plugin/vessel.hpp
@@ -22,7 +22,7 @@ class Vessel {
  public:
   Vessel() = delete;
   Vessel(Vessel const&) = delete;
-  Vessel(Vessel&&);
+  Vessel(Vessel&&);  // NOLINT(build/c++11)
   ~Vessel() = default;
 
   // Constructs a vessel whose parent is initially |*parent|.  No transfer of
--- a/ksp_plugin/vessel.hpp
+++ b/ksp_plugin/vessel.hpp
@@ -29,7 +29,6 @@ class Vessel {
   // ownership.
   explicit Vessel(not_null<Celestial const*> const parent);
 
-
   // True if, and only if, |history_| is not null.
   bool is_synchronized() const;
   // True if, and only if, |prolongation_| is not null, i.e., if either
--- a/ksp_plugin/vessel.hpp
+++ b/ksp_plugin/vessel.hpp
@@ -8,6 +8,7 @@
 #include "physics/massless_body.hpp"
 #include "physics/trajectory.hpp"
 #include "quantities/named_quantities.hpp"
+#include "serialization/ksp_plugin.pb.h"
 
 using principia::physics::MasslessBody;
 using principia::physics::Trajectory;
@@ -65,6 +66,11 @@ class Vessel {
   // The vessel must satisfy |is_synchronized()| and |is_initialized()|,
   // |owned_prolongation_| must be null.
   void ResetProlongation(Instant const& time);
+
+  // The vessel must satisfy |is_initialized()|.
+  void WriteToMessage(not_null<serialization::Vessel*> const message) const;
+  static Vessel ReadFromMessage(serialization::Vessel const& message,
+                                not_null<Celestial const*> const parent);
 
  private:
   MasslessBody const body_;
--- a/ksp_plugin/vessel.hpp
+++ b/ksp_plugin/vessel.hpp
@@ -22,12 +22,13 @@ class Vessel {
  public:
   Vessel() = delete;
   Vessel(Vessel const&) = delete;
-  Vessel(Vessel&&) = delete;
+  Vessel(Vessel&&);
   ~Vessel() = default;
 
   // Constructs a vessel whose parent is initially |*parent|.  No transfer of
   // ownership.
   explicit Vessel(not_null<Celestial const*> const parent);
+
 
   // True if, and only if, |history_| is not null.
   bool is_synchronized() const;
--- a/ksp_plugin/vessel_body.hpp
+++ b/ksp_plugin/vessel_body.hpp
@@ -78,5 +78,44 @@ inline void Vessel::ResetProlongation(Instant const& time) {
   prolongation_ = history_->NewFork(time);
 }
 
+inline void Vessel::WriteToMessage(
+    not_null<serialization::Vessel*> const message) const {
+  CHECK(is_initialized());
+  body_.WriteToMessage(message->mutable_body());
+  if (is_synchronized()) {
+    history_->WriteToMessage(
+        message->mutable_history_and_prolongation()->mutable_history());
+    prolongation_->WritePointerToMessage(
+        message->mutable_history_and_prolongation()->mutable_prolongation());
+  } else {
+    owned_prolongation_->WriteToMessage(message->mutable_owned_prolongation());
+  }
+}
+
+inline Vessel Vessel::ReadFromMessage(serialization::Vessel const& message,
+                                      not_null<Celestial const*> const parent) {
+  Vessel vessel(parent);
+  // NOTE(egg): for now we do not read the |MassiveBody| as can contain no
+  // information.
+  if (message.has_history_and_prolongation()) {
+    vessel.history_ =
+        std::make_unique<Trajectory<Barycentric>>(
+            Trajectory<Barycentric>::ReadFromMessage(
+                message.history_and_prolongation().history(), &vessel.body_));
+    vessel.prolongation_ =
+        Trajectory<Barycentric>::ReadPointerFromMessage(
+            message.history_and_prolongation().prolongation(),
+            vessel.history_.get());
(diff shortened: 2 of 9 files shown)


ksp_plugin/vessel.hpp
```diff
@@ -21,13 +22,14 @@ class Vessel {
...
   Vessel(Vessel const&) = delete;
-  Vessel(Vessel&&) = delete;
+  Vessel(Vessel&&);  // NOLINT(build/c++11)
   ~Vessel() = default;
 
   // Constructs a vessel whose parent is initially |*parent|.  No transfer of
   // ownership.
   explicit Vessel(not_null<Celestial const*> const parent);
 
+
```
pleroy: Why the new line?

ksp_plugin/vessel_body.hpp
```diff
@@ -5,6 +5,13 @@
 namespace principia {
 namespace ksp_plugin {
 
+inline Vessel::Vessel(Vessel&& other)  // NOLINT(build/c++11)
```
pleroy: So this would be an `= default` if this existed, right?  Maybe a TODO to use that when VS supports it.

ksp_plugin/vessel_body.hpp
```diff
@@ -78,5 +85,46 @@ inline void Vessel::ResetProlongation(Instant const& time) {
...
+        message->mutable_history_and_prolongation()->mutable_prolongation());
+  } else {
+    owned_prolongation_->WriteToMessage(message->mutable_owned_prolongation());
+  }
+}
+
+inline Vessel Vessel::ReadFromMessage(serialization::Vessel const& message,
+                                      not_null<Celestial const*> const parent) {
+  Vessel vessel(parent);
+  // NOTE(egg): for now we do not read the |MassiveBody| as can contain no
```
pleroy: as it can

pleroy: Also, it's probably Massless.

ksp_plugin_test/vessel_test.cpp
```diff
@@ -0,0 +1,97 @@
...
+                              15 * Metre / Second,
+                              16 * Metre / Second})};
+  Instant const t1_ = kUniversalTimeEpoch;
+  Instant const t2_ = kUniversalTimeEpoch + 42 * Second;
+};
+
+using VesselDeathTest = VesselTest;
+
+TEST_F(VesselTest, InitializationAndSynchronization) {
+  // TODO(egg): mutable_history() should probably return a |not_null| and check
```
pleroy: Open an issue?

serialization/ksp_plugin.proto
```diff
@@ -0,0 +1,88 @@
+syntax = "proto2";
+
+import "serialization/geometry.proto";
+import "serialization/physics.proto";
+import "serialization/quantities.proto";
+
+package principia.serialization;
+
+message HistoryAndProlongation {
```
pleroy: Order the messages alphabetically.

serialization/ksp_plugin.proto
```diff
@@ -0,0 +1,88 @@
...
+}
+
+message Part {
+  required Pair degrees_of_freedom = 1;
+  required Quantity mass = 2;
+  required Multivector gravitational_acceleration_to_be_applied_by_ksp = 3;
+}
+
+message PhysicsBubble {
+  message FullState {
```
pleroy: Do we really need this message?  It could also just be inlined in the containing message.

serialization/ksp_plugin.proto
```diff
@@ -0,0 +1,88 @@
...
+  }
+  required MasslessBody body = 1;
+  optional FullState current = 2;
+}
+
+message Plugin {
+  // NOTE(egg): Assuming 32-bit. If we have 2^32 vessels we'll probably
+  // encounter other issues on the way.
+  message VesselPointer {
+    required uint32 distance_from_vessels_begin = 1;
```
pleroy: No uint please unless it's a bitmask or an id or some such.

serialization/ksp_plugin.proto
```diff
@@ -0,0 +1,88 @@
...
+    required uint32 distance_from_celestials_begin = 1;
+  }
+  message VesselPointerToParent {
+    required VesselPointer vessel = 1;
+    required CelestialPointer parent = 2;
+  }
+  message CelestialPointerToParent {
+    required CelestialPointer celestial = 1;
+    optional CelestialPointer parent = 2;
+  }
```
pleroy: These message could be simplified by dropping the Vessel/Celestial distinction.  It seem to me that this would make things a bit simpler/more readable.

pleroy: As a matter of fact all of this is complicated because it follows too closely the C++ code.  For instance, I would represent the parent/child relationship by having an `required int32 parent` in GuidToVessel and IndexToCelestial, where that field is an index in the celestial array.  Similarly I would represent dirtyness by having a bool in GuidToVessel.

serialization/ksp_plugin.proto
```diff
@@ -0,0 +1,88 @@
...
+    required string guid = 1;
+    required Vessel vessel = 2;
+  }
+  message IndexToCelestial {
+    required int32 index = 1;
+    required Celestial celestial = 2;
+  }
+  repeated GuidToVessel vessels = 1;
+  repeated IndexToCelestial celestials = 2;
+  repeated VesselPointerToParent vessel_parents = 3;
```
pleroy: Remove 2x.

serialization/ksp_plugin.proto
```diff
@@ -0,0 +1,88 @@
...
+  repeated CelestialPointerToParent celestial_parents = 4;
+  repeated VesselPointer dirty_vessels = 5;
+  required PhysicsBubble bubble = 6;
+  required Quantity planetarium_rotation = 7;
+  required Point current_time = 8;
+  required CelestialPointer sun = 9;
```
pleroy: Use an int32 index.

ksp_plugin/vessel_body.hpp
eggrobin: yes but it's going to be a `delete` in the next CL.

serialization/ksp_plugin.proto
eggrobin: yes

ksp_plugin/vessel.hpp
eggrobin: Removed.

ksp_plugin/vessel_body.hpp
eggrobin: Done.

ksp_plugin_test/vessel_test.cpp
eggrobin: #422.

serialization/ksp_plugin.proto
eggrobin: Done.

serialization/ksp_plugin.proto
eggrobin: Done.

Issue #347, opened 2014-12-13
Verbosity levels


eggrobin:
We use `VLOG(1)` everywhere right now, and that quickly makes the logs unmanageable.
We should define criteria for verbosity levels. Since we're a library and we can't set flags, we also need to make verbosity tunable in-game.
###### Proposal

Where O is for the following variables going to positive infinity
f = frequency of `FixedUpdate`s
v = number of vessels
p = number of parts
c = number of celestials
t = length of histories

| severity or verbose level | frequency \* size | example |
| --: | :-- | :-- |
| `LOG(INFO)` | aperiodic | ships exploding / appearing |
| `VLOG(1)` | O(f) | calls to `AdvanceTime` |
| `VLOG(2)` | O((v+c)f) | calls to `InsertOrKeepVessel` |
| `VLOG(3)` | O((v+c+p)f) | arguments of `AddVesselToNextPhysicsBubble` |
| `VLOG(4)` | O((v+c+p+t)f) | return value of `RenderedVesselTrajectory` |

pleroy:
We don't debug with `VLOG` anymore.

PR #443, opened 2015-02-16
Complete normed vector space

eggrobin:
`PluginAdapter` was renamed since the class name is the `name` of the `SCENARIO` `ConfigNode` in which we write things, it's probably best to have the name `Principia` somewhere there.

### DIFF
--- a/ksp_plugin/interface.cpp
+++ b/ksp_plugin/interface.cpp
@@ -360,27 +360,35 @@ double principia__current_time(Plugin const* const plugin) {
 }
 
 char const* principia__SerializePlugin(Plugin const* const plugin) {
+  LOG(INFO) << __FUNCTION__;
   CHECK_NOTNULL(plugin);
   principia::serialization::Plugin message;
   plugin->WriteToMessage(&message);
   // TODO(egg): reimplement with |ZeroCopyStream|.
   std::vector<uint8_t> bytes(message.ByteSize());
+  LOG(INFO) << "allocated vector for serialization";
   message.SerializeWithCachedSizesToArray(bytes.data());
+  LOG(INFO) << "serialized";
   // Leave room for the null terminator.
   std::size_t const hexadecimal_size = (bytes.size() << 1) + 1;
   auto hexadecimal = std::make_unique<uint8_t[]>(hexadecimal_size);
+  LOG(INFO) << "allocated buffer for hex string";
   HexadecimalEncode(bytes.data(), bytes.size(),
                     hexadecimal.get(), hexadecimal_size);
+  LOG(INFO) << "wrote hex string";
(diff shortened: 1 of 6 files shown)


ksp_plugin/interface.cpp
```diff
@@ -360,6 +360,7 @@ double principia__current_time(Plugin const* const plugin) {
 }
 
 char const* principia__SerializePlugin(Plugin const* const plugin) {
+  LOG(INFO) << __FUNCTION__;
```
pleroy: VLOG(1), maybe?  (Many places.)

ksp_plugin_adapter/ksp_plugin_adapter.cs
```diff
@@ -6,8 +6,11 @@
 namespace principia {
 namespace ksp_plugin_adapter {
 
-[KSPAddon(startup : KSPAddon.Startup.MainMenu, once : false)]
-public partial class PluginAdapter : UnityEngine.MonoBehaviour {
+[KSPScenario(createOptions: ScenarioCreationOptions.AddToAllGames,
+             tgtScenes: new GameScenes[]{GameScenes.SPACECENTER,
+                                         GameScenes.EDITOR, GameScenes.FLIGHT,
```
pleroy: One parameter per line (unless it's autoformatted by VS).

ksp_plugin_adapter/ksp_plugin_adapter.cs
```diff
@@ -221,29 +228,54 @@ where is_in_physics_bubble(vessel)
...
+  // These functions override virtual ones from |ScenarioModule|, but it seems
+  // that they're actually called by reflection, so that bad things happen
+  // if you don't have, e.g., a function called |OnAwake()| that calls
+  // |base.OnAwake()|.  It doesn't matter whether the functions are public or
+  // private, overriding or hiding though.
+
+  public override void OnAwake() {
+    base.OnAwake();
+    // While we're here, we might as well log.
+    Log.Info("principia.ksp_plugin_adapter.PluginAdapter.OnAwake()");
```
pleroy: PrincipiaPluginAdapter?  (Several places.)

ksp_plugin/interface.cpp
eggrobin: It's aperiodic (called only on scene load/save) which should be `INFO` per #347.

ksp_plugin_adapter/ksp_plugin_adapter.cs
eggrobin: Done.

ksp_plugin_adapter/ksp_plugin_adapter.cs
eggrobin: Done.

PR #450, opened 2015-02-20
Add base::FindOrDie and use it instead of find when appropriate

pleroy:


base/map_util.hpp
```diff
@@ -0,0 +1,65 @@
...
+
+// Same as above, but returns a non-const reference.
+template <class Collection>
+typename Collection::value_type::second_type&
+FindOrDie(Collection& collection,  // NOLINT
+          const typename Collection::value_type::first_type& key) {
+  typename Collection::iterator it = collection.find(key);
+  CHECK(it != collection.end()) << "Map key not found: " << key;
+  return it->second;
+}
```
eggrobin: Do we actually use this one?

pleroy: As discussed, we don't currently, but it's generally useful.  Fixed the `NOLINT`.

PR #455, opened 2015-02-23
use transparent operators

eggrobin:
See [the proposal](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2012/n3421.htm) that led to their introduction, see [an example implementation](http://en.cppreference.com/w/cpp/utility/functional/multiplies_void). Use analogous construct for `Wedge` and `InnerProduct`.


geometry/r3_element_test.cpp
```diff
@@ -76,11 +78,9 @@ TEST_F(R3ElementTest, Dumb3Vector) {
 
 TEST_F(R3ElementTest, MixedProduct) {
   testing_utilities::TestBilinearMap(
-      Times<R3Element<Length>, Time, R3Element<Speed>>,
-      1 * Second, 1 * JulianYear, u_, v_, 42.0, 0, 1);
+      std::multiplies<>(), 1 * Second, 1 * JulianYear, u_, v_, 42.0, 0, 1);
   testing_utilities::TestBilinearMap(
-      Times<R3Element<Length>, R3Element<Speed>, Time>, w_, a_,
-      -1 * Day, 1 * Parsec / SpeedOfLight, -π, 0, 1);
+      std::multiplies<>(), u_, v_, -1 * Day, 1 * Parsec / SpeedOfLight, 1);
```
pleroy: You changed the parameters.

eggrobin: Done.

PR #489, opened 2015-03-28
Fix a bunch of Clang error and change Trajectory to hold children directly instead of through pointers

pleroy:


base/array.hpp
```diff
@@ -33,11 +34,13 @@ struct UniqueArray {
...
   template<typename Size,
-           typename = std::enable_if<std::is_integral<Size>::value>::type>
+           typename =
+              typename std::enable_if<std::is_integral<Size>::value>::type>
   explicit UniqueArray(Size const size);
   // Takes ownership of an existing array.
   template<typename Size,
-           typename = std::enable_if<std::is_integral<Size>::value>::type>
+           typename =
+              typename std::enable_if<std::is_integral<Size>::value>::type>
```
eggrobin: Consistent Badindent.

PR #503, opened 2015-05-02
Add a method to forget the histories before a particular point in time, for use from the C# code

pleroy:


eggrobin:
ForgetBefore will have to support forgetting before points that are not trajectory points, let's do that tomorrow in another CL.

PR #507, opened 2015-05-04
Add a hook to control cacheability of lazy trajectories

pleroy:
This fixes a bug where the graph of predictions was oscillating between two trajectories because of deallocated/reallocated pointers left in the cache.


ksp_plugin/plugin.cpp
```diff
@@ -310,9 +312,11 @@ Plugin::NewBarycentricRotatingTransforms(Index const primary_index,
       FindOrDie(celestials_, primary_index).get();
   not_null<Celestial*> secondary =
       FindOrDie(celestials_, secondary_index).get();
-  return RenderingTransforms::BarycentricRotating(*primary,
-             *secondary,
-             &MobileInterface::prolongation);
+  auto transforms = RenderingTransforms::BarycentricRotating(*primary,
```
eggrobin: moo

PR #515, opened 2015-05-09
Fix prediction stepsize selection

eggrobin:


mathematica/mathematica.vcxproj
```diff
@@ -128,6 +128,7 @@
   </ItemGroup>
   <ItemGroup>
     <None Include="generate_graphs.wl" />
+    <None Include="navball.wl" />
```
pleroy: Why is this in this PR.

PR #524, opened 2015-05-13
Change Makefile packaging to include GameData dir

Norgg:
Also switches to git describe for zip name


pleroy:
Thanks for doing this, merging.

PR #526, opened 2015-05-15
Support for representing and evaluating Чебышёв series

pleroy:


numerics/чебышёв_series.hpp
```diff
@@ -0,0 +1,41 @@
+﻿#pragma once
+
+#include <vector>
+
+#include "geometry/named_quantities.hpp"
+#include "quantities/quantities.hpp"
+
+// Spelling: Чебышёв ЧЕБЫШЁВ чебышёв
```
eggrobin: Remove

numerics/чебышёв_series_body.hpp
```diff
@@ -0,0 +1,46 @@
...
+  // Precomputed to save operations at the expense of some accuracy loss.
+  Time const duration = t_max_ - t_min_;
+  t_mean_ = t_min_ + 0.5 * duration;
+  two_over_duration_ = 2 / duration;
+}
+
+template<typename Scalar>
+Scalar ЧебышёвSeries<Scalar>::Evaluate(Instant const& t) const {
+  double const scaled_t = (t - t_mean_) * two_over_duration_;
+  double const two_scaled_t = scaled_t + scaled_t;
```
eggrobin: Does the compiler not optimize this on its own? Is the performance impact measurable?

numerics/чебышёв_series.hpp
pleroy: No, please, let's keep it here, it's a pain to have to look it up and do the case conversion.

numerics/чебышёв_series_body.hpp
pleroy: I am pretty sure that any decent compiler will hoist this out of the loop, but why depend on that?

PR #528, opened 2015-05-16
Renaming for nonsymplectic integrators

eggrobin:
Also Unicode filenames.


pleroy:
So change the includes, maybe?

PR #532, opened 2015-05-19
Mathematica notebook to implement the Newhall ephemerides and the resulting C++ matrices

pleroy:
Not used to build series yet.


eggrobin:
Remove include, add it as a source file, change the extension to `.mathematica.cpp`.
We'll make a separate binary for this kind of table at some point.

eggrobin:
Eventually, it will need a header.

PR #552, opened 2015-05-29
Add class ContinousTrajectory to approximate discrete points using Chebyshev series

pleroy:


numerics/чебышёв_series.hpp
```diff
@@ -27,13 +27,23 @@ class ЧебышёвSeries {
   // The element at position i in |coefficients| is the coefficient of Tᵢ.  The
   // polynomials are scaled to the interval [t_min, t_max], which must be
   // nonempty.
-  explicit ЧебышёвSeries(std::vector<Vector> const& coefficients,
-                         Instant const& t_min,
-                         Instant const& t_max);
+  ЧебышёвSeries(std::vector<Vector> const& coefficients,
+                Instant const& t_min,
+                Instant const& t_max);
+  explicit ЧебышёвSeries(ЧебышёвSeries&& other);
```
eggrobin: Why is the move constructor explicit? That means it's not MoveConstructible, so e.g. [`std::swap`](http://en.cppreference.com/w/cpp/algorithm/swap) doesn't work.

numerics/чебышёв_series_body.hpp
```diff
@@ -28,6 +28,27 @@ template<typename Vector>
   two_over_duration_ = 2 / duration;
 }
 
+
+template<typename Vector>
+ЧебышёвSeries<Vector>::ЧебышёвSeries(ЧебышёвSeries&& other)
+    : coefficients_(std::move(other.coefficients_)),
+      degree_(other.degree_),
```
eggrobin: Why don't we move `degree_`?

physics/continuous_trajectory_body.hpp
```diff
@@ -0,0 +1,228 @@
...
+
+template<typename Frame>
+bool ContinuousTrajectory<Frame>::empty() const {
+  return series_.empty();
+}
+
+template<typename Frame>
+Instant ContinuousTrajectory<Frame>::t_min() const {
+  CHECK(!empty()) << "Empty trajectory";
+  return *first_time_;
```
eggrobin: Why not `series_.front().t_min();`? Why do we need `first_time_`?

physics/continuous_trajectory_test.cpp
```diff
@@ -0,0 +1,196 @@
...
+  int const kNumberOfSubsteps = 50;
+  Time const kStep = 0.01 * Second;
+  Instant const t0;
+
+  auto position_function =
+      [t0](Instant const t) {
+        return World::origin +
+            Displacement<World>({(t - t0) * 3 * Metre / Second,
+                                (t - t0) * 5 * Metre / Second,
+                                (t - t0) * (-2) * Metre / Second});
```
eggrobin: Prepend a space to lines 68 and 69.

numerics/чебышёв_series.hpp
pleroy: Done.

numerics/чебышёв_series_body.hpp
pleroy: Readability?  Why have a call to `std::move` when it has no usefulness whatsoever.

physics/continuous_trajectory_body.hpp
pleroy: `first_time_` is set by `ForgetBefore` and may not be at the beginning of a series (the semantics of a *continuous* trajectory implies that you should be able to cut it at any place regardless of its implementation details).

physics/continuous_trajectory_test.cpp
pleroy: Done.

numerics/чебышёв_series_body.hpp
pleroy: Done.

PR #553, opened 2015-05-31
Ephemeris header proposal

eggrobin:


physics/ephemeris.hpp
```diff
@@ -0,0 +1,79 @@
...
+
+template<typename Frame>
+class Ephemeris {
+  static_assert(Frame::is_inertial, "Frame must be inertial");
+
+ public:
+  // The equation describing the motion of the |bodies_|.
+  using PlanetaryMotion =
+      SpecialSecondOrderDifferentialEquation<Position<Frame>>;
+  PlanetaryMotion::SystemState last_state_;
```
pleroy: Wat?  A public data member?

physics/ephemeris.hpp
```diff
@@ -0,0 +1,79 @@
...
+      std::vector<not_null<std::unique_ptr<MassiveBody>>> bodies,
+      std::vector<DegreesOfFreedom<Frame>> initial_state,
+      Instant const& initial_time,
+      FixedStepSizeIntegrator<PlanetaryMotion> const& planetary_integrator,
+      Time const& step_size);
+
+  ContinuousTrajectory<Frame> const& trajectory(
+      not_null<MassiveBody const*>) const;
+
+  // The common |t_min| of the trajectories.
```
pleroy: I would define the semantics in a way that doesn't assume that all trajectories are aligned: the max of the t_min of the trajectories.  From the perspective of the client, what matters is that it can safely evaluate the trajectories from t_min to t_max.

physics/ephemeris.hpp
```diff
@@ -0,0 +1,79 @@
...
+  Instant t_max() const;
+
+  // Calls |ForgetBefore| on all trajectories.
+  void ForgetBefore(Instant const& t);
+
+  // Prolongs the ephemeris up to at least |t|.  After the call, |t_max() >= t|.
+  void Prolong(Instant const& t);
+
+  // Computes the trajectory of a massless body in the gravitational potential
+  // described by |this| and subject to the given |intrinsic_acceleration| up to
```
pleroy: Unclear what 'up to' applies to: acceleration?  potential?  trajectory?

physics/ephemeris.hpp
```diff
@@ -0,0 +1,79 @@
...
+  // exactly |t|.
+  // If |t > t_max()|, calls |Prolong(t)| beforehand.
+  void Advance(
```
pleroy: Advance is not a good name.  I would expect Ephemeris::Advance to, well, advance the ephemeris.

physics/ephemeris.hpp
```diff
@@ -0,0 +1,79 @@
...
+      not_null<Trajectory<Frame>*> const trajectory,
+      std::function<
+          Vector<Acceleration, Frame>(
+              Instant const&)> intrinsic_acceleration,
+      RelativeDegreesOfFreedom const& tolerance,
```
pleroy: The tolerance needs documentation.

physics/ephemeris.hpp
```diff
@@ -0,0 +1,79 @@
...
+  // The common |t_min| of the trajectories.
+  Instant t_min() const;
+  // The common |t_max| of the trajectories.
+  Instant t_max() const;
+
+  // Calls |ForgetBefore| on all trajectories.
+  void ForgetBefore(Instant const& t);
+
+  // Prolongs the ephemeris up to at least |t|.  After the call, |t_max() >= t|.
+  void Prolong(Instant const& t);
```
pleroy: Should this take some indication of tolerance?

physics/ephemeris.hpp
```diff
@@ -0,0 +1,79 @@
...
+namespace principia {
+
+using integrators::AdaptiveStepSizeIntegrator;
+using integrators::FixedStepSizeIntegrator;
+using integrators::SpecialSecondOrderDifferentialEquation;
+
+namespace physics {
+
+template<typename Frame>
+class Ephemeris {
```
pleroy: So I think that this should ultimately just replace the existing NBodySystem (and be called NBodySystem), although we cannot do that yet as it would break everything.

physics/ephemeris.hpp
eggrobin: wat indeed.

physics/ephemeris.hpp
eggrobin: makes sense.

physics/ephemeris.hpp
eggrobin: Let's pass the interpolation tolerances at construction.

physics/ephemeris.hpp
pleroy: Suggestion:

Computes the trajectory of a massless body in the gravitational potential of the bodies of this object's continuous trajectories, using the given |integrator|.  The massless body may have an |intrinsic_acceleration| representing e.g. engine burns.  |tolerance| is ... The last instant of the computed trajectory is exactly |t|.

physics/ephemeris.hpp
pleroy: Suggestion: ProlongContinuousTrajectories.

physics/ephemeris.hpp
pleroy: Suggestion: ProlongMasslessBodyTrajectory.

PR #554, opened 2015-05-31
Benchmark automation

eggrobin:


benchmark_automation/benchmark_automation.cs
```diff
@@ -0,0 +1,110 @@
...
+            CreateNoWindow = true
+          }
+        };
+        process.Start();
+        while (!process.StandardOutput.EndOfStream) {
+          String line = process.StandardOutput.ReadLine();
+          String[] words =
+              line.Split(separator : new Char[]{' '},
+                         options   : StringSplitOptions.RemoveEmptyEntries);
+          if (words[0].EndsWith("_mean")) {
```
pleroy: This mean/stddev analysis is only going to work for benchmarks that have a repetition flag.  For the others you want to get the (single) result line too.

benchmark_automation/benchmark_automation.cs
```diff
@@ -0,0 +1,110 @@
...
+            String benchmark_name = words[0].Substring(0, words[0].Length - 5);
```
pleroy: If we gave a name to "_mean" we could use its length instead of 5.  Same for 7.

eggrobin: Done.

benchmark_automation/benchmark_automation.cs
eggrobin: Done.

eggrobin: Not that it matters with the commands we currently run.

PR #560, opened 2015-06-03
First implementation of Ephemeris

pleroy:
Only Prolong, not Flow, for the time being, and with limited testing (only the Earth-Moon system).

### DIFF
--- a/physics/continuous_trajectory_body.hpp
+++ b/physics/continuous_trajectory_body.hpp
@@ -5,8 +5,12 @@
 #include <vector>
 
 #include "physics/continuous_trajectory.hpp"
+#include "testing_utilities/numerics.hpp"
 
 namespace principia {
+
+using testing_utilities::ULPDistance;
+
 namespace physics {
 
 namespace {
@@ -55,7 +59,11 @@ void ContinuousTrajectory<Frame>::Append(
   if (first_time_ == nullptr) {
     first_time_ = std::make_unique<Instant>(time);
   } else {
-    CHECK_EQ(last_points_.back().first + step_, time)
+    Instant const t0;
+    CHECK_GE(1,
+             ULPDistance((last_points_.back().first + step_ - t0) /
+                             SIUnit<Time>(),
+                         (time - t0) / SIUnit<Time>()))
         << "Append at times that are not equally spaced";
   }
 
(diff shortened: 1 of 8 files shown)


physics/continuous_trajectory_body.hpp
```diff
@@ -5,8 +5,12 @@
 #include <vector>
 
 #include "physics/continuous_trajectory.hpp"
+#include "testing_utilities/numerics.hpp"
```
eggrobin: Perhaps this should be promoted to non-testing numerics.

physics/ephemeris_body.hpp
```diff
@@ -0,0 +1,318 @@
...
+  problem.equation = equation_;
+  problem.append_state = std::bind(&Ephemeris::AppendState, this, _1);
+  problem.t_final = t;
+  problem.initial_state = &last_state_;
+
+  // Perform the integration.  Note that we may have to iterate until |t_max()|
+  // actually reaches |t| because the last series may not be fully determined
+  // after the first integration.
+  do {
+    planetary_integrator_.Solve(problem, step_);
```
eggrobin: It's not trivial that this isn't stupid. Comment.

physics/ephemeris_body.hpp
```diff
@@ -0,0 +1,318 @@
...
+    if (body1_is_oblate || body2_is_oblate) {
+      Exponentiation<Length, -2> const one_over_Δq_squared = 1 / Δq_squared;
+      if (body1_is_oblate) {
+        Vector<Acceleration, Frame> const order_2_zonal_acceleration1 =
+            Order2ZonalAcceleration<Frame>(
+                static_cast<OblateBody<Frame> const &>(body1),
+                Δq,
+                one_over_Δq_squared,
+                one_over_Δq_cubed);
+        (*accelerations)[b2] += order_2_zonal_acceleration1;
```
eggrobin: This looks like it doesn't respect Newton's third law (see above)...
Investigate.
Later.

physics/ephemeris_body.hpp
pleroy: Done.

physics/ephemeris_body.hpp
pleroy: Commented why.

PR #563, opened 2015-06-04
And then we'll document the code.

eggrobin:


integrators/ordinary_differential_equations.hpp
```diff
@@ -37,9 +42,13 @@ struct SpecialSecondOrderDifferentialEquation {
     std::vector<Displacement> position_error;
     std::vector<Velocity> velocity_error;
   };
+  // A functor that computes f(q, t) and stores it in |*accelerations|.
+  // |*accelerations| must have the same size as |positions|.
```
pleroy: On return, |*accelerations|...

Also, do we ensure anything about the input value of accelerations, eg it's empty?  Or it has the right number of 0 accelerations?

integrators/ordinary_differential_equations.hpp
```diff
@@ -37,9 +42,13 @@ struct SpecialSecondOrderDifferentialEquation {
...
   RightHandSideComputation compute_acceleration;
 };
 
+// An initial value problem, together with a final time for the solution
```
pleroy: "initial value problem" is a bit mysterious.  This should have the word "integration" somewhere.

integrators/ordinary_differential_equations.hpp
```diff
@@ -48,6 +57,9 @@ struct IntegrationProblem {
   std::function<void(typename ODE::SystemState const& state)> append_state;
 };
 
+// Settings for for adaptive step size integration.
+// The step size will be scaled so that the estimated new tolerance to error
+// ratio becomes |safety_factor|.
```
pleroy: The functor and the initial time step need better documentation.

integrators/ordinary_differential_equations.hpp
```diff
@@ -59,22 +71,32 @@ struct AdaptiveStepSize {
...
 class Integrator {
  public:
   using ODE = DifferentialEquation;
 };
 
+// An integrator using a fixed step size.
 template<typename DifferentialEquation>
 class FixedStepSizeIntegrator : public Integrator<DifferentialEquation> {
  public:
+  // The last call to |problem.append_state| have a |state.time.value| equal to
```
pleroy: s/have/has/

integrators/ordinary_differential_equations.hpp
```diff
@@ -59,22 +71,32 @@ struct AdaptiveStepSize {
...
+  // the unique |Instant|of the form |problem.t_final + n * step| in
```
pleroy: s/of/ of/

integrators/ordinary_differential_equations.hpp
eggrobin: We'll, it's the [standard terminology](http://en.wikipedia.org/wiki/Initial_value_problem), I'm not going to put a Calculus I course in there...

integrators/ordinary_differential_equations.hpp
eggrobin: No, that statement about the size is a precondition, not a postcondition, I'll reword to clarify.
There is no precondition on the values stored in `*accelerations` (in the current implementations I think they're just what's left from the previous call).

integrators/ordinary_differential_equations.hpp
eggrobin: Done.

integrators/ordinary_differential_equations.hpp
eggrobin: Done.

integrators/ordinary_differential_equations.hpp
eggrobin: Done.

integrators/ordinary_differential_equations.hpp
eggrobin: Done.

PR #586, opened 2015-06-17
Continuous trajectory serialization

pleroy:


geometry/serialization.hpp
```diff
@@ -0,0 +1,22 @@
...
+namespace geometry {
+
+// A helper class that serializes a |double|, a |Quantity| or a |Multivector|
+// to a protobuf structure like:
+//
+// message Message {
+//   oneof message {
+//     double double = 1;
+//     Quantity quantity = 2;
+//     Multivector multivector = 3;
```
eggrobin: The actual values of 1, 2, and 3 don't matter, do they? this should be made clear.

pleroy: As discussed, let's not do that.

PR #591, opened 2015-06-20
Serialization of Ephemeris

pleroy:
And of several other auxiliary classes: DoublePrecision, SystemState, the integrators.  And numerous small fixes here and there.


physics/ephemeris_body.hpp
```diff
@@ -75,7 +75,7 @@ Ephemeris<Frame>::Ephemeris(
   CHECK(!bodies.empty());
   CHECK_EQ(bodies.size(), initial_state.size());
 
-  last_state_.time = initial_time;
+  last_state_.time.value = initial_time;
```
eggrobin: revert

PR #595, opened 2015-07-22
Fixed step termination

eggrobin:
We like that bug, let's make it a feature.

### DIFF
--- a/integrators/symplectic_runge_kutta_nyström_integrator_test.cpp
+++ b/integrators/symplectic_runge_kutta_nyström_integrator_test.cpp
@@ -79,6 +79,60 @@ void ComputeHarmonicOscillatorAcceleration(
   }
 }
 
+template<typename Integrator>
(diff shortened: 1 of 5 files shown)


integrators/symplectic_runge_kutta_nyström_integrator_test.cpp
```diff
@@ -465,6 +524,11 @@ TEST_P(SymplecticRungeKuttaNyströmIntegratorTest, Convergence) {
   GetParam().RunConvergence();
 }
 
+TEST_P(SymplecticRungeKuttaNyströmIntegratorTest, Termination) {
+  LOG(INFO) << GetParam();
```
pleroy: Do we need the noisy LOG?

eggrobin: Yes.

PR #614, opened 2015-08-10
Fix #442 by setting a log handler for the protobuf library

### LINKED ISSUE #442: Forward protobuf logs to glog
using [this](https://github.com/google/protobuf/blob/master/src/google/protobuf/stubs/common.h#L790)
and the following from glog.
```cpp
// A non-macro interface to the log facility; (useful
// when the logging level is not a compile-time constant).
inline void LogAtLevel(int const severity, std::string const &msg) {
  LogMessage(__FILE__, __LINE__, severity).stream() << msg;
}
```

pleroy:


ksp_plugin/interface.cpp
```diff
@@ -94,6 +94,16 @@ void principia__InitGoogleLogging() {
     google::SetLogDestination(google::WARNING, "glog/Principia/WARNING.");
     google::SetLogDestination(google::INFO, "glog/Principia/INFO.");
     google::InitGoogleLogging("Principia");
+
+    google::protobuf::SetLogHandler(
+        [](google::protobuf::LogLevel const level,
+           const char* const filename,
```
eggrobin: `char const* const`

PR #635, opened 2015-08-30
Use optional in a few places

pleroy:


eggrobin:
Looks good, but we'll want a checking `optional`, otherwise it's going to be even clumsier than with pointers.

PR #665, opened 2015-09-22
SolarSystem is a class for constructing bodies and ephemeris from configuration files

pleroy:
Also adding the project `astronomy`.

SolarSystem is not used yet outside of its test.


astronomy/initial_state.proto.txt
```diff
@@ -0,0 +1,316 @@
+initial_state {
+  needs : "RealSolarSystem"
+  frame : ICRF_J2000_EQUATOR
+
+  # The time of the initial state, as a Julian date.
```
eggrobin: as a TDB Julian date.

astronomy/initial_state.proto.txt
```diff
@@ -0,0 +1,316 @@
...
+  # This is 1950-01-01T00:00:00 TDB, we set the epoch one non-leap
```
eggrobin: This is 1950-01-01T00:00:00 TDB. the rest is no longer relevant or sensible.

astronomy/initial_state.proto.txt
```diff
@@ -0,0 +1,316 @@
+initial_state {
```
eggrobin: The scripts that generate this file should be updated in a separate CL. They should also sort the bodies so that diffing is easier. #666.

PR #673, opened 2015-09-27
Generate configuration for KSP from physics::SolarSystem

pleroy:


eggrobin:
`needs` isn't astronomy, it shouldn't be in the proto (similar to the 1951 real solar system date).

PR #675, opened 2015-09-27
Now that Clang 3.7 is out, fix all (?) the problems that have accumulated in the code base

pleroy:


ksp_plugin/interface.cpp
```diff
@@ -618,7 +618,7 @@ void principia__DeserializePlugin(char const* const serialization,
 
   // Push the data, taking ownership of it.
   (*deserializer)->Push(Bytes(&bytes[0], byte_size),
-                        [bytes]() { delete bytes; });
+                        [bytes]() { delete[] bytes; });
```
eggrobin: Wow.

PR #677, opened 2015-09-28
[WIP] Linux clang fixes

Norgg:
Still work in progress, based off mockingbirdnest master, with some older build script changes cherry-picked back on and a couple other code fixes applied.


Norgg:
Current build errors: http://norgg.org/files/principia_build_errors_2015081515-Bourbaki-398-g4095795.html

base/macros.hpp
```diff
@@ -162,6 +162,13 @@ inline void noreturn() { std::exit(0); }
 #  define TEMPLATE template
 #endif
 
+// Use libc++ optional if possible
+#if PRINCIPIA_COMPILER_MSVC
+#  define OPTIONAL_HEADER "optional.hpp"
+#else
+#  define OPTIONAL_HEADER <experimental/optional>
+#endif
+
```
pleroy: I hacked our version of `optional` so that it can be included as `<experimental/optional>` and I prepared #678 to fix Principia.  So this macro should not be needed anymore, which is a good thing because the linter is not in love with it.  It would be good if you could revert the definition and usages of `OPTIONAL_HEADER` (and then git pull, just to be sure).

physics/forkable_body.hpp
```diff
@@ -340,7 +340,7 @@ void Forkable<Tr4jectory>::WriteSubTreeToMessage(
   for (auto const& pair : children_) {
     Instant const& fork_time = pair.first;
     Tr4jectory const& child = pair.second;
-    if (!last_instant || fork_time != last_instant) {
+    if (!last_instant || fork_time != *last_instant) {
```
pleroy: I'm a bit puzzled by this change, as it seems to me that http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2015/n4480.html#optional.synop implies that this should be legal without the `*`.  Are you sure that you are using the latest greatest version of `libwhateverc++`?  I am asking because your latest error log has many weird diagnostics related to `optional`.

install_deps.sh
```diff
@@ -28,32 +28,23 @@ CXX_FLAGS="-std=c++1y $LD_FLAGS"
...
 
-git clone https://github.com/Norgg/glog
+git clone "https://github.com/mockingbirdnest/glog"
 pushd glog
-./configure CC=clang CXX=clang++ CFLAGS="$C_FLAGS" CXXFLAGS="$CXX_FLAGS" LDFLAGS="$LD_FLAGS"
+./configure CC=clang CXX=clang++ CFLAGS="$C_FLAGS" CXXFLAGS="$CXX_FLAGS" LDFLAGS="$LD_FLAGS" LIBS="-lc++ -lc++abi"
 make -j8
 popd
 
-svn checkout http://googletest.googlecode.com/svn/trunk/ gtest
```
pleroy: I suspect that you'll need `gtest` if/when you want to compile the tests...

eggrobin: It's not there anymore though.

eggrobin: Same for gmock, that script is pulling our gmock rather than our googlemock at the moment.

eggrobin:
We took a look specifically at the following error:

```
./physics/discrete_trajectory.hpp:58:25: error: only virtual member functions can be marked 'override'
  ~DiscreteTrajectory() override;
                        ^~~~~~~~
```

and tried the following toy example in http://gcc.godbolt.org/:

```
template<typename T>
class Forkable {
 public:
  virtual ~Forkable() = default;
};

template<typename T>
class Trajectory : public Forkable<Trajectory<T>> {
 public:
  ~Trajectory() override;
};
```

It turns out that we could reproduce the error message by picking "x86 clang 3.0" as the compiler.  All versions after 3.2 appear to think that the code is legit.

Are you using a clang 3.0?  We'd be surprised if it was possible to compile Principia with that version: you might consider investing in a newer clang...

pleroy:
I have a minimal program that seems to exhibit the same problem (including some of the same line numbers in the library) at http://goo.gl/Eh5E7k.  The weird part is that it fails with clang 3.5.2 but works for more recent versions (you have to switch to `libstdc++` because this site doesn't have `libc++` for recent clangs).  I'd be curious to know how this example behaves with your clang.

eggrobin:
News from the linux VM front:
@pleroy 's latest minimum failing example exhibits the following behaviour with either of clang 3.7 or clang 3.6.
standard optional: fails with libc++, works with libstdc++.
our optional: fails with libc++, works with libstdc++.
It therefore seems that the difference lies in more indirection in libstdc++'s implementation of `map` or something like that, rather than in the `optional`s.

Errors from compiling the plugin with libstdc++, standard optional:
`clang++ -c -std=c++14 -stdlib=libstdc++ -O3 -g -fPIC -fexceptions -ferror-limit=0 -fno-omit-frame-pointer -Wall -Wpedantic
 -m64 -I. -Ideps/glog/src -Ideps/protobuf/src -Ideps/benchmark/include -Ideps/gmock/include -Ideps/gmock/gtest/include -Ideps/gmock -Ideps/gmock/gtest ksp_plugin/plugin.cpp -o k`
https://gist.github.com/eggrobin/5d3d093a528729df4fb4

eggrobin:
More news from the Linux VM front:
Using the VagrantFile in this pull request and pulling the latest changes from mockingbirdnest master, then reinstating the changes from #678, the plugin builds with either libc++ or libstdc++. Using either library's optional fails.

In the current state of [mockingbirdnest/Optional](github.com/mockingbirdnest/Optional), optional can be included as `#include <experimental/optional>`; if ` -Ideps/Optional` is passed to clang, it uses that implementation, otherwise it uses the stdlib's.

pleroy:
Looks all good, thanks for all the work.  I am going to merge as soon as Jenkins is happy.  We'll continue in a separate PR.

PR #684, opened 2015-10-03
No acceleration in DiscreteTrajectory, it moved to Ephemeris

pleroy:


physics/ephemeris.hpp
```diff
@@ -31,6 +31,13 @@ class Ephemeris {
   static_assert(Frame::is_inertial, "Frame must be inertial");
 
  public:
+  // The acceleration due to the engines of a massless body.
```
eggrobin: The acceleration of a massless body not due to gravity.

PR #690, opened 2015-10-15
Rigid motions for barycentric rotating frames.

pleroy:


physics/barycentric_rotating_dynamic_frame_body.hpp
```diff
@@ -0,0 +1,113 @@
...
+}
+
+template<typename InertialFrame, typename ThisFrame>
+void BarycentricRotatingDynamicFrame<InertialFrame, ThisFrame>::
+FromBasisOfInertialFrameToBasisOfThisFrame(
+    DegreesOfFreedom<InertialFrame> const& barycentre_degrees_of_freedom,
+    DegreesOfFreedom<InertialFrame> const& primary_degrees_of_freedom,
+    DegreesOfFreedom<InertialFrame> const& secondary_degrees_of_freedom,
+    not_null<Rotation<InertialFrame, ThisFrame>*> const rotation,
+    not_null<AngularVelocity<InertialFrame>*> const angular_velocity) {
```
eggrobin: We are not using `secondary_degrees_of_freedom` here, and in the implementation we could replace `barycentre_degrees_of_freedom` by `secondary_degrees_of_freedom` (assuming the contract `|barycentre_degrees_of_freedom| must be a convex combination of the two other degrees of freedom` is respected).

This is called only once, so it would be unneeded encapsulation, but since we want to add the 2-body corotating frame fixing one of the bodies (rather than the barycentre) in Brouwer, it could be reused there. Move this function out of the class instead, drop the `barycentre_degrees_of_freedom` parameter.

The name suggests that it returns some sort of linear map, change to `ComputeTwoBodyRotationalMotion` or something like that.

physics/barycentric_rotating_dynamic_frame_body.hpp
```diff
@@ -0,0 +1,113 @@
...
+      secondary_trajectory_->EvaluateDegreesOfFreedom(t, &secondary_hint_);
+  DegreesOfFreedom<InertialFrame> const barycentre_degrees_of_freedom =
+      Barycentre<InertialFrame, GravitationalParameter>(
+          {primary_degrees_of_freedom,
+           secondary_degrees_of_freedom},
+          {primary_->gravitational_parameter(),
+           secondary_->gravitational_parameter()});
+
+  Rotation<InertialFrame, ThisFrame>
+      from_basis_of_inertial_frame_to_basis_of_this_frame =
```
eggrobin: Just call it `rotation`, this is not just Hungarian, it is verbose Hungarian...

physics/barycentric_rotating_dynamic_frame_body.hpp
```diff
@@ -0,0 +1,113 @@
...
+  return ToThisFrameAtTime(t).Inverse();
+}
+
+template<typename InertialFrame, typename ThisFrame>
+Vector<Acceleration, ThisFrame>
+BarycentricRotatingDynamicFrame<InertialFrame, ThisFrame>::
+GeometricAcceleration(
+    Instant const& t,
+    DegreesOfFreedom<ThisFrame> const& degrees_of_freedom) const {
+  return Vector<Acceleration, ThisFrame>();
```
eggrobin: Could we at least have a `LOG(FATAL)` here?

physics/barycentric_rotating_dynamic_frame_body.hpp
pleroy: Changing the parameters and name.  Not moving out now, we'll see when we need it elsewhere.

physics/barycentric_rotating_dynamic_frame_body.hpp
pleroy: Done.

physics/barycentric_rotating_dynamic_frame_body.hpp
pleroy: Done.

PR #697, opened 2015-10-20
Test individually all 4 components of the acceleration

pleroy:


geometry/grassmann_body.hpp
```diff
@@ -236,7 +238,9 @@ inline Bivector<quantities::Product<LScalar, RScalar>, Frame> Commutator(
 template<typename Scalar, typename Frame, int rank>
 inline Multivector<double, Frame, rank> Normalize(
     Multivector<Scalar, Frame, rank> const& multivector) {
-  return multivector / multivector.Norm();
+  Scalar const norm = multivector.Norm();
+  CHECK_NE(Scalar(), norm);
+  return multivector / norm;
```
eggrobin: Why don't we let it NaN?

pleroy: Because I spent a lot of time tracking all the NaNs that would propagate to weird places.  You should be able to assume that the result actually has norm 1.  NaNs are not useful in this context as they break the postcondition.
For instance, there is a benchmark where this check fires.  It's better to know about it than to silently benchmark how the processor handles NaNs.

physics/barycentric_rotating_dynamic_frame_test.cpp
```diff
@@ -176,5 +209,265 @@ TEST_F(BarycentricRotatingDynamicFrameTest, Inverse) {
...
+  EXPECT_THAT(barycentre_dof.velocity(), Eq(Velocity<ICRFJ2000Equator>()));
+
+  EXPECT_CALL(mock_big_trajectory_, EvaluateDegreesOfFreedom(t, _))
+      .Times(2)
+      .WillRepeatedly(Return(big_dof));
+  EXPECT_CALL(mock_small_trajectory_, EvaluateDegreesOfFreedom(t, _))
+      .Times(2)
+      .WillRepeatedly(Return(small_dof));
+  {
+    // The acceleration is centripete + tangential.
```
eggrobin: centripetal

PR #711, opened 2015-11-04
In game Frenet trihedron

eggrobin:
Some more factoring might be welcome later (we compute the rotation from the Frenet frame three times), but this works.

### DIFF
--- a/ksp_plugin/plugin.cpp
+++ b/ksp_plugin/plugin.cpp
@@ -980,7 +980,7 @@ RenderedTrajectory<World> Plugin::RenderTrajectory(
 Vector<double, World> Plugin::FromVesselFrenetFrame(
     Vessel const& vessel,
     not_null<RenderingFrame*> const rendering_frame,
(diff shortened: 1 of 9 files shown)


ksp_plugin/plugin.cpp
```diff
@@ -492,24 +492,25 @@ FrameField<World> Plugin::Navball(
...
+}
 
-  // The vector with coordinates {1, 0, 0} in the Frenet frame of the
-  // vessel's free-falling trajectory in the given |rendering_frame|, converted
-  // to |WorldSun| coordinates.
-  return Identity<WorldSun, World>()(
-      BarycentricToWorldSun()(
-          rendering_frame->FromThisFrameAtTime(time).orthogonal_map()(
-              from_frenet_frame_to_rendering_frame(tangent))));
+Vector<double, World> Plugin::VesselNormal(
```
pleroy: Please add a test for the normal and binormal.

PR #743, opened 2015-11-26
Reference frame selector in a separate window

eggrobin:


ksp_plugin_adapter/ksp_plugin_adapter.cs
```diff
@@ -14,6 +14,8 @@ namespace ksp_plugin_adapter {
                                          GameScenes.TRACKSTATION})]
 public partial class PrincipiaPluginAdapter : ScenarioModule {
```
pleroy: Maybe no blank line here?

ksp_plugin_adapter/reference_frame_selector.cs
```diff
@@ -38,7 +40,29 @@ class ReferenceFrameSelector {
...
+    }
+    UnityEngine.GUI.skin = old_skin;
+  }
+
+  public void RenderWindow() {
+    var old_skin = UnityEngine.GUI.skin;
+    UnityEngine.GUI.skin = null;
+    if (show_selector_) {
+      window_rectangle_ = UnityEngine.GUILayout.Window(
+                              id         : 2,
```
pleroy: Don't comment this `2`, it's totally clear what it does.

PR #755, opened 2015-12-05
C# should not own things

eggrobin:
The selector no longer owns its navigation frame, it is (explicitly, safely) owned by the plugin instead.
Two transfers of ownership occur at creation, but locally in the C#.

### DIFF
--- a/ksp_plugin/interface.cpp
+++ b/ksp_plugin/interface.cpp
@@ -447,6 +447,12 @@ NavigationFrame* principia__NewBarycentricRotatingNavigationFrame(
                                            secondary_index).release());
 }
 
+// TODO(phl): Journalling.
+void principia__SetPlottingFrame(Plugin* const plugin,
+                                 NavigationFrame** const navigation_frame) {
+  CHECK_NOTNULL(plugin)->SetPlottingFrame(TakeOwnership(navigation_frame));
+}
+
 void principia__DeleteNavigationFrame(
     NavigationFrame** const navigation_frame) {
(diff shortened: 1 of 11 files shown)


ksp_plugin/interface.cpp
```diff
@@ -679,24 +681,24 @@ XYZ principia__VesselTangent(Plugin const* const plugin,
                              char const* const vessel_guid,
                              NavigationFrame* const navigation_frame) {
```
pleroy: Remove `navigation_frame` from the methods that don't need it.

ksp_plugin_adapter/reference_frame_selector.cs
```diff
@@ -194,18 +189,18 @@ public enum FrameType {
...
   private void ResetFrame() {
     on_change_();
-    Interface.DeleteNavigationFrame(ref frame_);
+    IntPtr navigation_frame = IntPtr.Zero;
     switch (frame_type) {
       case FrameType.BODY_CENTRED_NON_ROTATING:
-        frame_ = plugin_.NewBodyCentredNonRotatingNavigationFrame(
-                     selected_celestial_.flightGlobalsIndex);
+        navigation_frame = plugin_.NewBodyCentredNonRotatingNavigationFrame(
+                               selected_celestial_.flightGlobalsIndex);
```
pleroy: Bad indent.

ksp_plugin/interface.cpp
eggrobin: If I do that I have to mess with the `Run` functions in the journal.

PR #762, opened 2015-12-06
Fix bugs related to journal replay

pleroy:
* Enter the out pointers in the pointer map.
* Remove the pointers from the map when they are deleted.

### DIFF
--- a/journal/profiles.cpp
+++ b/journal/profiles.cpp
@@ -20,10 +20,20 @@ template<typename T>
 void Insert(not_null<Player::PointerMap*> const pointer_map,
             std::uint64_t const address,
             T* const pointer) {
-  auto const inserted_pointer = const_cast<std::remove_cv<T>::type*>(pointer);
+  void* const inserted_pointer =
+      static_cast<void*>(const_cast<std::remove_cv<T>::type*>(pointer));
   auto inserted = pointer_map->emplace(address, inserted_pointer);
   if (!inserted.second) {
     CHECK_EQ(inserted.first->second, inserted_pointer);
+  }
+}
+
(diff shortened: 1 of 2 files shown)


journal/profiles.cpp
```diff
@@ -20,13 +20,23 @@ template<typename T>
 void Insert(not_null<Player::PointerMap*> const pointer_map,
             std::uint64_t const address,
             T* const pointer) {
-  auto const inserted_pointer = const_cast<std::remove_cv<T>::type*>(pointer);
+  void* const inserted_pointer =
+      static_cast<void*>(const_cast<std::remove_cv<T>::type*>(pointer));
```
eggrobin: Do we still need the `const_cast` then?

journal/profiles.cpp
```diff
@@ -596,6 +607,8 @@ void SetPlottingFrame::Run(Message const& message,
   auto* navigation_frame = DeserializePointer<NavigationFrame*>(
                                *pointer_map, in.navigation_frame());
   ksp_plugin::principia__SetPlottingFrame(plugin, &navigation_frame);
+  // TODO(phl): This may delete the previous plotting frame, which should thus
+  // be removed from |pointer_map|.
```
eggrobin: No, the `pointer_map` doesn't represent the memory (we don't care about the memory), it represents the objects that the caller of the DLL owns directly. Thus `SetPlottingFrame` "deletes" (consumes) *its argument*.

journal/profiles.cpp
pleroy: Yes, of course.

journal/profiles.cpp
pleroy: Done.

PR #764, opened 2015-12-06
Make the plugin tests cover the journal

eggrobin:


pleroy:
If there is no solution there is no problem.

PR #767, opened 2015-12-10
Egglipses

pdn4kd:
Eclipse testing first pass -- U1234 for the 1950-04-2 Lunar event. Also a Python support file for getting Julian Days.


eggrobin:

eggrobin:
retest this please

eggrobin:
Lint errors found by `cpplint`:
```
physics\eclipse_test.cpp(24):  Lines should very rarely be longer than 100 characters  [whitespace/line_length] [4]
physics\eclipse_test.cpp(24):  At least two spaces is best between code and comments  [whitespace/comments] [2]
physics\eclipse_test.cpp(32):  Do not indent within a namespace  [runtime/indentation_namespace] [4]
physics\eclipse_test.cpp(33):  protected: should be indented +1 space inside class EclipseTest  [whitespace/indent] [3]
physics\eclipse_test.cpp(36):  Line ends in whitespace.  Consider deleting these extra spaces.  [whitespace/end_of_line] [4]
physics\eclipse_test.cpp(37):  Lines should be <= 80 characters long  [whitespace/line_length] [2]
physics\eclipse_test.cpp(42):  Line ends in whitespace.  Consider deleting these extra spaces.  [whitespace/end_of_line] [4]
physics\eclipse_test.cpp(169):  You don't need a ; after a }  [readability/braces] [4]
physics\eclipse_test.cpp(44):  Lines should very rarely be longer than 100 characters  [whitespace/line_length] [4]
physics\eclipse_test.cpp(48):  Lines should very rarely be longer than 100 characters  [whitespace/line_length] [4]
physics\eclipse_test.cpp(48):  At least two spaces is best between code and comments  [whitespace/comments] [2]
physics\eclipse_test.cpp(53):  Line ends in whitespace.  Consider deleting these extra spaces.  [whitespace/end_of_line] [4]
physics\eclipse_test.cpp(75):  Lines should be <= 80 characters long  [whitespace/line_length] [2]
physics\eclipse_test.cpp(77):  Lines should very rarely be longer than 100 characters  [whitespace/line_length] [4]
physics\eclipse_test.cpp(77):  At least two spaces is best between code and comments  [whitespace/comments] [2]
physics\eclipse_test.cpp(81):  Lines should very rarely be longer than 100 characters  [whitespace/line_length] [4]
physics\eclipse_test.cpp(91):  Lines should be <= 80 characters long  [whitespace/line_length] [2]
physics\eclipse_test.cpp(93):  Lines should very rarely be longer than 100 characters  [whitespace/line_length] [4]
physics\eclipse_test.cpp(93):  At least two spaces is best between code and comments  [whitespace/comments] [2]
physics\eclipse_test.cpp(97):  Lines should be <= 80 characters long  [whitespace/line_length] [2]
physics\eclipse_test.cpp(107):  Lines should be <= 80 characters long  [whitespace/line_length] [2]
physics\eclipse_test.cpp(109):  Lines should very rarely be longer than 100 characters  [whitespace/line_length] [4]
physics\eclipse_test.cpp(109):  At least two spaces is best between code and comments  [whitespace/comments] [2]
physics\eclipse_test.cpp(113):  Lines should be <= 80 characters long  [whitespace/line_length] [2]
physics\eclipse_test.cpp(123):  Lines should be <= 80 characters long  [whitespace/line_length] [2]
physics\eclipse_test.cpp(125):  Lines should very rarely be longer than 100 characters  [whitespace/line_length] [4]
physics\eclipse_test.cpp(125):  At least two spaces is best between code and comments  [whitespace/comments] [2]
physics\eclipse_test.cpp(129):  Lines should very rarely be longer than 100 characters  [whitespace/line_length] [4]
physics\eclipse_test.cpp(157):  Lines should very rarely be longer than 100 characters  [whitespace/line_length] [4]
physics\eclipse_test.cpp(167):  Lines should very rarely be longer than 100 characters  [whitespace/line_length] [4]
physics\eclipse_test.cpp(168):  Lines should very rarely be longer than 100 characters  [whitespace/line_length] [4]
physics\eclipse_test.cpp(170):  Line ends in whitespace.  Consider deleting these extra spaces.  [whitespace/end_of_line] [4]
physics\eclipse_test.cpp(171):  Namespace should be terminated with "// namespace physics"  [readability/namespace] [5]
physics\eclipse_test.cpp(171):  At least two spaces is best between code and comments  [whitespace/comments] [2]
physics\eclipse_test.cpp(173):  Namespace should be terminated with "// namespace principia"  [readability/namespace] [5]
physics\eclipse_test.cpp(173):  At least two spaces is best between code and comments  [whitespace/comments] [2]
```

astronomy/ephemeris.py
```diff
@@ -0,0 +1,31 @@
...
+
+# Last Lunar Eclipse will be 2048-01-01 
+# P1 = 03:52:39 UT
+# U1 = 05:05:17 UT
+# U2 = 06:24:27
+# U3 = 07:20:23
+# U4 = 08:39:33
+# P4 = 09:52:05
+
+P1U1234P4 = ['1950-04-02T18:10:49', '1950-04-02T19:09:19', '1950-04-02T20:30:38', '1950-04-02T20:57:33', '1950-04-02T22:18:54', '1950-04-02T23:17:21']
```
pleroy: Please keep lines shorters than 80 chars.

astronomy/ephemeris.py
```diff
@@ -0,0 +1,31 @@
...
+eclipseTimes = Time(P1U1234P4, format='isot', scale='utc')
+et2 = eclipseTimes.tdb
+csv = open('eclipsetimes.csv', 'w')
+csv.write(str(et2.jd))
+csv.write('\n')
+
+P1U1234P4 = ['2048-01-01T03:52:39', '2048-01-01T05:05:17', '2048-01-01T06:24:27', '2048-01-01T07:20:23', '2048-01-01T08:39:33', '2048-01-01T09:52:05']
```
pleroy: Same, should be <= 80 chars.

physics/eclipse_test.cpp
```diff
@@ -0,0 +1,167 @@
+#include "gmock/gmock.h"
+#include "gtest/gtest.h"
+
```
pleroy: Please remove this blank line and alphabetize all the includes.

physics/eclipse_test.cpp
```diff
@@ -0,0 +1,167 @@
...
+#include "quantities/elementary_functions.hpp"
+#include "geometry/grassmann.hpp"
+#include "testing_utilities/numerics.hpp"
+
+namespace principia {
+
+using astronomy::ICRFJ2000Equator;
+using geometry::JulianDate;
+using integrators::McLachlanAtela1992Order5Optimal;
+using quantities::si::Minute;
```
pleroy: Please alphabetize these 5 lines (the ones for `quantities::si`).

physics/eclipse_test.cpp
```diff
@@ -0,0 +1,167 @@
...
+using quantities::si::Metre;
+using quantities::si::Nano;
+using quantities::si::Milli;
+using quantities::si::Kilo;
+using quantities::ArcCos;
```
pleroy: Please move before the `using`s for `quantities::si`.  In general we put less nested declarations (e.g., `quantities::ArcCos`) before more nested ones (e.g., `quantities::si::Anything`).

physics/eclipse_test.cpp
```diff
@@ -0,0 +1,167 @@
...
+using testing_utilities::AbsoluteError;
+using ::testing::Gt;
```
pleroy: Please alphabetize these 3 lines (the ones for `::testing`).

physics/eclipse_test.cpp
```diff
@@ -0,0 +1,167 @@
...
+
+TEST_F(EclipseTest, Dummy) {
+  auto ephemeris = solar_system_1950_.MakeEphemeris(
+      McLachlanAtela1992Order5Optimal<Position<ICRFJ2000Equator>>(),
+      45 * Minute, 5 * Milli(Metre));
+
+  ephemeris->Prolong(JulianDate(2433374.5));  // Prolong until date of eclipse
+                                              // (Eclipse was 1950-04-02 but JD
+                                              // is 1950-04-03:00:00:00)
+  // pass body to Ephemeris.trajectory
```
pleroy: Comments should start with a capital letter and end with a period.  Also the wording is a bit odd confusing since it doesn't use `Ephemeris::trajectory`.  Maybe you meant something like "Extract the bodies from the ephemeris" or some such.

physics/eclipse_test.cpp
```diff
@@ -0,0 +1,167 @@
...
+  auto sun = ephemeris->bodies()[solar_system_1950_.index("Sun")];
```
pleroy: This is technically correct, but it works because the indices returned by `index()` are suitable for using in the vector returned by `bodies()`.  This coupling makes me a bit uncomfortable as I could see these two classes changing independently in the future.  The function `SolarSystem::massive_body` is what you want here, thus:
`solar_system_1950_.massive_body(*ephemeris, "Sun")`.

Also, we try to use constants whenever possible, so the type should be `auto const`.

Same comments for the two lines below.

physics/eclipse_test.cpp
```diff
@@ -0,0 +1,167 @@
...
+  auto earth = ephemeris->bodies()[solar_system_1950_.index("Earth")];
+  auto moon = ephemeris->bodies()[solar_system_1950_.index("Moon")];
+
+  // Massive_body eventually needs radius information. Or non-hardcoded data
+  // pulled from
+  // https://github.com/mockingbirdnest/Principia/blob/master/astronomy/gravity_model.proto.txt
+  auto r_sun = 696000.0 * Kilo(Metre);
```
pleroy: `auto const`, I presume that the radius of the Earth won't change ;-)

physics/eclipse_test.cpp
```diff
@@ -0,0 +1,167 @@
...
+
+  ephemeris->Prolong(JulianDate(2433374.5));  // Prolong until date of eclipse
+                                              // (Eclipse was 1950-04-02 but JD
+                                              // is 1950-04-03:00:00:00)
+  // pass body to Ephemeris.trajectory
+  auto sun = ephemeris->bodies()[solar_system_1950_.index("Sun")];
+  auto earth = ephemeris->bodies()[solar_system_1950_.index("Earth")];
+  auto moon = ephemeris->bodies()[solar_system_1950_.index("Moon")];
+
+  // Massive_body eventually needs radius information. Or non-hardcoded data
```
pleroy: `Massive_body` should be spelled `MassiveBody`.

More importantly, could you please open an issue for adding the radius to `MassiveBody`?  (And then you can reference the issue in this comment.)

physics/eclipse_test.cpp
```diff
@@ -0,0 +1,167 @@
...
+  // pulled from
+  // https://github.com/mockingbirdnest/Principia/blob/master/astronomy/gravity_model.proto.txt
+  auto r_sun = 696000.0 * Kilo(Metre);
+  auto r_earth = 6378.1363 * Kilo(Metre);
+  auto r_moon = 1738.0 * Kilo(Metre);
+
+  // Dates are TDB Julian Day for 1948-04-02
+  auto P1 = JulianDate(2433374.25788409);  // 18:10:49 UT
```
pleroy: `auto const` for all these dates.

Also, is the terminology `U1`, `P1`, etc. well-known in astronomy?  If so could you add a link to some page that describes it (e.g. on Wikipedia or some other reference site).

physics/eclipse_test.cpp
```diff
@@ -0,0 +1,167 @@
...
+  auto U1 = JulianDate(2433374.29850909);  // 19:09:19
+  auto U2 = JulianDate(2433374.354979);    // 20:30:38
+  auto U3 = JulianDate(2433374.37367113);  // 20:57:33
+  auto U4 = JulianDate(2433374.43016419);  // 22:18:54
+  auto P4 = JulianDate(2433374.47075446);  // 23:17:21
+
+  // Lunar eclipse -- U1 (time is start of umbral phase)
+  auto q_sun = ephemeris->trajectory(sun)->EvaluatePosition(U1, nullptr);
```
pleroy: `auto const` for all the positions.  Also, the `nullptr` at the end is a bit mysterious, so it's best to comment it thus:
``
EvaluatePosition(U1, /*hint=*/nullptr);
``

Same for the following two lines.

physics/eclipse_test.cpp
```diff
@@ -0,0 +1,167 @@
...
+  auto q_moon = ephemeris->trajectory(moon)->EvaluatePosition(U1, nullptr);
+  auto q_earth = ephemeris->trajectory(earth)->EvaluatePosition(U1, nullptr);
+
+  // check body angles at target time
+  // Earth/Sun lineup
+  auto alpha = ArcSin((r_sun - r_earth) / (q_sun - q_earth).Norm());
```
pleroy: When using Greek names we normally use the Greek alphabet instead of the English transliteration.  Thus `α`, `β`, `γ` instead of `alpha`, `beta`, `gamma`.

This being said, `α` is not very informative.  To the best of the my understanding this is half the [aperture of the cone](https://en.wikipedia.org/wiki/Cone#Further_terminology) made by the Sun and the Earth.  A more explicit name would be good, even if it's longer, e.g., `half_sun_earth_aperture` (or possibly drop `half_` and multiply by 2 everywhere, this is a test, we don't care about its performance).

physics/eclipse_test.cpp
```diff
@@ -0,0 +1,167 @@
...
+  auto q_U14 =
+      q_earth + Normalize(q_earth - q_sun) * (r_earth + r_moon) / Sin(alpha);
+  // Earth/Moon lineup for U14
+  auto gamma = ArcCos(InnerProduct(q_U14 - q_earth, q_U14 - q_moon) /
+                      ((q_U14 - q_moon).Norm() * (q_U14 - q_earth).Norm()));
+  // Do the angles work?
+  LOG(ERROR) << alpha;
+  LOG(ERROR) << gamma;
```
pleroy: It's best to avoid `LOG` statements in tests, they pollute the output.  What you probably want is to log the angles when the expectation below fails.  This can be done by streaming a message in the `EXPECT_THAT` macro:
```
EXPECT_THAT(...) << "alpha = " << alpha << ", gamma = " << gamma;
```

eggrobin: Note that for concision and maintainability, we have a macro that will expand `alpha` to
```cpp
"alpha" << ": " << alpha
```
namely [`NAMED(alpha)`](https://github.com/mockingbirdnest/Principia/blob/master/base/macros.hpp#L161), thus @pleroy's suggestion becomes
```cpp
EXPECT_THAT(...) << NAMED(alpha) << ", " << NAMED(beta);
```

physics/eclipse_test.cpp
```diff
@@ -0,0 +1,167 @@
...
+  EXPECT_THAT(AbsoluteError(alpha, gamma),
+              AllOf(Lt(1.0 * Milli(Radian)), Gt(1.0 * Nano(Radian))));
+
+  // Lunar eclipse -- U2 (time is start of full eclipse)
```
pleroy: Hmm.  By now the alert reader notices a pattern: it seems that you are doing the exact same computation four times for `U1` to `U4` (although I'll confess that I am not sure why there is a distinction between `β` and `γ`).  It would be much nicer to factor out this code in a helper method located in class EclipseTest.  Something like:
```
class EclipseTest : public testing::Test {
  ...
  void CheckEclipseTime(Instant const& t) {
    auto const q_sun = ...;
    ...
  }
  ...
};
```
and then:
```
CheckEclipseTime(U1);
CheckEclipseTime(U2);
...
```

physics/eclipse_test.cpp
```diff
@@ -0,0 +1,167 @@
...
+  // U4 = 08:39:33
+  // P4 = 09:52:05
+  // 2469076.66235167  2469076.71279148  2469076.76776833  2469076.80661092
+  // 2469076.86158778  2469076.91195815
+  // etimes = {2469076.66235167, 2469076.71279148, 2469076.76776833,
+  // 2469076.80661092, 2469076.86158778, 2469076.91195815};
+};
+
+}  // namespace physics
+
```
pleroy: Remove blank line.

physics/eclipse_test.cpp
```diff
@@ -0,0 +1,167 @@
...
+  // U1 = 05:05:17 UT
+  // U2 = 06:24:27
+  // U3 = 07:20:23
+  // U4 = 08:39:33
+  // P4 = 09:52:05
+  // 2469076.66235167  2469076.71279148  2469076.76776833  2469076.80661092
+  // 2469076.86158778  2469076.91195815
+  // etimes = {2469076.66235167, 2469076.71279148, 2469076.76776833,
+  // 2469076.80661092, 2469076.86158778, 2469076.91195815};
+};
```
pleroy: The semicolon here is unnecessary, I think.

physics/eclipse_test.cpp
pdn4kd: Wikipedia uses it in various locations, with an explanation on their [lunar eclipse page.](https://en.wikipedia.org/wiki/Lunar_eclipse#Timing) I'm not sure how vulnerable this terminology is to [citogenesis](https://xkcd.com/978/), though. [The NASA eclipse PDFs](http://eclipse.gsfc.nasa.gov/eclipse.html) that get passed around moderately often also uses it.

physics/eclipse_test.cpp
pdn4kd: Well, as long as we aren't simulating the nuclear and/or heat transfer processes in enough detail that variation is possible.

physics/eclipse_test.cpp
```diff
@@ -0,0 +1,175 @@
...
+TEST_F(EclipseTest, Dummy) {
+  auto ephemeris = solar_system_1950_.MakeEphemeris(
+      McLachlanAtela1992Order5Optimal<Position<ICRFJ2000Equator>>(),
+      45 * Minute, 5 * Milli(Metre));
+
+  ephemeris->Prolong(
+      JulianDate(2433374.5));  // Prolong just past date of eclipse
+                               // (Eclipse was 1950-04-02 but JD
+                               // is 1950-04-03:00:00:00).
+  // Pass body to Ephemeris.trajectory.
```
pleroy: This comment doesn't match the code and should be fixed.

physics/eclipse_test.cpp
```diff
@@ -0,0 +1,175 @@
...
+  // Dates are TDB Julian Day for 1948-04-02.
+  auto const P1 = JulianDate(2433374.25788409);  // 18:10:49 UT
+  auto const U1 = JulianDate(2433374.29850909);  // 19:09:19
+  auto const U2 = JulianDate(2433374.354979);    // 20:30:38
+  auto const U3 = JulianDate(2433374.37367113);  // 20:57:33
+  auto const U4 = JulianDate(2433374.43016419);  // 22:18:54
+  auto const P4 = JulianDate(2433374.47075446);  // 23:17:21
+
+  // Lunar eclipse -- U1 (time is start of umbral phase).
+  auto q_sun = ephemeris->trajectory(sun)->EvaluatePosition(U1, nullptr);
```
pleroy: Please comment all the `nullptr`s as follows: `/*hint=*/nullptr`.

physics/eclipse_test.cpp
```diff
@@ -0,0 +1,175 @@
...
+  auto half_earth_moon_aperture =
+      ArcCos(InnerProduct(q_U14 - q_earth, q_U14 - q_moon) /
+             ((q_U14 - q_moon).Norm() * (q_U14 - q_earth).Norm()));
+  // Do the angles work?
+  EXPECT_THAT(AbsoluteError(half_sun_earth_aperture, half_earth_moon_aperture),
+              AllOf(Lt(1.0 * Milli(Radian)), Gt(1.0 * Nano(Radian))))
+      << NAMED(half_sun_earth_aperture) << ", "
+      << NAMED(half_earth_moon_aperture);
+
+  // Lunar eclipse -- U2 (time is start of full eclipse).
```
pleroy: I think you missed the comment where I was asking for the code to be restructured with a utility method instead of replicated 4 times.  Basically the lines 71-92 and the lines 93-114, 115-136 and 137-158 are very similar and should be implemented using a common function.

physics/eclipse_test.cpp
```diff
@@ -0,0 +1,128 @@
...
+    ephemeris->Prolong(
+        JulianDate(2433374.5));  // Prolong just past date of eclipse
+                                 // (Eclipse was 1950-04-02 but JD
+                                 // is 1950-04-03 00:00:00).
+    auto const sun = solar_system_1950_.massive_body(*ephemeris, "Sun");
+    auto const earth = solar_system_1950_.massive_body(*ephemeris, "Earth");
+    auto const moon = solar_system_1950_.massive_body(*ephemeris, "Moon");
+
+    auto const q_sun = ephemeris->trajectory(sun)->EvaluatePosition(
+        current_time, /* hint= */ nullptr);
```
eggrobin: No spaces within the comment, nor after it, thus
```cpp
/*hint=*/nullptr
```
We are currently not consistently following this convention throughout the code (sometimes we use `arg  // arg`, sometimes `arg /*arg*/`), but this new syntax is [checked by clang-tidy](http://clang.llvm.org/extra/clang-tidy/checks/misc-argument-comment.html) (which we might start using at some point).

pleroy:
Thank you for your patience during the review.  Merging.

PR #781, opened 2015-12-30
Penumbrae

pdn4kd:
Penumbral eclipses are now checkable, and several additional eclipses (through 1952) added. The 2048 eclipse is excluded because running the ephemeris out that far requires >1.5 GiB of RAM/swap.

### DIFF
--- a/physics/eclipse_test.cpp
+++ b/physics/eclipse_test.cpp
@@ -17,18 +17,23 @@ using geometry::JulianDate;
 using geometry::Sign;
 using integrators::McLachlanAtela1992Order5Optimal;
 using quantities::ArcCos;
+using quantities::si::Day;
 using quantities::si::Kilo;
 using quantities::si::Metre;
 using quantities::si::Milli;
 using quantities::si::Minute;
 using quantities::si::Nano;
-using quantities::si::Second;
 using testing_utilities::AbsoluteError;
 using ::testing::AllOf;
 using ::testing::Gt;
 using ::testing::Lt;
 
 namespace physics {
+
+namespace {
+Sign const U14 = Sign(1);
+Sign const U23 = Sign(-1);
+}  // namespace
 
 class EclipseTest : public testing::Test {
  protected:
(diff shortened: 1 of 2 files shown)


physics/eclipse_test.cpp
```diff
@@ -38,16 +39,14 @@ class EclipseTest : public testing::Test {
...
     auto ephemeris = solar_system_1950_.MakeEphemeris(
         McLachlanAtela1992Order5Optimal<Position<ICRFJ2000Equator>>(),
         45 * Minute, 5 * Milli(Metre));
 
-    ephemeris->Prolong(
-        JulianDate(2433374.5));  // Prolong just past date of eclipse
-                                 // (Eclipse was 1950-04-02 but JD
-                                 // is 1950-04-03 00:00:00).
+    ephemeris->Prolong(current_time +
+                       86400 * Second);  // Prolong 1 day past date of eclipse.
```
pleroy: Please use `1 * Day` (`Day` is defined in `si.hpp`).  You won't believe how easy it is to type `84600` instead of the correct value, and that's impossible to detect at code review time.

physics/eclipse_test.cpp
```diff
@@ -90,38 +89,143 @@ class EclipseTest : public testing::Test {
...
   }
 
+  void CheckLunarPenumbralEclipse(Instant const& current_time,
+                                  Sign const moon_offset_sign) {
+    auto ephemeris = solar_system_1950_.MakeEphemeris(
+        McLachlanAtela1992Order5Optimal<Position<ICRFJ2000Equator>>(),
+        45 * Minute, 5 * Milli(Metre));
+
+    ephemeris->Prolong(current_time +
+                       86400 * Second);  // Prolong 1 day past date of eclipse.
```
pleroy: Same, `1 * Day`.

physics/eclipse_test.cpp
```diff
@@ -90,38 +89,143 @@ class EclipseTest : public testing::Test {
...
+    EXPECT_THAT(AbsoluteError(penumbral_half_aperture, earth_moon_angle),
+                AllOf(Lt(1.0 * Milli(Radian)), Gt(1.0 * Nano(Radian))))
+        << NAMED(penumbral_half_aperture) << ", " << NAMED(earth_moon_angle)
+        << ", " << NAMED(current_time);
+  }
+
   SolarSystem<ICRFJ2000Equator> solar_system_1950_;
 };
 
 TEST_F(EclipseTest, Dummy) {
```
pleroy: This test is becoming biggish, it would be nice to split it.  This makes analysis/debugging easier when it fails.  For instance:

```
TEST_F(EclipseTest, Year1950) {
  // The eclipses for 1950 go here.
}

TEST_F(EclipseTest, Year1951) {
  // For 1951.
}

// etc.
```

This will require that you move `U14` and `U23` to a common place.  I suggest an unnamed namespace after line 31 thus:

```
namespace {
  Sign const U14 = Sign(1);
  Sign const U23 = Sign(-1);
}  // namespace
```

pleroy:
Looks good in general, just a few nitpicks.

pleroy:
Looks good, merging, thanks.

PR #789, opened 2016-01-02
Makefile changes to generate profile sources

Norgg:
Also includes a couple minor fixes to the linux build.

### DIFF
--- a/base/macros.hpp
+++ b/base/macros.hpp
@@ -123,7 +123,7 @@ inline void noreturn() { std::exit(0); }
 #if PRINCIPIA_COMPILER_CLANG    ||  \
     PRINCIPIA_COMPILER_CLANG_CL ||  \
     PRINCIPIA_COMPILER_GCC
-#  define FORCE_INLINE [[gnu::always_inline]]  // NOLINT(whitespace/braces)
+#  define FORCE_INLINE [[gnu::always_inline]] inline  // NOLINT(whitespace/braces)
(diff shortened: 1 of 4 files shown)


journal/method_body.hpp
```diff
@@ -1,6 +1,7 @@
 #pragma once
 
 #include "journal/method.hpp"
+#include "journal/recorder.hpp"
```
pleroy: Lint says: `Found C++ system header after other header. Should be: method_body.h, c system, c++ system, other.`

This include should go after the include of glog/logging.h at line 8.

base/macros.hpp
```diff
@@ -123,7 +123,7 @@ inline void noreturn() { std::exit(0); }
 #if PRINCIPIA_COMPILER_CLANG    ||  \
     PRINCIPIA_COMPILER_CLANG_CL ||  \
     PRINCIPIA_COMPILER_GCC
-#  define FORCE_INLINE [[gnu::always_inline]]  // NOLINT(whitespace/braces)
+#  define FORCE_INLINE [[gnu::always_inline]] inline  // NOLINT(whitespace/braces)
```
pleroy: Lint says: 'Lines should be <= 80 characters long`.

Just remove `(whitespace/braces)` (keeping only `NOLINT`) and it should fit.

pleroy:
Thanks for doing this change.  Just two lint errors, otherwise it looks fine.

PR #793, opened 2016-01-03
Code generation for interface.hpp

pleroy:
@Norgg: note that this generates a new file (`ksp_plugin/interface.generated.h`) so I suspect that the Make file will need to be updated to represent this new dependency.


tools/generate_profiles.cpp
```diff
@@ -44,6 +47,18 @@ void GenerateProfiles() {
...
+
+  std::ofstream interface_generated_h(ksp_plugin / "interface.generated.h");
+  CHECK(interface_generated_h.good());
+  interface_generated_h << kWarning;
+  for (auto const& cpp_interface_type_declaration :
+          processor.GetCppInterfaceTypeDeclarations()) {
+    interface_generated_h << cpp_interface_type_declaration;
+  }
+  for (auto const& cpp_interface_method_declaration :
+          processor.GetCppInterfaceMethodDeclarations()) {
```
eggrobin: badindentsn

pleroy: Done.

PR #799, opened 2016-01-05
Better eclipse performance (probably)

pdn4kd:
Ephemeris are now only computed as-needed. Also the ephemeris and solar system are made static. (Performance on real hardware vs my VM is ambiguous for the last one.)


pleroy:
Thanks for doing this, the speedup is really noticeable: on my machine this test goes down from 88.2 s to 5.5 s.

There are two lint failures that you inherited from @Norgg but I'll fix them after merging.

PR #803, opened 2016-01-09
Bisection

eggrobin:
1-ULP generic bisection.

### DIFF
--- a/geometry/sign.hpp
+++ b/geometry/sign.hpp
@@ -21,6 +21,8 @@ class Sign {
   bool Negative() const;
   bool Positive() const;
 
+  bool operator==(Sign const& other) const;
+
   void WriteToMessage(not_null<serialization::Sign*> const message) const;
   static Sign ReadFromMessage(serialization::Sign const& message);
 
--- a/geometry/sign.hpp
+++ b/geometry/sign.hpp
@@ -22,6 +22,7 @@ class Sign {
   bool Positive() const;
 
   bool operator==(Sign const& other) const;
+  bool operator!=(Sign const& other) const;
 
   void WriteToMessage(not_null<serialization::Sign*> const message) const;
   static Sign ReadFromMessage(serialization::Sign const& message);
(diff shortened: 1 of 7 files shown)


geometry/sign.hpp
```diff
@@ -21,6 +21,8 @@ class Sign {
   bool Negative() const;
   bool Positive() const;
 
+  bool operator==(Sign const& other) const;
```
pleroy: Should we also have `!=`?

numerics/root_finders.hpp
```diff
@@ -0,0 +1,17 @@
+#pragma once
+
+namespace principia {
+namespace numerics {
+
+// Approximates a root of |f| between |lower_bound| and |upper_bound| by
```
pleroy: Mention that lower_ and upper_bound must have values of different signs and the function must be monotonic.

numerics/root_finders_body.hpp
```diff
@@ -0,0 +1,46 @@
...
+
+template<typename Argument, typename Function>
+Argument Bisect(Function f,
+                Argument const& lower_bound,
+                Argument const& upper_bound) {
+  using Value = decltype(f(lower_bound));
+  Value const zero{};
+  Value f_upper = f(upper_bound);
+  Value f_lower = f(lower_bound);
+  CHECK(f_lower > zero &&  zero > f_upper || f_lower < zero &&  zero < f_upper);
```
pleroy: Move inside the loop?

geometry/sign.hpp
eggrobin: Done.

numerics/root_finders.hpp
eggrobin: That's not it, but documented the contract.

numerics/root_finders_body.hpp
eggrobin: wat.

PR #812, opened 2016-01-14
Tests for interface flight plan

pleroy:


testing_utilities/make_not_null_body.hpp
```diff
@@ -0,0 +1,15 @@
+#pragma once
+
+#include "testing_utilities/make_not_null.hpp"
+
+namespace principia {
+namespace testing_utilities {
+
+// Returns a bogus not-null value of the right type.  Dereferencing it is UB.
```
eggrobin: Do not replicate comments between the declaration and definition, thus remove this one.

PR #818, opened 2016-01-18
Normalize files to start with BOM and single newline.

eggrobin:
The normalization is performed by the following commands, stored in the makefile:
Remove BOM and initial empty lines:
```sh
for f in `ls */*.hpp && ls */*.cpp`
do
  awk 'NR==1{sub(/^\xef\xbb\xbf/,"")}1' $f | awk NF{p=1}p > $f.nobom
  mv $f.nobom $f
done
```
Add BOM and newline:
```sh
for f in `ls */*.hpp && ls */*.cpp`
do
  awk 'NR==1{sub(/^/,"\xef\xbb\xbf\n")}1' $f > $f.withbom
  mv $f.withbom $f
done
```


pleroy:
This should not modify generated files (nrwhall). The generator should be fixed instead.

PR #822, opened 2016-01-23
Change the eclipse test to include a first-order approximation of the atmosphere 

pleroy:


pleroy:
Nope, these numbers create sizeable errors in the Sputnik1-to-Sputnik2 test.

PR #824, opened 2016-01-23
Use Danjon's modelling for the atmosphere

pleroy:
It gives much better results than our cheezy bisection.  Kudos to @pdn4kd for explaining to us that the atmosphere really matters a lot for lunar eclipses.


astronomy/eclipse_test.cpp
```diff
@@ -250,28 +214,29 @@ class EclipseTest : public testing::Test {
         current_time - bisection_interval,
         current_time + bisection_interval);
     EXPECT_EQ(Sign(actual_contact_time - current_time),
-              Sign(time_error)) << NAMED(time_error);
+              Sign(time_error)) << NAMED(actual_contact_time - current_time);
```
eggrobin: Isn't the expected, actual backwards?
`EXPECT_EQ` is very confusing because of that in the midst of `EXPECT_THAT`s, preference for `EXPECT_THAT(actual, Eq(expected))` (also, this sounds like what we'd want is a `SameSign` matcher).

eggrobin: Why are we testing the sign of the difference and its absolute value separately anyway (especially since we end up remultiplying them for logging)?

pleroy: As, discussed, it's pretty messy to check that the error has the right sign and the right order of magnitude in a single statement:
```
AnyOf(AllOf(Lt(2), Gt(1)), AllOf(Gt(-2), Lt(-1)))
```
The logging can be simplified, though.  Also changed the sign check to use `EXPECT_THAT`.

PR #842, opened 2016-02-07
Fix a crash in the deserialization of manoeuvres

pleroy:
Also change the interface to make sure that the journal may be replayed.


ksp_plugin/interface_flight_plan.cpp
```diff
@@ -52,9 +52,41 @@ Burn GetBurn(NavigationManœuvre const& manœuvre) {
   Velocity<Frenet<NavigationFrame>> const Δv =
       manœuvre.Δv() == Speed() ? Velocity<Frenet<NavigationFrame>>()
                                : manœuvre.Δv() * manœuvre.direction();
+
+  // When building the parameters, make sure that the "optional" fields get a
+  // deterministic default.
+  NavigationFrameParameters parameters;
+  parameters.centre_index = -1;
+  parameters.primary_index = -1;
+  parameters.secondary_index = -1;
```
eggrobin: Can we put those as in-class initializers for NavigationFrameParameters?

pleroy: Opened #843.

PR #877, opened 2016-02-12
Maximal step count

eggrobin:
A simple implementation, working towards #846 (still needs to be used in the ephemeris).
Do we want to have a callback or a status of sorts to report early termination? In the flight plan, we will need to know it occured in order not to compute the remaining segments.

### DIFF
--- a/integrators/ordinary_differential_equations.hpp
+++ b/integrators/ordinary_differential_equations.hpp
@@ -2,6 +2,7 @@
 #ifndef PRINCIPIA_INTEGRATORS_ORDINARY_DIFFERENTIAL_EQUATIONS_HPP_
 #define PRINCIPIA_INTEGRATORS_ORDINARY_DIFFERENTIAL_EQUATIONS_HPP_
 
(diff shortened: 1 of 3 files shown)


integrators/ordinary_differential_equations.hpp
```diff
@@ -92,6 +93,9 @@ struct AdaptiveStepSize {
   // In both cases, the new step size is chosen so as to try and make the result
   // of the next call to |tolerance_to_error_ratio| close to |safety_factor|.
   ToleranceToErrorRatio tolerance_to_error_ratio;
+  // If |max_steps|, integration will stop after |*max_steps| even if it has not
+  // reached |t_final|.
+  std::experimental::optional<std::int64_t> max_steps;
```
pleroy: I'd prefer to make `max_steps` a plain int, initialized here to `std::numeric_limits<std::int64_t>::max()`.  That would be more efficient and readable than having to check of the optional is armed.

PR #887, opened 2016-02-16
Revert "Add mean_radius to MassiveBody"

eggrobin:
Rolling this back to put the radius in `RotatingBody` instead, as discussed.


eggrobin:
OBE.

Issue #901, opened 2016-02-21
Custom marshaling for UTF-8 and returned unowned strings


eggrobin:
At the moment we have to return unowned strings as `IntPtr` and convert them with `Marshal.PtrToStringAnsi`. Moreover, on Windows `Marshal.PtrToStringAnsi` will not use UTF-8 (it will with Mono on Linux).
While waiting for [.NET to add a UTF-8 marshaller](https://github.com/dotnet/coreclr/issues/1012) and for that to percolate into whichever antiquated Mono KSP is using, we could try use libgit2sharp's custom marshalers (MIT licensed) which provide everything we need.
See [Utf8Marshaler.cs](https://github.com/libgit2/libgit2sharp/blob/3010f8373313a47e9a889e8b3e61a30cd2eb359b/LibGit2Sharp/Core/Utf8Marshaler.cs), [EncodingMarshaler.cs](https://github.com/libgit2/libgit2sharp/blob/3010f8373313a47e9a889e8b3e61a30cd2eb359b/LibGit2Sharp/Core/EncodingMarshaler.cs).

pleroy:
Fixed by #909.

PR #909, opened 2016-02-25
A more robust implementation of out parameters in the interface

eggrobin:
Allow out parameters of types which don't have 0 as a value, in preparation for #901.
Require that out parameters be [`TriviallyCopyable`](http://en.cppreference.com/w/cpp/concept/TriviallyCopyable), since otherwise they cannot be truly out (note: could we get away with just trivial assignments and destructor?).
Enforce that they be truly out by stomping the memory before passing them.

The use of `make_unique` means we also require a (not necessarily trivial) default constructor, but this could easily be circumvented using stuff from from [`<new>`](http://en.cppreference.com/w/cpp/header/new), or `malloc` and a `unique_ptr` with `free` as its custom deleter, if needed.

See the changes to the generated code at 5f631b4.


tools/journal_proto_processor.cpp
```diff
@@ -481,10 +488,17 @@ void JournalProtoProcessor::ProcessRequiredField(
     }
     field_cxx_type_[descriptor] += "*";
```
pleroy: We should have two wrapper lambdas `field_cxx_mode_fn_` and `field_cs_mode_fn_`.

tools/journal_proto_processor.cpp
```diff
@@ -481,10 +488,17 @@ void JournalProtoProcessor::ProcessRequiredField(
...
     field_cxx_arguments_fn_[descriptor] =
         [](std::string const& identifier) -> std::vector<std::string> {
           return {"&" + identifier};
         };
+    } else {
+      field_cxx_arguments_fn_[descriptor] =
+        [](std::string const& identifier) -> std::vector<std::string> {
+        return { identifier + ".get()" };
+      };
+    }
```
pleroy: OBE.

tools/journal_proto_processor.cpp
```diff
@@ -606,24 +620,55 @@ void JournalProtoProcessor::ProcessInOut(
...
-          ";\n";
+      if (Contains(out_, field_descriptor)) {
+        // For an out parameter, we need to allocate some uninitializeed
+        // storage.  The |make_unique| below requires that
+        // |field_cxx_type_[field_descriptor]| be default-constructible.
+        // For more generality, we could initialize the |unique_ptr| to
+        // |reinterpret_cast<T*>(std::operator new(sizeof(T)))|.
+        cxx_run_body_prolog_[descriptor] +=
+            "  auto const " + run_local_variable + " = std::make_unique<" +
+            field_cxx_direct_type_[field_descriptor] + ">();\n";
```
pleroy: No need for a `unique_ptr<>`, this should generate something like:
```
  T t;
```
where `T` is `field_cxx_type_`.

tools/journal_proto_processor.cpp
```diff
@@ -606,24 +620,55 @@ void JournalProtoProcessor::ProcessInOut(
...
+        cxx_run_body_prolog_[descriptor] +=
+            "  std::memset(" + run_local_variable + ".get(), 0xAB, sizeof (" +
+            field_cxx_direct_type_[field_descriptor] + "));\n";
+        cxx_run_body_prolog_[descriptor] +=
+            "  static_assert(\n    std::is_trivially_copyable<" +
+            field_cxx_direct_type_[field_descriptor] +
+            ">::value,\n    \"out parameter |" +
+            run_local_variable + "| of |" +
+            field_descriptor->containing_type()->containing_type()->name() +
+            "| must have a trivially copyable type\");\n";
```
pleroy: Let's remove the stomping and the `static_assert` and have tighter `static_assert`s in the interface.

tools/journal_proto_processor.cpp
```diff
@@ -606,24 +620,55 @@ void JournalProtoProcessor::ProcessInOut(
...
+      } else {
+        std::string const cxx_run_field_deserializer_getter =
+            cxx_run_field_getter;
```
pleroy: This variable is no longer needed, it was just here to isolate a `?:` expression.

tools/journal_proto_processor.cpp
```diff
@@ -606,24 +620,55 @@ void JournalProtoProcessor::ProcessInOut(
...
       cxx_run_body_epilog_[descriptor] +=
           field_cxx_deleter_fn_[field_descriptor](cxx_run_field_getter);
     }
     if (Contains(field_cxx_inserter_fn_, field_descriptor)) {
+      bool const is_out = Contains(out_, field_descriptor) &&
+                          !Contains(in_out_, field_descriptor);
       cxx_run_body_epilog_[descriptor] +=
           field_cxx_inserter_fn_[field_descriptor](
               ToLower(name) + "." + field_descriptor_name + "()",
+              (is_out ? "*" : "") +
```
pleroy: OBE.

tools/journal_proto_processor.hpp
```diff
@@ -182,6 +182,9 @@ class JournalProtoProcessor {
   // declaration, in a typedef, etc.
   std::map<FieldDescriptor const*, std::string> field_cs_type_;
   std::map<FieldDescriptor const*, std::string> field_cxx_type_;
+  // For out or in out parameters, the type of the parameter without the
+  // indirection due to the mode.
+  std::map<FieldDescriptor const*, std::string> field_cxx_direct_type_;
```
pleroy: OBE.

eggrobin:
Changes to generated code now at 2df1b36.

eggrobin:
Changes to the generated code now at e6c02c5.

tools/journal_proto_processor.cpp
```diff
@@ -601,20 +624,19 @@ void JournalProtoProcessor::ProcessInOut(
                 std::back_inserter(cxx_run_arguments_[descriptor]));
     }
     if (must_generate_code) {
```
pleroy: Merge this `if` with the previous one.

tools/journal_proto_processor.hpp
```diff
@@ -149,6 +149,18 @@ class JournalProtoProcessor {
                                      std::string const& expr2)>>
       field_cxx_inserter_fn_;
 
+  // For all fields, a lambda that takes a C#/C++ parameter type as stored in
+  // |field_cs_type_|, repsectively |field_cxx_type_|, and adds a mode to it,
```
pleroy: respectively

tools/journal_proto_processor.hpp
```diff
@@ -149,6 +149,18 @@ class JournalProtoProcessor {
                                      std::string const& expr2)>>
       field_cxx_inserter_fn_;
 
+  // For all fields, a lambda that takes a C#/C++ parameter type as stored in
```
pleroy: parameter or member

tools/journal_proto_processor.hpp
```diff
@@ -149,6 +149,18 @@ class JournalProtoProcessor {
...
+  // |field_cs_type_|, repsectively |field_cxx_type_|, and adds a mode to it,
+  // yielding |T| for an in parameter in either language, |ref T| for an in out
```
pleroy: I'd drop everything from "yielding" to the end, no point in leaking implementation details.

PR #910, opened 2016-02-25
Add NormalizeOrZero and call it.

### LINKED ISSUE #854: NormalizeOrZero
The flight plan interface currently relies on `Normalize` on null vectors, which results in a `CHECK` failure in debug (and returns a vector of `NaN`s otherwise).

pleroy:
Fix #854.  This is journal-incompatible.

### DIFF
--- a/geometry/grassmann.hpp
+++ b/geometry/grassmann.hpp
@@ -186,7 +186,7 @@ template<typename Scalar, typename Frame, int rank>
 Multivector<double, Frame, rank> Normalize(
     Multivector<Scalar, Frame, rank> const& multivector);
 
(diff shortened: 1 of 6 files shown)


geometry/grassmann.hpp
```diff
@@ -186,6 +186,11 @@ template<typename Scalar, typename Frame, int rank>
 Multivector<double, Frame, rank> Normalize(
     Multivector<Scalar, Frame, rank> const& multivector);
 
+// Same as above, but returns its argument if it is the zero multivector.
```
eggrobin: Types don't match.

PR #915, opened 2016-02-27
Marshal in and unowned out strings

eggrobin:
But not returned strings nor owned out strings.
Changes to the generated code at c896474.


serialization/journal.proto
```diff
@@ -587,10 +587,9 @@ message GetVersion {
     optional GetVersion extension = 5078;
   }
   message Out {
-    required fixed64 build_date = 1 [(pointer_to) = "char const",
-                                     (is_produced) = true];
-    required fixed64 version = 2 [(pointer_to) = "char const",
-                                  (is_produced) = true];
+    reserved 1, 2;
```
pleroy: I wouldn't put a `reserved`, I would just reuse the tags 1 and 2: we don't care that much about journal compatibility.

Issue #911, opened 2016-02-26
Parameterize the number of steps of the flow functions


pleroy:
It is currently fixed at 1000 which confused the benchmarks.

pleroy:
Fixed by #919.

PR #917, opened 2016-02-28
Wrap the parameters of the Ephemeris functions into parameter classes

pleroy:
No semantic changes at this point.  In particular, `max_steps` is still hard-wired to 1000.  Partially addresses #911.

### DIFF
--- a/ksp_plugin/flight_plan.cpp
+++ b/ksp_plugin/flight_plan.cpp
@@ -120,11 +120,9 @@ bool FlightPlan::SetFinalTime(Instant const& final_time) {
   }
 }
 
-void FlightPlan::SetTolerances(
-    Length const& length_integration_tolerance,
-    Speed const& speed_integration_tolerance) {
-  adaptive_parameters_.SetTolerances(length_integration_tolerance,
-                                     speed_integration_tolerance);
+void FlightPlan::SetAdaptiveStepParameters(
+    Ephemeris<Barycentric>::AdaptiveStepParameters const& adaptive_parameters) {
+  adaptive_parameters_ = adaptive_parameters;
   RecomputeSegments();
 }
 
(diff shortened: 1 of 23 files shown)


ksp_plugin/flight_plan.cpp
```diff
@@ -124,8 +123,8 @@ bool FlightPlan::SetFinalTime(Instant const& final_time) {
 void FlightPlan::SetTolerances(
```
eggrobin: Make this `SetAdaptiveParameters` (there's no reason to keep the integrator the same if we recompute everything).

physics/ephemeris.hpp
```diff
@@ -42,15 +42,52 @@ class Ephemeris {
...
+
+    template<typename T>
+    void WriteToMessage(not_null<T*> const t) const;
+
+    template<typename T>
+    static AdaptiveStepParameters ReadFromMessage(T const& t);
+
+   private:
+    // This will refer to a static object returned by a factory.
+    AdaptiveStepSizeIntegrator<NewtonianMotionEquation> const& integrator_;
```
eggrobin: This makes the class non-assignable, which is a bit annoying for something that's just a parameter container.We tend to use pointers in those cases (the exception to our reference conventions is that we may take the address of a reference to a static object, this is also somewhat easier to enforce).
Make it a `not_null<AdaptiveStepSizeIntegrator<NewtonianMotionEquation>>`.

physics/ephemeris.hpp
```diff
@@ -42,15 +42,52 @@ class Ephemeris {
...
+   public:
+    // The |length_| and |speed_integration_tolerance|s are used to compute the
+    // |tolerance_to_error_ratio| for step size control.
+    AdaptiveStepParameters(
+        AdaptiveStepSizeIntegrator<NewtonianMotionEquation> const& integrator,
+        Length const& length_integration_tolerance,
+        Speed const& speed_integration_tolerance);
+
+    void SetTolerances(Length const& length_integration_tolerance,
+                       Speed const& speed_integration_tolerance);
```
eggrobin: Once this is assignable we don't need this function (1 use) anymore, remove.

ksp_plugin/flight_plan.cpp
pleroy: Done.

physics/ephemeris.hpp
pleroy: Done.

Issue #844, opened 2016-02-07
Do not serialize the recomputable trajectories


pleroy:
That is, everything except history and prolongation.

PR #926, opened 2016-03-13
Do not serialize the trajectories of the FlightPlan

pleroy:
Change the serialization of DiscreteTrajectory so that the tree and pointers to it are serialized and deserialized together.  Add a test to exercise pre-Bunyakovsky compatibility.  This addresses a good chunk of #844.


ksp_plugin/flight_plan.cpp
```diff
@@ -166,19 +163,31 @@ std::unique_ptr<FlightPlan> FlightPlan::ReadFromMessage(
...
-  }
-  for (int i = 0; i < message.manoeuvre_size(); ++i) {
-    auto const& manoeuvre = message.manoeuvre(i);
-    flight_plan->manœuvres_.push_back(
-        NavigationManœuvre::ReadFromMessage(manoeuvre, ephemeris));
-    flight_plan->manœuvres_[i].set_coasting_trajectory(
-        flight_plan->segments_[2 * i]);
+
+  bool const is_pre_буняко́вский = message.segment_size() > 0;
+  if (is_pre_буняко́вский) {
```
eggrobin: remove the combining accents (everywhere)

serialization/physics.proto
```diff
@@ -1,4 +1,4 @@
-syntax = "proto2";
+﻿syntax = "proto2";
```
eggrobin: I think when run from visual studio (as opposed to from msbuild itself called from jenkins's shell) the BOM in the protos causes issues. Set encoding to UTF-8 without BOM, see the comment on #924.

serialization/ksp_plugin.proto
```diff
@@ -107,6 +108,9 @@ message Plugin {
 message Vessel {
   required MasslessBody body = 1;
   optional FlightPlan flight_plan = 4;
+  optional Trajectory history = 6;  // required
```
eggrobin: That's a tad confusing, maybe required since Буняковский? or required since #926? or since (SHA1)?

ksp_plugin/physics_bubble.cpp
```diff
@@ -249,7 +249,7 @@ void PhysicsBubble::WriteToMessage(
     current_->centre_of_mass->WriteToMessage(
         full_state->mutable_centre_of_mass());
     current_->centre_of_mass_trajectory->WriteToMessage(
-        full_state->mutable_centre_of_mass_trajectory());
+        full_state->mutable_centre_of_mass_trajectory(), {});
```
eggrobin: Comment the `{}` (several places).

ksp_plugin/flight_plan.cpp
pleroy: Done.

serialization/physics.proto
pleroy: Done.

serialization/ksp_plugin.proto
pleroy: No point in putting the history in the source code, it's much easier to obtain it through "blame".  However, it's important to state that even though this field is "optional" (for compatibility) it really is required in new code.  We have comments like that everywhere.

PR #927, opened 2016-03-14
Serialization of the parameter classes of Ephemeris at the right place

pleroy:
This contributes to #844.


serialization/ksp_plugin.proto
```diff
@@ -38,16 +38,18 @@ message FlightPlan {
...
-  // we stop serializing recomputable data.
+  optional Ephemeris.AdaptiveStepParameters
+      adaptive_step_parameters = 11;  // required
+
+  // Pre-Буняковский.
+  optional Quantity length_integration_tolerance = 4;
+  optional Quantity speed_integration_tolerance = 5;
+  optional AdaptiveStepSizeIntegrator integrator = 6;
+  repeated Trajectory.Pointer segment = 7;
   optional int32 anomalous_segments = 9 [default = 0];
```
eggrobin: Why keep it? just make it reserved, we're not reading it, even when reading pre-Буняковский messages.

Issue #938, opened 2016-03-22
Make the maximal step count for the flight plan customizable


eggrobin:
Requested by Scott Manley.

eggrobin:
We just want to make `max_steps` and possibly the tolerances customizable, not the integrator, thus FlightPlan will need a `GetAdaptiveStepParameters` and `Ephemeris::AdaptiveStepParameters` will need a `set_max_steps`. In addition, accessors (`get_max_steps`, `get_length_integration_tolerance`) will be needed to display the right numbers in the UI.

pleroy:
We should have a "plain old struct" representations of the `...StepParameters` classes for interchange through the interface (this requires new protobufs).

eggrobin:
Probably (although we probably don't want to pass the integrator choice through the interface yet, that's another kettle of fish), but frankly I'm not sure why `Ephemeris::AdaptiveStepParameters` can't be a struct (with public members) itself, since we're going to need to expose its members one way or another.

pleroy:
Reassigning to @eggrobin for the UI part.

PR #948, opened 2016-03-25
Change SetAdaptiveStepParameters to preserve the invariants of FlightPlan

### LINKED ISSUE #940: Correctness of SetAdaptiveStepParameters
What happens if we decrease `max_steps` or the tolerances? It seems like the invariant `anomalous_segments_ <= 2` can be violated, perhaps `SetAdaptiveStepSizeParameters` should return a bool.

pleroy:
Fix #940.

### DIFF
--- a/ksp_plugin/flight_plan.cpp
+++ b/ksp_plugin/flight_plan.cpp
@@ -120,10 +120,18 @@ bool FlightPlan::SetFinalTime(Instant const& final_time) {
   }
 }
 
-void FlightPlan::SetAdaptiveStepParameters(
+bool FlightPlan::SetAdaptiveStepParameters(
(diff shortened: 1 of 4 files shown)


ksp_plugin/flight_plan.cpp
```diff
@@ -166,7 +174,7 @@ std::unique_ptr<FlightPlan> FlightPlan::ReadFromMessage(
             AdaptiveStepSizeIntegrator<
                 Ephemeris<Barycentric>::NewtonianMotionEquation>::
                     ReadFromMessage(message.integrator()),
-            /*max_steps=*/1000,
+            /*max_steps=*/10000,
```
eggrobin: Why? We'll deal with that properly in #938. If there's a good reason, it deserves a comment.

pleroy: Oops.

Issue #761, opened 2015-12-06
Predictions should not be recomputed at every frame 


eggrobin:
Currently the prediction is recomputed at every frame, i.e., at 50 Hz. This is probably too often.

However, in order to properly take advantage of doing it less often, we'd to parallelize, which is a whole other rabbit hole (although at least MSVC now has `shared_lock`).

eggrobin:
A random thought while going through the issues: when the initial state doesn't change due to unpredictable factors (engines/collisions), we definitely shouldn't recompute the prediction often (the only change in the solution there is due to differences in integration methods between the history and prediction).
When things like engines are at play, it might make sense to (empirically) estimate the dependence on initial conditions and recompute based on that. The work on pile-ups should make the behaviour of unpacked vessels in free fall identical to that of packed vessels, so the situation should nicely split between actually no thrust and under thrust.

pleroy:
Finally fixed by #2092.

PR #956, opened 2016-03-26
Change the Vessel class so that an initialized vessel always has a prediction

pleroy:
This is in preparation for addressing #761.


ksp_plugin/plugin.cpp
```diff
@@ -445,10 +419,18 @@ void Plugin::SetPredictionLength(Time const& t) {
 
 void Plugin::SetPredictionLengthTolerance(Length const& l) {
   prediction_parameters_.set_length_integration_tolerance(l);
+  for (auto const& pair : vessels_) {
```
eggrobin: This probably means that what we want is to have vessel-specific prediction parameters, and that eventually SetPredictionLengthTolerance should take a vessel. #957.

ksp_plugin/vessel.hpp
```diff
@@ -55,26 +57,27 @@ class Vessel {
   virtual not_null<Celestial const*> parent() const;
   virtual void set_parent(not_null<Celestial const*> const parent);
 
-  // Requires |is_initialized()|.
+  // Require |is_initialized()|.
```
eggrobin: Why?

pleroy: Because it applies to 3 functions.

Issue #953, opened 2016-03-25
Apsis altitudes


eggrobin:
Markers may be complicated, but it appears that altitudes are vital. We can give an interval for the time of the apsis by the intermediate value theorem, and we can then do a silly third-order interpolation or something to get an idea of the actual time and altitude.

dylanede:
Should this not stay open since apsides for predictions aren't yet in?

pleroy:
Apsides for predictions were introduced in #984 (arguably that pull request should be referenced here).  Do you have evidence that they don't work?  If so, it would be good if you could post a screenshot and a save (as a gist, preferably).

dylanede:
Hmmm, I'll double check my end then, maybe I was doing something wrong. Thanks for the quick reply!

dylanede:
Ah, it's just that the apsides only appear when the relevant body is the center of the frame of reference. Makes sense - wouldn't work any other way.

eggrobin:
> Makes sense - wouldn't work any other way.

Yes, we had briefly tried drawing the apsides with respect to both bodies in the barycentric rotating frame, but this got confusing:
- there were too many apsides (on low Moon orbit in the EMB frame, you ended up with two Moon apsides and two Earth apsides, and in a single Earth-Moon L1 orbit you would get four Moon apsides and four Earth apsides);
- since the bodies are not fixed in the frame (they move back and forth along the line between them), the apsides appeared at positions unrelated from the position of the map view body, showing a collision on a comfortable periapsis, or the reverse.

As a result, we decided to go with apsides only for the body whose centre is fixed in the frame (if any). This means other body-centric frames can be useful, e.g. the body-centred frame fixing the direction of its parent (#995). For the Earth-Moon system this would be somewhat like the EMB frame, but more suited to Moon approaches; the Jupiter-fixed, Sun-direction fixed frame is sometimes used for plots of trajectories of Jupiter probes, see Galileo below; the Earth-fixed, Sun-direction fixed frame would help with heliosynchronous trajectories.
![](http://lasp.colorado.edu/home/galileo/files/2012/08/OrbitTrajectoriesPetalPlot.jpg)

PR #966, opened 2016-03-28
Compute the apsides of a trajectory

pleroy:
This is the physical computation for addressing #953.

### DIFF
--- a/physics/ephemeris_body.hpp
+++ b/physics/ephemeris_body.hpp
@@ -15,7 +15,6 @@
 #include "geometry/grassmann.hpp"
 #include "geometry/r3_element.hpp"
 #include "numerics/hermite3.hpp"
-#include "numerics/root_finders.hpp"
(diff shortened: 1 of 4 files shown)


physics/ephemeris_body.hpp
```diff
@@ -590,6 +593,106 @@ ComputeGravitationalAccelerationOnMassiveBody(
...
+      }
+      if (valid_extrema != 1) {
+        // Something went wrong when finding the extrema of
+        // |squared_distance_approximation|. Use a linear interpolation of
+        // |squared_distance_derivative| instead.
+        apsis_time = Barycentre<Instant, Variation<Square<Length>>>(
+            {time, *previous_time},
+            {*previous_squared_distance_derivative,
+             -squared_distance_derivative});
+      }
```
eggrobin: Making an issue to cover that with some degenerate decadent orbit.

PR #976, opened 2016-04-01
Stronger SFINAE conditions

eggrobin:
With some not-entirely-trivial helper traits.

### DIFF
--- a/journal/method.hpp
+++ b/journal/method.hpp
@@ -29,23 +29,27 @@ namespace journal {
 //                    not_null<Player::PointerMap*> const pointer_map);
 //  };
 
+namespace internal {
+
 template<typename T>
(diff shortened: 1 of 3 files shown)


journal/method.hpp
```diff
@@ -29,31 +29,57 @@ namespace journal {
 //                    not_null<Player::PointerMap*> const pointer_map);
 //  };
 
+template<typename T>
```
pleroy: Put this in an `internal` namespace to make it clear that this is not part of the API.

journal/method.hpp
```diff
@@ -29,31 +29,57 @@ namespace journal {
...
+using specialize = void;
```
pleroy: `void_if_exists`

Issue #869, opened 2016-02-11
Configurable trajectory colours


eggrobin:
As well as dashing.

eggrobin:
Considering the upcoming fading-in-the-past stock trajectories (in 1.1), the following seems to make sense (discussed with @ferram4).
History fading in the past (opaque in the present, getting more transparent as it gets more distant) With #942, this includes the celestials.
Predictions in a solid colour. With #942, this includes the celestials.
Flight plan dashed (much like stock's manœuvre predictions).

PR #983, opened 2016-04-04
Colours and style

eggrobin:
Progress on #869: dashing and fading, switching the flight plan to a more visible colour. Configurable colours still need to be added.

### DIFF
--- a/ksp_plugin_adapter/gl_lines.cs
+++ b/ksp_plugin_adapter/gl_lines.cs
@@ -7,6 +7,12 @@ namespace principia {
 namespace ksp_plugin_adapter {
 
 internal static class GLLines {
+  public enum Style {
+    SOLID,
+    DASHED,
+    FADED,
+  }
+
   public static void Draw(Action line_vertices) {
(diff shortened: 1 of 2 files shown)


ksp_plugin_adapter/gl_lines.cs
```diff
@@ -90,19 +96,28 @@ internal static class GLLines {
...
+      for (int i = 0;
+           !trajectory_iterator.IteratorAtEnd();
+           trajectory_iterator.IteratorIncrement(), ++i) {
         Vector3d current_point = (Vector3d)trajectory_iterator.IteratorGetXYZ();
         if (previous_point.HasValue) {
-          AddSegment(previous_point.Value,
-                     current_point,
-                     hide_behind_bodies : true);
+          if (style == Style.FADED) {
+            colour.a = (float)i / (float)size;
```
pleroy: Could we not start at 0 but maybe at 20% or some such?  Otherwise the beginning is just invisible which is not useful.

PR #1013, opened 2016-05-01
Normalize version header

eggrobin:


generate_version_header.sh
```diff
@@ -1,12 +1,13 @@
...
+
 namespace principia {
 namespace base {
 
-  char const kBuildDate[] = \"%%DATE%%\";
-  char const kVersion[] =
-      u8\"%%VERSION%%\";
+char const kBuildDate[] = \"%%DATE%%\";
+char const kVersion[] =
+    u8\"%%VERSION%%\";
```
pleroy: On the previous line?

PR #1029, opened 2016-05-10
Only persist minimal data for the Ephemeris and recompute the rest

### LINKED ISSUE #1002: Don't persist the Ephemeris, recompute it
We need to keep all intermediate states in double precision between segments.  Then persist the "forgot before" state.

We could also have a RecordIO cache but that's more work and it may not even be faster.

pleroy:
Fixes #1002.

### DIFF
--- a/physics/ephemeris_body.hpp
+++ b/physics/ephemeris_body.hpp
@@ -725,10 +725,10 @@ void Ephemeris<Frame>::WriteToMessage(
     }
     checkpoints_.front().system_state.WriteToMessage(
         message->mutable_last_state());
+    t_max().WriteToMessage(message->mutable_t_max());
   }
   parameters_.WriteToMessage(message->mutable_fixed_step_parameters());
   fitting_tolerance_.WriteToMessage(message->mutable_fitting_tolerance());
-  t_max().WriteToMessage(message->mutable_t_max());
   LOG(INFO) << NAMED(message->SpaceUsed());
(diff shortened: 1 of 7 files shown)


physics/ephemeris_body.hpp
```diff
@@ -712,12 +711,24 @@ void Ephemeris<Frame>::WriteToMessage(
...
+      trajectories_[i]->WriteToMessage(message->add_trajectory(),
+                                       checkpoints[i]);
+    }
+    checkpoints_.front().system_state.WriteToMessage(
+        message->mutable_last_state());
   }
   parameters_.WriteToMessage(message->mutable_fixed_step_parameters());
   fitting_tolerance_.WriteToMessage(message->mutable_fitting_tolerance());
-  last_state_.WriteToMessage(message->mutable_last_state());
+  t_max().WriteToMessage(message->mutable_t_max());
```
eggrobin: only write if there's a checkpoint.

serialization/physics.proto
```diff
@@ -87,7 +87,8 @@ message Ephemeris {
   repeated ContinuousTrajectory trajectory = 2;
   required Quantity fitting_tolerance = 5;
   required SystemState last_state = 6;
-  optional FixedStepParameters fixed_step_parameters = 7;  // required
+  optional FixedStepParameters fixed_step_parameters = 7;  // required.
+  optional Point t_max = 8;  // required.
```
eggrobin: Not required.

PR #1091, opened 2016-07-13
UT1

eggrobin:
20 000 lines of code, as well as today's episode of *Confusing C++*, starring "```{foo(), bar();}``` is not equivalent to ```{foo(); bar();}```" or rather, dealing with expressions, "```(foo(), bar())``` is not equivalent to ```([]{foo();}(), bar())```".


astronomy/date_body.hpp
```diff
@@ -941,28 +960,35 @@ constexpr std::array<int, (2017 - 1972) * 2> leap_seconds = {{
...
                    ModernUTCMinusTAI(Date::Calendar(utc_date.year(), 1, 1));
 }
 
 constexpr bool IsValidModernUTC(DateTime const& date_time) {
   return !date_time.time().is_leap_second() ||
-         (date_time.date().month() == 6 &&
-          leap_seconds[(date_time.date().year() - 1972) * 2] == +1) ||
-         (date_time.date().month() == 12 &&
-          leap_seconds[(date_time.date().year() - 1972) * 2 + 1] == +1);
+         LeapSecond(date_time.date().year(), date_time.date().month()) == +1;
```
pleroy: What happens with negative leap seconds?

astronomy/date_body.hpp
```diff
@@ -1041,6 +1067,148 @@ constexpr bool IsValidStretchyUTC(DateTime const& utc) {
...
+  return TimeScale(utc()) + ut1_minus_utc;
+}
+
+constexpr quantities::Time EOPC04Entry::ut1_minus_tai() const {
+  return utc().date().year() >= 1972
+             ? ut1_minus_utc + ModernUTCMinusTAI(utc().date())
+             : ut1_minus_utc - TAIMinusStretchyUTC(utc());
+}
+
+// NOTE(egg): these have to be defined here, because they depend on internal
```
pleroy: I would actually like to move the entrie c02/c04 to their own file.  This file is getting too big anyway and the modelling of UT1 feels somewhat different from the arithmetic on dates and times.

astronomy/date_body.hpp
```diff
@@ -1041,6 +1067,148 @@ constexpr bool IsValidStretchyUTC(DateTime const& utc) {
...
+// classes defined above.
+#include "astronomy/experimental_eop_c02.generated.h"
+#include "astronomy/eop_c04.generated.h"
+
+// Returns the last entry in [begin, begin + size[ whose UT1 is less than or
+// equal to the given |ut1|.  The range [begin, begin + size[ must be sorted by
+// UT1.
+// We have to use |begin| and |size| rather than |begin| and |end| because
+// otherwise the MSVC complains about undefinedness of |end - begin| (even
```
pleroy: the MSVC compiler

astronomy/date_body.hpp
```diff
@@ -1041,6 +1067,148 @@ constexpr bool IsValidStretchyUTC(DateTime const& utc) {
...
+// though Intellisense is fine with it).
+constexpr ExperimentalEOPC02Entry const* LookupUT1(
+    quantities::Time const& ut1,
+    ExperimentalEOPC02Entry const* begin,
+    std::ptrdiff_t const size) {
+  return CHECKING(size > 0,
+                  size == 1
+                      ? CHECKING(begin->ut1_mjd <= mjd(ut1), begin)
+                      : (begin + size / 2)->ut1_mjd <= mjd(ut1)
```
pleroy: This better be seriously tested.  Binary searches are always wrong.

astronomy/date_body.hpp
```diff
@@ -1041,6 +1067,148 @@ constexpr bool IsValidStretchyUTC(DateTime const& utc) {
...
+  // TODO(egg): figure out whether using the divided difference of the
+  // |p->ut1_minus_tai()|s leads to less catastrophic cancellation than using
+  // the divided difference of the |DateTimeAsUTC(p->utc())|s.
+  return FromTAI(ut1 - (low->ut1_minus_tai() +
+                        (ut1 - low->ut1()) * ((low + 1)->ut1_minus_tai() -
+                                              low->ut1_minus_tai()) /
+                            ((low + 1)->ut1() - low->ut1())));
+}
+
+constexpr Instant InterpolatedExperimentalEOPC02(
```
pleroy: Comment this method.

astronomy/date_body.hpp
```diff
@@ -1041,6 +1067,148 @@ constexpr bool IsValidStretchyUTC(DateTime const& utc) {
...
+    quantities::Time const& ut1) {
+  return FromTAI(ut1 - (low->ut1_minus_tai +
+                        (mjd(ut1) - low->ut1_mjd) *
+                            ((low + 1)->ut1_minus_tai - low->ut1_minus_tai) /
+                            ((low + 1)->ut1_mjd - low->ut1_mjd)));
+}
+
+// Linear interpolation in the segment between the UT1s |low->ut1_mjd| and
+// |eop_c04.front().ut1()|, used to get continuity when switching between the
+// series.
```
pleroy: The API would be easier to understand if `eop_c04.front()` was passed to the method as a parameter, instead of being accessed as a global.

astronomy/eop_c04.awk
```diff
@@ -0,0 +1,17 @@
+BEGIN {
```
pleroy: Should there be a comment briefly commenting the format?

astronomy/experimental_eop_c02.awk
```diff
@@ -0,0 +1,16 @@
+BEGIN {
```
pleroy: Same.

astronomy/date_body.hpp
eggrobin: Let's assume that will not happen; we don't handle leap seconds at the end of months other than December or June, and that on the other hand is bound to happen eventually...

astronomy/date_body.hpp
eggrobin: We generally just call it "MSVC".

astronomy/date_body.hpp
eggrobin: Done.

astronomy/date_body.hpp
eggrobin: Done.

astronomy/eop_c04.awk
eggrobin: Done.

astronomy/experimental_eop_c02.awk
eggrobin: Done.

astronomy/date_body.hpp
eggrobin: Moving the EOP entries to their own file would be very difficult, since their methods depend on internals of the UTC interpretation (`TAIMinusStretchyUTC` and the like). On the other hand, it is reasonably easy to split Date, Time, DateTime, and the parsing of their literals from the interpretation according to various time scales.

astronomy/astronomy.vcxproj.filters
```diff
@@ -93,18 +93,24 @@
...
     </ClInclude>
-    <ClInclude Include="eop_c04.generated.h">
+    <ClInclude Include="time_scales.hpp">
       <Filter>Header Files</Filter>
     </ClInclude>
-    <ClInclude Include="experimental_eop_c02.generated.h">
+    <ClInclude Include="time_scales_body.hpp">
+      <Filter>Source Files</Filter>
+    </ClInclude>
+    <ClInclude Include="calendar.hpp">
```
pleroy: As discussed, let's call this file `date_time.hpp`.

PR #1103, opened 2016-08-11
Serialize game epoch

eggrobin:


ksp_plugin/plugin.cpp
```diff
@@ -791,6 +798,7 @@ Plugin::Plugin(GUIDToOwnedVessel vessels,
                Ephemeris<Barycentric>::AdaptiveStepParameters const&
                    prediction_parameters,
                Angle planetarium_rotation,
+               Instant game_epoch,
```
pleroy: Why are these things not `const&`?

eggrobin: I don't know...

PR #1117, opened 2016-08-28
Clang inf const expr

pdn4kd:
As we discovered in building Principia tests on a Debian bird, clang dislikes floating point arithmetic generating infinities, so this change is necessary to get epoch.hpp to compile.

> <@egg> UmbralRaptor: I'm not sure whether it's a clang bug or a GCC & MSVC bug (and in that case, bad design in the standard), and, if the former, whether it's worth filing a bug report.


eggrobin:
This property of clang seems to have been introduced in 2012, see https://android.googlesource.com/platform/external/clang/+/7b48a29%5E!/ and https://github.com/llvm-mirror/clang/blob/72a2930a202d56bcec209eef295f5922da1de899/lib/AST/ExprConstant.cpp#L1921.

PR #1118, opened 2016-08-29
Euler and Cardan

### LINKED ISSUE #810: Constructors for Euler angles
Rotations need a constructor from Euler angles.
The convention should be an enum offering all possibilities (ZXZ, XYX, YZY, ZYZ, XZX, YXY).
While we're at it, a constructor from Cardan-Tait-Bryan (heading-elevation-bank, yaw-pitch-roll) angles should be provided (again with all conventions XYZ YZX ZXY XZY ZYX YXZ, for which we already have a type).
Beware the Jabberwock! game developers, being confused about geometry, [call Cardan-Tait-Bryan angles Euler angles](http://docs.unity3d.com/ScriptReference/Transform-eulerAngles.html).

eggrobin:
This fixes #810.

### DIFF
--- a/geometry/r3_element_body.hpp
+++ b/geometry/r3_element_body.hpp
@@ -313,17 +313,9 @@ quantities::Product<LScalar, RScalar> Dot(
 }
 
 inline R3Element<double> BasisVector(int const i) {
-  switch (i) {
-    case 0:
-      return {1, 0, 0};
-    case 1:
-      return {0, 1, 0};
-    case 2:
-      return {0, 0, 1};
-    default:
-      LOG(FATAL) << FUNCTION_SIGNATURE << ": " << NAMED(i);
-      base::noreturn();
-  }
+  DCHECK_GE(i, 0) << i;
+  DCHECK_LT(i, 3) << i;
+  return {i == 0, i == 1, i == 2};
 }
 
 }  // namespace geometry
--- a/geometry/r3_element_body.hpp
+++ b/geometry/r3_element_body.hpp
@@ -315,7 +315,9 @@ quantities::Product<LScalar, RScalar> Dot(
 inline R3Element<double> BasisVector(int const i) {
   DCHECK_GE(i, 0) << i;
   DCHECK_LT(i, 3) << i;
-  return {i == 0, i == 1, i == 2};
+  return {static_cast<double>(i == 0),
+          static_cast<double>(i == 1),
+          static_cast<double>(i == 2)};
 }
 
 }  // namespace geometry
--- a/geometry/r3_element_body.hpp
+++ b/geometry/r3_element_body.hpp
@@ -312,5 +312,19 @@ quantities::Product<LScalar, RScalar> Dot(
   return left.x * right.x + left.y * right.y + left.z * right.z;
 }
 
+inline R3Element<double> BasisVector(int const i) {
+  switch (i) {
+    case 0:
+      return {1, 0, 0};
+    case 1:
+      return {0, 1, 0};
+    case 2:
+      return {0, 0, 1};
+    default:
+      LOG(FATAL) << FUNCTION_SIGNATURE << ": " << NAMED(i);
+      base::noreturn();
+  }
+}
+
 }  // namespace geometry
 }  // namespace principia
--- a/geometry/rotation.hpp
+++ b/geometry/rotation.hpp
@@ -148,12 +148,11 @@ class Rotation : public LinearMap<FromFrame, ToFrame> {
   // Constructors from Cardan angles.
   // Example: if |Aircraft| is the frame of an aircraft (x forward, y right,
   // z down), and |Ground| is a local-vertical, local-horizontal frame (x North,
-  // y East, z Down), given the |heading|, |elevation|, and |bank| of the
-  // aircraft,
-  //   Rotation<Ground, Aircraft>(heading, elevation, bank, CardanAngles::ZYX,
+  // y East, z Down), given the |heading|, |pitch|, and |roll| of the aircraft,
+  //   Rotation<Ground, Aircraft>(heading, pitch, roll, CardanAngles::ZYX,
   //                              DefinesFrame<Aircraft>{})
   // and
-  //   Rotation<Aircraft, Ground>(heading, elevation, bank, CardanAngles::ZYX,
+  //   Rotation<Aircraft, Ground>(heading, pitch, roll, CardanAngles::ZYX,
   //                              DefinesFrame<Aircraft>{})
   // are the transformations between |Aircraft| and |Ground|.
 
--- a/geometry/rotation.hpp
+++ b/geometry/rotation.hpp
@@ -41,13 +41,7 @@ enum class CardanAngles {
 };
 
 template<typename Frame>
-struct DefinesFrameTag {
-  constexpr DefinesFrameTag() = default;
-  ~DefinesFrameTag() = delete;
-};
-
-template<typename Frame>
-constexpr DefinesFrameTag<Frame> DefinesFrame{};
+struct DefinesFrame {};
 
 // An orientation-preserving orthogonal map between the inner product spaces
 // |FromFrame| and |ToFrame|, as well as the induced maps on the exterior
@@ -97,9 +91,9 @@ class Rotation : public LinearMap<FromFrame, ToFrame> {
   // confusing, and Euler angles are a constant trap.  We therefore require a
   // tag that specifies which frame is being defined by the given angle and
   // axis.
-  // These constructors come in pairs (with either value of DefinesFrameTag),
-  // like the matrix constructors above.  For each pair, one of the constructors
-  // has an additional |typename = void|, see above.
+  // These constructors come in pairs (with either value of DefinesFrame), like
+  // the matrix constructors above.  For each pair, one of the constructors has
+  // an additional |typename = void|, see above.
 
   template<typename Scalar,
            typename F = FromFrame,
@@ -107,7 +101,7 @@ class Rotation : public LinearMap<FromFrame, ToFrame> {
            typename = std::enable_if_t<!std::is_same<F, T>::value>>
   Rotation(Angle const& angle,
            Bivector<Scalar, FromFrame> const& axis,
-           DefinesFrameTag<ToFrame> tag);
+           DefinesFrame<ToFrame> tag);
 
   template<typename Scalar,
            typename F = FromFrame,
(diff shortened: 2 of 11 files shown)


geometry/rotation.hpp
```diff
@@ -20,16 +20,52 @@ template<typename FromFrame, typename ToFrame>
 std::ostream& operator<<(std::ostream& out,
                          Rotation<FromFrame, ToFrame> const& rotation);
 
+// |EulerAngles| and |CardanAngles| have values in binary-coded ternary
```
pleroy: CardanoAngles?

geometry/rotation.hpp
```diff
@@ -20,16 +20,52 @@ template<typename FromFrame, typename ToFrame>
...
+// representing the sequence of rotation axes.
+
+enum class EulerAngles {
+  // |ZXZ| is The most common convention, e.g. orbital elements (Ω, i, ω),
+  // rotational elements (90˚ + α₀, 90˚ - δ₀, W).
+  ZXZ = 0b10'00'10,
```
pleroy: Can we write this` 0b'10'00'10`?

geometry/rotation.hpp
```diff
@@ -20,16 +20,52 @@ template<typename FromFrame, typename ToFrame>
 std::ostream& operator<<(std::ostream& out,
                          Rotation<FromFrame, ToFrame> const& rotation);
 
+// |EulerAngles| and |CardanAngles| have values in binary-coded ternary
+// representing the sequence of rotation axes.
```
pleroy: Explain that `X=0b00`, etc.

geometry/rotation.hpp
```diff
@@ -53,6 +89,92 @@ class Rotation : public LinearMap<FromFrame, ToFrame> {
...
+  // axis.
+  // These constructors come in pairs (with either value of DefinesFrame), like
+  // the matrix constructors above.  For each pair, one of the constructors has
+  // an additional |typename = void|, see above.
+
+  template<typename Scalar,
+           typename F = FromFrame,
+           typename T = ToFrame,
+           typename = std::enable_if_t<!std::is_same<F, T>::value>>
+  Rotation(Angle const& angle,
```
pleroy: I realize that I suggested the tag type, but I think that it would be a bit nicer/more readable to use the "to frame" template parameter to document that you are defining the frame.  I'd like to be able to write:
```
geometry::Rotation<ICRFJ2000Equator, DefineFrame<ICRFJ2000Ecliptic>>(
  23 * Degree + 26 * ArcMinute + 21.448 * ArcSecond,
  geometry::Bivector<double, ICRFJ2000Equator>({-1, 0, 0}));
```
So basically, enable if the "to frame" parameter is a `DefineFrame`.

geometry/rotation_body.hpp
```diff
@@ -59,26 +61,30 @@ FORCE_INLINE Quaternion ToQuaternion(R3x3Matrix const& matrix) {
...
 }
 
+// Returns a rotation of |angle| around |axis|.  |axis| must be normalized.
+Quaternion AngleAxis(Angle const& angle, R3Element<double> const& axis) {
+  quantities::Angle const half_angle = 0.5 * angle;
+  return Quaternion(Cos(half_angle), Sin(half_angle) * axis);
+}
+
+// Returns a rotation of |angle| around the |axis_index|th axis.
+Quaternion AngleAxis(Angle const& angle, int const axis_index) {
```
pleroy: Do we need this special case?  I'd prefer to have something like `BaseVector(int)` in `R3Element`.  That would be more general.

geometry/rotation_body.hpp
```diff
@@ -110,6 +116,64 @@ Rotation<FromFrame, ToFrame>::Rotation(
...
+
+template<typename FromFrame, typename ToFrame>
+template<typename F, typename T, typename, typename>
+Rotation<FromFrame, ToFrame>::Rotation(
+    Angle const& α,
+    Angle const& β,
+    Angle const& γ,
+    EulerAngles const axes,
+    DefinesFrame<FromFrame> tag)
+    : Rotation(AngleAxis(α, (static_cast<int>(axes) >> (2 * 2)) & 0b11) *
```
pleroy: `2` and `0b11` should be constants.

geometry/rotation.hpp
eggrobin: I'm not sure what MSVC thinks, but not according to clang or GCC; `0b0'10'00'10` would be an option.
Amusingly `0'10` appears to be a valid literal equal to `8` in both clang and GCC, but`0x` and `0b` don't seem to work similarly.

geometry/rotation_test.cpp
```diff
@@ -279,5 +284,240 @@ TEST_F(RotationTest, Basis) {
...
+    // Positive angle of attack results in positive z velocity.
+    EXPECT_THAT(v_aircraft.coordinates().z, Gt(0 * Metre / Second));
+
+    EXPECT_THAT(sideslip, Lt(0 * Degree));
+  }
+
+  {
+    // The angle between forward and north.  This is almost certainly a terrible
+    // formula for that, cf. "A Case Study of Bits Lost in Space", in Kahan's
+    // How Futile are Mindless Assessments of Roundoff in Floating-Point
```
pleroy: Quotes around the title.

geometry/rotation.hpp
eggrobin: That would require an implicit conversion (the result has to be a `Rotation<ICRFJ2000Equator, ICRFJ2000Ecliptic>` at the end of the day, which tends to break things, and would be very confusing: you'd expect a constructor to give you the type you asked for).
Moreover, that syntax doesn't sound as good for the inverse (`Rotation<DefinesFrame<A>, B>`?).

geometry/rotation.hpp
eggrobin: If I declare constants, I have to use shifts rather than just chaining the numbers; we did that for permutations, with things like `(Y << X * 2)`, but there it made some sense (x maps to y); here it's not a mapping from XYZ to whichever sequence, so that would be confusing.
If you prefer that to the integer literals, I could declare a function that chains binary-coded ternary digits (e.g. something like `AxisConvention(X, Z, X)`.

geometry/rotation_body.hpp
eggrobin: Even then I'd want to factor `AngleAxis(α, BasisVector(static_cast<int>(axes) >> (2 * 2)) & 0b11))`

geometry/rotation_body.hpp
eggrobin: I'm not sure `binary_coded_ternary_digit_width` and `binary_coded_ternary_least_significant_digit_bitmask` would add to clarity; this reminds me of https://lkml.org/lkml/2016/8/2/1589.
Would a function (templated on the enum class) for `(static_cast<int>(axes) >> (2 * digit)) & 0b11` be better? Maybe `BinaryCodedTernaryDigit(k, axes)`?

geometry/rotation.hpp
eggrobin: Done (also `cardano` in the test, and some `Cardano`s in the comments).

geometry/rotation_test.cpp
eggrobin: Done.

geometry/rotation_body.hpp
pleroy: On second thought, let's not go to LKML, 'tis a silly place.

`BinaryCodedTernaryDigit(k, axes)` sounds reasonable.  Make it constexpr if possible.

geometry/rotation.hpp
pleroy: A function sounds good.

geometry/rotation_body.hpp
eggrobin: Done.

geometry/rotation_body.hpp
eggrobin: Done.

geometry/rotation.hpp
eggrobin: Done.

geometry/r3_element_body.hpp
```diff
@@ -312,5 +312,19 @@ quantities::Product<LScalar, RScalar> Dot(
   return left.x * right.x + left.y * right.y + left.z * right.z;
 }
 
+inline R3Element<double> BasisVector(int const i) {
```
pleroy: This could be written:
```
return {i == 0, i == 1, i == 2};
```
which is probably more efficient despite the ugly implicit conversions.

geometry/rotation_body.hpp
```diff
@@ -67,11 +67,12 @@ Quaternion AngleAxis(Angle const& angle, R3Element<double> const& axis) {
...
-// Returns a rotation of |angle| around the |axis_index|th axis.
-Quaternion AngleAxis(Angle const& angle, int const axis_index) {
-  R3Element<double> axis{0, 0, 0};
-  axis[axis_index] = 1;
-  return AngleAxis(angle, axis);
+// Returns the digits of the 3ⁿs from the given |BinaryCodedTernary number|.
+// Note that this does not check that |number| is valid binary-coded ternary,
+// nor that the result is between 1 and 3.
+template<typename BinaryCodedTernary>
+int BinaryCodedTernaryDigit(int n, BinaryCodedTernary number) {
```
pleroy: const the parameters.

geometry/rotation_body.hpp
```diff
@@ -67,11 +67,12 @@ Quaternion AngleAxis(Angle const& angle, R3Element<double> const& axis) {
...
 }
 
-// Returns a rotation of |angle| around the |axis_index|th axis.
-Quaternion AngleAxis(Angle const& angle, int const axis_index) {
-  R3Element<double> axis{0, 0, 0};
-  axis[axis_index] = 1;
-  return AngleAxis(angle, axis);
+// Returns the digits of the 3ⁿs from the given |BinaryCodedTernary number|.
+// Note that this does not check that |number| is valid binary-coded ternary,
+// nor that the result is between 1 and 3.
```
pleroy: 0 and 2

geometry/rotation_test.cpp
```diff
@@ -320,37 +320,37 @@ TEST_F(RotationTest, Enums) {
...
+  auto const check_cardano_angles = [&α, &β, &γ, &xyz_cardano, this](
+      CardanoAngles axes,
       Permutation<World, World>::CoordinatePermutation permutation) {
     Permutation<World, World> const σ(permutation);
     Permutation<World1, World1> const τ =
         (Permutation<World, World1>::Identity() * σ *
          Permutation<World1, World>::Identity()).Inverse();
-    Rotation<World, World1> const cardan(σ.Determinant() * α,
+    Rotation<World, World1> const cardano(σ.Determinant() * α,
                                          σ.Determinant() * β,
```
pleroy: Bad indent.

geometry/r3_element_body.hpp
eggrobin: Done.

geometry/rotation_body.hpp
eggrobin: Done.

geometry/rotation_body.hpp
eggrobin: Done.

geometry/rotation_test.cpp
eggrobin: Done.

PR #1128, opened 2016-09-08
Change MassiveBody to have a name

pleroy:


eggrobin:
Does this affect the fingerprinting used in stock solar system detection?

PR #1140, opened 2016-10-18
An apocalypse detector...

eggrobin:
... but somehow I cannot trigger it. Fixes some issues in the interface though.
Changes to the code generator: for the changes to the output see https://github.com/mockingbirdnest/Principia/commit/c81f490ddf6af0d8998ed2579bd0f85fdb6716d4.


ksp_plugin/interface.cpp
```diff
@@ -406,12 +406,17 @@ void principia__GetVersion(
...
-  journal::Method<journal::HasEncounteredApocalypse> m({{plugin}, {details}});
+  journal::Method<journal::HasEncounteredApocalypse> m({plugin}, {details});
   // Ownership will be transfered to the marshmallow.
-  std::string* const allocated_details = new std::string;
+  std::string details_string;
   bool const has_encountered_apocalypse =
-      CHECK_NOTNULL(plugin)->HasEncounteredApocalypse(allocated_details);
-  *CHECK_NOTNULL(details) = allocated_details->c_str();
+      CHECK_NOTNULL(plugin)->HasEncounteredApocalypse(&details_string);
+  UniqueBytes allocated_details(details_string.size() + 1);
```
pleroy: This seems very complicated.  I don't think you need allocated_details at all: just copy details_string into *details.

pleroy: Agreed.

PR #1143, opened 2016-10-29
Report collisions

eggrobin:
A tiny diagnosis field; we may want to remove that later, but for now it's semi-useful and a good reference snippet.


ksp_plugin_adapter/ksp_plugin_adapter.cs
```diff
@@ -1036,6 +1036,25 @@ where is_in_physics_bubble(vessel)
       plugin_state = "managing physics bubble";
     }
     UnityEngine.GUILayout.TextArea(text : "Plugin is " + plugin_state);
+    if (FlightGlobals.ActiveVessel != null) {
```
pleroy: Maybe a comment saying that this if statement should be removed once we properly handle collisions.

Issue #1142, opened 2016-10-26
Local Horizontal / Polar frame field


eggrobin:

pleroy:
Needs to be added to the UI.

pleroy:
Fixed in #1263.

PR #1146, opened 2016-11-02
Create a frame field for the surface of a celestial

pleroy:
This is the first part towards addressing #1142.  Some more work is needed in the plugin to making possible to select that frame.


ksp_plugin/plugin.cpp
```diff
@@ -666,29 +666,49 @@ Velocity<World> Plugin::BubbleVelocityCorrection(
                                                  reference_body));
 }
 
-FrameField<World> Plugin::Navball(
+std::unique_ptr<FrameField<World, Navball>> Plugin::NavballFrameField(
```
pleroy: @eggrobin: Please review this part carefully [because...](http://i2.kym-cdn.com/photos/images/original/000/234/739/fa5.jpg)

ksp_plugin/frames.hpp
```diff
@@ -42,6 +42,11 @@ using CelestialSphere = Frame<serialization::Frame::PluginTag,
 using BodyWorld = Frame<serialization::Frame::PluginTag,
                         serialization::Frame::BODY_WORLD, false>;
 
+// The frame used for the navball.  Its definition depends on the choice of a
+// subclass of FrameField.
+using Navball = Frame<serialization::Frame::PluginTag,
+                      serialization::Frame::NAVBALL, false>;
```
eggrobin: Document orientation.

ksp_plugin/plugin.cpp
```diff
@@ -666,29 +666,49 @@ Velocity<World> Plugin::BubbleVelocityCorrection(
...
-    auto const orthogonal_map = to_world *
-        plotting_frame_->FromThisFrameAtTime(current_time_).orthogonal_map() *
-        Permutation<World, Navigation>(
-            Permutation<World, Navigation>::XZY).Forget() *
-        Rotation<World, World>(π / 2 * Radian,
-                               Bivector<double, World>({0, 1, 0})).Forget();
-    CHECK(orthogonal_map.Determinant().Positive());
-    return orthogonal_map.rotation();
+
+  class NavballFrameField : public FrameField<World, Navball> {
```
eggrobin: I don't like this class in its current state; what's actually being returned is `CoordinateFrame<Navigation, RightHandedNavball>`, whose coordinate (lack of) rotation is composed left and right with (position-independent) orthogonal maps. Here the dependence on `CoordinateFrame` not made explicit, in fact `RightHandedNavball` is never formalized. This matters, because when `FrameField`s other than `CoordinateField` are used (that's the whole point of `BodySurfaceFrameField`), we can just put them through the same thing.
Thus, this class should take a `CoordinateFrame<Navigation, RightHandedNavball>` at construction, and use that. `RightHandedNavball` need not be exposed too high; a forward-declaration within `NavballFrameField` will do the job well enough.

physics/body_surface_frame_field_body.hpp
```diff
@@ -0,0 +1,55 @@
...
+    : body_axis_(body->polar_axis()),
+      body_position_(
+          ephemeris.trajectory(body)->EvaluatePosition(t, /*hint=*/nullptr)) {}
+
+template<typename Frame, typename ThisFrame>
+Rotation<ThisFrame, Frame>
+BodySurfaceFrameField<Frame, ThisFrame>::FromThisFrame(
+    Position<Frame> const& q) const {
+  // A vector going from the centre of the body to the point in the field.
+  Displacement<Frame> const displacement = q - body_position_;
```
eggrobin: A comment that reminds the reader of the definition of position subtraction isn't all that useful; call the variable `from_body_centre` or something like that, so that we have some idea what it is below.

physics/body_surface_frame_field_body.hpp
```diff
@@ -0,0 +1,55 @@
...
+
+  Vector<double, Frame> const normalized_displacement = Normalize(displacement);
```
eggrobin: s/normalized_displacement/up/ (or zenith if you prefer)

physics/body_surface_frame_field_body.hpp
```diff
@@ -0,0 +1,55 @@
...
+  double const axis_projection =
+      InnerProduct(normalized_displacement, body_axis_);
+  double const axis_projection² = axis_projection * axis_projection;
+
+  // The unit vector x directed along the polar axis is
+  // |λ * body_axis + μ * normalized_displacement|.  Note that λ is positive.
+  double const λ = 1 / Sqrt(1 - axis_projection²);
+  auto const μ = -λ * axis_projection;
+
+  Vector<double, Frame> const x = λ * body_axis_ + μ * normalized_displacement;
```
eggrobin: `north` (or more accurately `to_high_latitudes`, for correctness with retrograde small bodies)

physics/body_surface_frame_field_body.hpp
```diff
@@ -0,0 +1,55 @@
...
+  Vector<double, Frame> const& z = normalized_displacement;
+  Bivector<double, Frame> const y = Wedge(z, x);
```
eggrobin: `east`

physics/body_surface_frame_field_body.hpp
```diff
@@ -0,0 +1,55 @@
...
+
+  return Rotation<ThisFrame, Frame>(x, y, z);
```
eggrobin: The comments above yield `(up, east, north)`; this is more informative, since it tells the reader that x is up etc.
Thinking about it, this is probably not what we want; where we have `CoordinateFrame`s, the reference plane of the navball is the xy plane (the body equator in https://github.com/mockingbirdnest/Principia/blob/master/physics/body_surface_dynamic_frame.hpp#L33-L36, the celestial equator in https://github.com/mockingbirdnest/Principia/blob/master/astronomy/frames.hpp#L22-L30).
Moreover, the reference direction on that plane is x.
Since here the Navball reference plane is the horizontal, this should be `(north, west, up)`. It is probably preferable to go with the more aerospace-friendly convention [`(north, east, down)`](https://en.wikipedia.org/wiki/Axes_conventions#Ground_reference_frames:_ENU_and_NED), which also has xy as the reference plane and x as the reference direction on the plane.

This may imply that the comment in the `Plugin` about our navball orientation and KSP's be incorrect.

physics/body_surface_frame_field_body.hpp
pleroy: It's `west`, I think.

physics/body_surface_frame_field_body.hpp
pleroy: The comments above actually yield `(north, west, up)` I think.  Oh, and this is `physics`, let's not put aerospace-friendliness there.

physics/body_surface_frame_field_body.hpp
eggrobin: right. Hence the importance of giving them names, I had managed to confuse myself over the course of that handful of lines.

physics/body_surface_frame_field_body.hpp
eggrobin: Right, it's NWU. It's not just aerospace-friendlyness though; one of the conventions is a standard, so let's go with it rather than make our own.

eggrobin: a North-West-Up system is generally a Really Bad Idea, because spherical coordinates on there get headings backwards (90° becomes W instead of E). Hence the use of East-North-Up or North-East-Down. It is quite clear here that NED is the one we want, for consistency with our use of x as the reference direction on the reference plane.

eggrobin: I'll make a CL to change https://github.com/mockingbirdnest/Principia/blob/master/ksp_plugin_adapter/assets/navball_compass.png accordingly (this is just a matter of rotating the image by π, since on the physical ball we do want the numbers facing up). Note that this will make the direction of the numbering consistent with, e.g., https://github.com/mockingbirdnest/Principia/blob/master/ksp_plugin_adapter/assets/navball_inertial.png.

physics/body_surface_frame_field_body.hpp
pleroy: Done.

physics/body_surface_frame_field_body.hpp
pleroy: Let's use `north`, it's clear enough.

physics/body_surface_frame_field_body.hpp
pleroy: Done.

physics/body_surface_frame_field_body.hpp
pleroy: Done.

ksp_plugin/frames.hpp
pleroy: Done.

ksp_plugin/plugin.cpp
pleroy: I tried to do something, but see the `TODO`.

PR #1151, opened 2016-11-13
Change the integrator API to separate the problem from the integrator instance

pleroy:


integrators/symplectic_runge_kutta_nyström_integrator.hpp
```diff
@@ -81,17 +81,31 @@ class SymplecticRungeKuttaNyströmIntegrator
...
 
  private:
+  struct Instance : public IntegrationInstance {
+    Instance(IntegrationProblem<ODE> problem,
+             AppendState<ODE> append_state,
+             Time step);
+    IntegrationProblem<ODE> const problem;
+    AppendState<ODE> const append_state;
+    Time const step;
+  };
```
eggrobin: Resumability; for the adaptive case, store things for seamless resumability at the last natural point (and we'll want to make reaching exactly t_final an option).

integrators/embedded_explicit_runge_kutta_nyström_integrator_body.hpp
```diff
@@ -239,19 +250,55 @@ Status EmbeddedExplicitRungeKuttaNyströmIntegrator<Position,
...
-    problem.append_state(current_state);
+    append_state(current_state);
     ++step_count;
     if (step_count == adaptive_step_size.max_steps && !at_end) {
       return Status(termination_condition::ReachedMaximalStepCount,
                     "Reached maximum step count " +
                         std::to_string(adaptive_step_size.max_steps) +
                         " at time " + DebugString(t.value) +
-                        "; problem t_final is " + DebugString(problem.t_final) +
+                        "; problem t_final is " + DebugString(t_final) +
```
eggrobin: it's not the problem anymore; "requested t_final"?

integrators/embedded_explicit_runge_kutta_nyström_integrator_body.hpp
```diff
@@ -239,19 +250,55 @@ Status EmbeddedExplicitRungeKuttaNyströmIntegrator<Position,
...
                         ".");
     }
   }
+  // The resolution is restartable from the last non-truncated state.
+  if (final_state) {
```
eggrobin: I think by this point `final_state` is always engaged, thus no need for the conditional; if that's not the case, this code might be incorrect, since setting `final_state` can be followed by picking a shorter time step instead if `tolerance_to_error_ratio > 1`.

PR #1153, opened 2016-11-22
Bundle

eggrobin:


base/bundle.hpp
```diff
@@ -0,0 +1,79 @@
...
+  Bundle(int threads);
+
+  // Returns the first non-OK status encountered.  All threads are joined;
+  // no calls to |Add| or |Join| may follow this call.
+  Status Join();
+
+  // The |Task| returned by |Add| may be used in calls to |Abort| or |JoinTask|;
+  // |this| shares ownership until |task| has returned or is dequeued by
+  // |Abort|.
+  std::shared_ptr<Task> Add(std::function<Status()> task);
```
pleroy: Calling the parameter `task` makes this unclear.  Here and in the comments below I would like to distinguish the function/callback from the `Task`.  (You use `body` elsewhere, that works better.)

base/bundle_test.cpp
```diff
@@ -0,0 +1,93 @@
...
+  for (int i = 0; i < short_dimension; ++i) {
+    EXPECT_THAT(product[i],
+                Eq(long_dimension * (long_dimension - 1) / 2 + i));
+  }
+}
+
+TEST_F(BundleTest, Abort) {
+  int i = 0;
+  auto const wait = [this]() {
+    while (!bundle_.abort_requested()) {}
```
pleroy: Alert!  Ugly busy waiting detected!  Put a sleep in the loop otherwise the test might flake.

base/bundle.cpp
```diff
@@ -0,0 +1,113 @@
...
+namespace base {
+
+Bundle::Bundle(int threads) {
+  {
+    std::unique_lock<std::mutex> l(lock_);
+    joining_ = false;
+  }
+  for (int i = 0; i < threads; ++i) {
+    threads_.emplace_back([this]() {
+      std::shared_ptr<Task> current_task;
```
pleroy: Please extract this lambda to a member function.  It is hard to figure out what the constructor is doing, especially with the lambdas nested within the lambda.

base/bundle.cpp
```diff
@@ -0,0 +1,113 @@
...
+namespace principia {
+namespace base {
+
+Bundle::Bundle(int threads) {
+  {
+    std::unique_lock<std::mutex> l(lock_);
+    joining_ = false;
+  }
+  for (int i = 0; i < threads; ++i) {
+    threads_.emplace_back([this]() {
```
pleroy: The threads should only be created on demand.  If the bundle is not used it should cost nothing.

base/bundle.cpp
```diff
@@ -0,0 +1,113 @@
...
+      std::shared_ptr<Task> current_task;
+      for (;;) {
+        {
+          std::unique_lock<std::mutex> l(lock_);
+          if (joining_ && tasks_.empty()) {
+            return;
+          }
+          tasks_not_empty_or_joining_.wait(
+              l,
+              [this]() { return !tasks_.empty() || joining_; });
```
pleroy: Make this a function.

base/bundle.cpp
```diff
@@ -0,0 +1,113 @@
...
+    threads_.emplace_back([this]() {
+      std::shared_ptr<Task> current_task;
+      for (;;) {
+        {
+          std::unique_lock<std::mutex> l(lock_);
+          if (joining_ && tasks_.empty()) {
+            return;
+          }
+          tasks_not_empty_or_joining_.wait(
+              l,
```
pleroy: It took me a while to figure out that this is not `1`.  Maybe name the `unique_lock` `ul`.  Stupid language.

base/bundle.cpp
```diff
@@ -0,0 +1,113 @@
...
+  for (int i = 0; i < threads; ++i) {
+    threads_.emplace_back([this]() {
+      std::shared_ptr<Task> current_task;
+      for (;;) {
+        {
+          std::unique_lock<std::mutex> l(lock_);
+          if (joining_ && tasks_.empty()) {
+            return;
+          }
+          tasks_not_empty_or_joining_.wait(
```
pleroy: I think that you are missing the protection against the spurious wake.

base/bundle.cpp
```diff
@@ -0,0 +1,113 @@
...
+        FindOrDie(abort_requests_, std::this_thread::get_id()) = false;
+      }
+    });
+    abort_requests_.emplace(threads_.back().get_id(), false);
+  }
+}
+
+std::shared_ptr<Bundle::Task> Bundle::Add(std::function<Status()> task) {
+  // Private constructor, cannot |make_shared|.
+  auto result = std::shared_ptr<Task>(new Task(task));
```
pleroy: `std::move(task)`

base/bundle.cpp
```diff
@@ -0,0 +1,113 @@
...
+  }
+  tasks_not_empty_or_joining_.notify_all();
+  for (auto& thread : threads_) {
+    thread.join();
+  }
+  std::unique_lock<std::mutex> l(lock_);
+  return status_;
+}
+
+void Bundle::Abort(not_null<Task*> task) {
```
pleroy: `const task`

base/bundle.cpp
```diff
@@ -0,0 +1,113 @@
...
+    tasks_.erase(*task->in_queue_);
+    task->in_queue_ = std::experimental::nullopt;
+  } else if (task->abort_requested_ != nullptr) {
+    *task->abort_requested_ = true;
+  }
+  // TODO(egg): Should we do something, e.g., set task->status_ to aborted if it
+  // terminated normally then was aborted?
+}
+
+Status Bundle::JoinTask(not_null<Task*> task) {
```
pleroy: `const task`

base/bundle.hpp
```diff
@@ -0,0 +1,79 @@
...
+    Task(std::function<Status()> body);
+
+    std::function<Status()> body_;
+
+    // |status_|, |in_queue_|, and |abort_requested_| are guarded by the |lock_|
+    // of the |Bundle|.
+
+    // Has a value if and only if the task has completed.
+    std::experimental::optional<Status> status_;
+    // Has a value if and only if the task has yet to start.
```
pleroy: This needs a better comment.  The position of the task in the queue?  Also explain why you need it.

base/bundle.hpp
```diff
@@ -0,0 +1,79 @@
...
+    std::experimental::optional<std::list<std::shared_ptr<Task>>::iterator>
+        in_queue_;
+    // Is non-null if and only if the task is running.
+    std::atomic<bool>* abort_requested_ = nullptr;
```
pleroy: Why does it have to be a pointer and not an atomic bool directly?  Who owns it?  The map I presume.

base/bundle.hpp
```diff
@@ -0,0 +1,79 @@
...
+    std::condition_variable done_;
+
+    friend class Bundle;
+  };
+
+ private:
+  std::mutex lock_;
+  std::condition_variable tasks_not_empty_or_joining_;
+
+  bool joining_ GUARDED_BY(lock_);
```
pleroy: This means "joining with all the tasks" presumably.  Comments would help for all these member variables.

base/bundle.hpp
```diff
@@ -0,0 +1,79 @@
...
+  std::list<std::shared_ptr<Task>> tasks_ GUARDED_BY(lock_);
+  Status status_ GUARDED_BY(lock_);
+
+  std::vector<std::thread> threads_;
+  std::map<std::thread::id, std::atomic<bool>> abort_requests_;
```
pleroy: I am not sure why we need this map at all.  Why can't we live with just the boolean in the `Task`?  Also, depending on the `thread::id` is ugly.

It would be good to explain that these variables don't need locking because they are fixed at construction (but see my comment about creating threads lazily).

base/bundle.cpp
```diff
@@ -0,0 +1,113 @@
...
+  std::unique_lock<std::mutex> l(lock_);
+  // TODO(egg): this should use has_value rather than static_cast once we have
+  // a better optional.
+  task->done_.wait(l, [task]() { return static_cast<bool>(task->status_); });
```
pleroy: Spurious wake.

base/bundle.cpp
```diff
@@ -0,0 +1,113 @@
...
+          }
+          tasks_not_empty_or_joining_.wait(
+              l,
+              [this]() { return !tasks_.empty() || joining_; });
+          if (tasks_.empty()) {
+            return;
+          }
+          current_task = std::move(tasks_.front());
+          tasks_.pop_front();
+          current_task->in_queue_ = std::experimental::nullopt;
```
pleroy: It seems to me that what you are doing with `in_queue_` and `abort_requested_` would be better implemented with a state machine: *queued*->*started*->*completed*->*done* with some additional transitions to *aborted* and *joining*.  That would bring clarity to the code.

base/bundle.cpp
```diff
@@ -0,0 +1,155 @@
...
+  {
+    std::unique_lock<std::mutex> lock(lock_);
+    CHECK(!terminate_on_empty_);
+    terminate_on_empty_ = true;
+  }
+  tasks_not_empty_or_terminate_.notify_all();
+  for (auto& thread : workers_) {
+    thread.join();
+  }
+  std::unique_lock<std::mutex> lock(lock_);
```
pleroy: Not needed since no other thread is running at this point (hopefully you cannot call Add and succeed once you've started joining).

base/bundle.cpp
```diff
@@ -0,0 +1,155 @@
...
+  return status_;
+}
+
+Bundle::TaskHandle Bundle::Add(std::function<Status()> task_body) {
+  if (workers_.size() < max_workers_) {
+    workers_.emplace_back(&Bundle::Toil, this);
+  }
+  // Private constructor, cannot |make_shared|.
+  TaskHandle result(std::make_shared<Task>(this, std::move(task_body)));
```
pleroy: Call this variable `handle`, it will be more transparent.

base/bundle.hpp
```diff
@@ -0,0 +1,140 @@
...
+
+    // Requires |execution_state() == Inactive|.  Sets the |Active| bit, and
+    // calls |body_()|.  When |body_()| returns, sets the |Terminated| bit, and
+    // notifies |terminated_|.
+    void Execute();
+
+    // Requires |execution_state() & Terminated|.
+    // Returns the status returned by |body()|, or a status with
+    // |Error::ABORTED| if |Abort| was called while |Inactive|.
+    Status status() const;
```
pleroy: It would be more logical to return the status in `wait_for_termination` as calling it at any other point is incorrect anyway.

base/bundle.cpp
```diff
@@ -0,0 +1,155 @@
...
+  task->Abort();
+}
+
+Status Bundle::JoinTask(TaskHandle const handle) {
+  not_null<Task*> task = handle.task_.get();
+  task->wait_for_termination();
+  return task->status();
+}
+
+void Bundle::Toil() {
```
pleroy: `Toil` is a cute name but `WorkerLoop` would be clearer...

base/bundle.cpp
```diff
@@ -0,0 +1,155 @@
...
+  std::shared_ptr<Task> current_task;
+  for (;;) {
+    {
+      std::unique_lock<std::mutex> lock(lock_);
+      if (terminate_on_empty_ && tasks_.empty()) {
```
pleroy: I am going to claim that you don't need this statement, or rather that you can merge it with the one at line 78.  You can always wait on the condition variable, and if you must exit the loop it will immediately be true.

base/bundle.cpp
```diff
@@ -0,0 +1,155 @@
...
+          lock,
+          [this] { return !tasks_.empty() || terminate_on_empty_; });
+      if (tasks_.empty()) {
+        return;
+      }
+      current_task = std::move(tasks_.front());
+      tasks_.pop_front();
+    }
+    current_task->Execute();
+    Status status = current_task->status();
```
pleroy: It would seem more logical for `Execute` to return its status.  Mostly I would avoid the function `status` altogether.

base/bundle.cpp
```diff
@@ -0,0 +1,155 @@
...
+    {
+      std::unique_lock<std::mutex> lock(lock_);
+      if (!status.ok() && status_.ok()) {
+        status_ = status;
+      }
+    }
+    current_task.reset();
```
pleroy: Is this needed?  You could just declare `current_task` in the loop I think and the right destruction would happen.

base/bundle.cpp
```diff
@@ -0,0 +1,155 @@
...
+  }
+}
+
+Bundle::Task::Task(not_null<Bundle*> bundle, std::function<Status()> body)
+    : bundle_(bundle),
+      body_(std::move(body)) {}
+
+std::uint8_t Bundle::Task::execution_state() const {
+  std::unique_lock<std::mutex> bundle_lock(bundle_->lock_);
```
pleroy: Is there a non-exclusive lock in this language?  Just asking, maybe not (yet).

base/bundle.cpp
```diff
@@ -0,0 +1,155 @@
...
+  terminated_.notify_all();
+}
+
+Status Bundle::Task::status() const {
+  std::unique_lock<std::mutex> bundle_lock(bundle_->lock_);
+  CHECK_NE(execution_state_ & Terminated, 0) << execution_state_;
+  return status_;
+}
+
+void Bundle::Task::wait_for_termination() {
```
pleroy: I'd call this `WaitForTermination`.  It does real work.

base/bundle.cpp
```diff
@@ -0,0 +1,155 @@
...
+    std::unique_lock<std::mutex> bundle_lock(bundle_->lock_);
+    status_ = std::move(status);
+    execution_state_ |= Terminated;
+  }
+  terminated_.notify_all();
+}
+
+Status Bundle::Task::status() const {
+  std::unique_lock<std::mutex> bundle_lock(bundle_->lock_);
+  CHECK_NE(execution_state_ & Terminated, 0) << execution_state_;
```
pleroy: What if the task is aborted.

base/bundle.hpp
```diff
@@ -0,0 +1,140 @@
...
+
+namespace principia {
+namespace base {
+
+// Functions that can be cooperatively aborted should check |AbortRequested()|
+// when appropriate.
+extern thread_local std::function<bool()> AbortRequested;
+
+class Bundle {
+  class Task;
```
pleroy: I'm not in love with the names, it exposes the ugly `TaskHandle` and keeps the nice `Task` internally.  Maybe `Task` and `TaskState`?

base/bundle.hpp
```diff
@@ -0,0 +1,140 @@
...
+    Status status() const;
+
+    // Returns when the |Terminated| bit is set.
+    void wait_for_termination();
+
+   private:
+    not_null<Bundle*> const bundle_;
+    std::function<Status()> body_;
+
+    std::uint8_t execution_state_ GUARDED_BY(bundle_->lock_) = Inactive;
```
pleroy: It's very bad practice to have a class operating on a mutex owned by another class.  I would prefer to say that `Task` is not thread-safe and the `Bundle` must do proper locking.

base/bundle.hpp
```diff
@@ -0,0 +1,140 @@
...
+#include "base/macros.hpp"
+#include "base/not_null.hpp"
+#include "base/status.hpp"
+
+namespace principia {
+namespace base {
+
+// Functions that can be cooperatively aborted should check |AbortRequested()|
+// when appropriate.
+extern thread_local std::function<bool()> AbortRequested;
```
pleroy: Could we make this a static member of the bundle, it seems disconnected here?

Also, the comment should mention that this should exclusively be called from the `task_body`.

This doesn't really work when `task_body` itself starts threads.  I think that I preferred it when this was a member function of the `Bundle` because then it was clear at the call site what was being checked.

base/bundle.hpp
```diff
@@ -0,0 +1,92 @@
...
+  // workers will terminate without considering queued tasks.  Set by |Abort|,
+  // accessed by |Aborting|, returned by |Join|.
+  Status status_ GUARDED_BY(abort_lock_);
+  std::experimental::optional<std::chrono::steady_clock::time_point> deadline_
+      GUARDED_BY(abort_lock_);
+
+  // Whether the workers should terminate when no tasks are available.  Set by
+  // |Join|.
+  bool terminate_on_empty_ GUARDED_BY(lock_);
+  std::queue<Task> tasks_ GUARDED_BY(lock_);
```
pleroy: Blank line before.

base/bundle.hpp
```diff
@@ -0,0 +1,92 @@
...
+  std::vector<std::thread> workers_ GUARDED_BY(lock_);
+
+  // A pointer to |AbortRequested| on the thread which created the |Bundle|.
```
pleroy: Move down one line.

base/bundle.cpp
```diff
@@ -0,0 +1,119 @@
...
+
+#include "base/map_util.hpp"
+#include "base/status.hpp"
+
+namespace principia {
+namespace base {
+
+thread_local std::function<bool()> AbortRequested = [] { return false; };
+
+Bundle::Bundle(int threads)
```
pleroy: `const int`

base/bundle.hpp
```diff
@@ -0,0 +1,92 @@
...
+  // If |!status_.ok()|, currently-running tasks should cooperatively abort, and
+  // workers will terminate without considering queued tasks.  Set by |Abort|,
+  // accessed by |Aborting|, returned by |Join|.
+  Status status_ GUARDED_BY(abort_lock_);
+  std::experimental::optional<std::chrono::steady_clock::time_point> deadline_
+      GUARDED_BY(abort_lock_);
+
+  // Whether the workers should terminate when no tasks are available.  Set by
+  // |Join|.
+  bool terminate_on_empty_ GUARDED_BY(lock_);
```
pleroy: Could this be a `Monostable`?

base/bundle.hpp
```diff
@@ -0,0 +1,92 @@
...
+  bool DeadlineExceeded();
+
+  // |abort_lock_| should not be held when locking |lock_|.
+  std::mutex lock_;
+  // Notified once when a task is available for execution.  Notified to all
+  // waiting workers when either |terminate_on_empty_| or |abort_| is set.
+  // Workers wait on this when no |tasks_| are available.
+  std::condition_variable tasks_not_empty_or_terminate_;
+
+  std::shared_mutex abort_lock_;
```
pleroy: Would be better named `status_lock_`.

base/bundle.cpp
```diff
@@ -0,0 +1,119 @@
...
+          });
+      // The call to |BundleShouldAbort| checks for deadline expiry and master
+      // abort.
+      if (tasks_.empty() || BundleShouldAbort()) {
+        return;
+      }
+      current_task = std::move(tasks_.front());
+      tasks_.pop();
+    }
+    Status status = current_task();
```
pleroy: `const`

base/bundle.cpp
```diff
@@ -0,0 +1,119 @@
...
+    if ((*master_abort_)()) {
+      Abort(Status(Error::ABORTED, "abort requested on bundle master"));
+    } else if (DeadlineExceeded()) {
+      Abort(Status(Error::DEADLINE_EXCEEDED, "bundle deadline exceeded"));
+    }
+  }
+  return Aborting();
+}
+
+void Bundle::Abort(Status status) {
```
pleroy: `const`

base/bundle.cpp
```diff
@@ -0,0 +1,119 @@
...
+
+namespace principia {
+namespace base {
+
+thread_local std::function<bool()> AbortRequested = [] { return false; };
+
+Bundle::Bundle(int threads)
+    : terminate_on_empty_(false),
+      max_workers_(threads),
+      master_abort_(&AbortRequested) {
```
pleroy: I am not sure how this works.  This is evaluated on the thread that creates the `Bundle`.  Later, it is called on a worker thread.  A comment would be good to explain how the two get connected.

base/bundle.hpp
eggrobin: Done.

base/bundle.hpp
eggrobin: Done.

base/bundle.cpp
eggrobin: `int const` surely

base/bundle.cpp
eggrobin: Done.

base/bundle.cpp
eggrobin: Done.

base/bundle.cpp
eggrobin: Wrote a wall of text before the definition of |BundleShouldAbort|.

PR #1159, opened 2016-12-04
The old SPRK and the new API

eggrobin:
Note the Finnish E. J. Nyström, but the Norwegian C. Størmer.


integrators/symplectic_partitioned_runge_kutta_integrator.hpp
```diff
@@ -0,0 +1,343 @@
...
+#include "symplectic_runge_kutta_nyström_integrator.hpp"
+
+namespace principia {
+
+using numerics::FixedVector;
+
+namespace integrators {
+
+// A symplectic partitioned Runge-Kutta integrator.  Does not subclass
+// |Integrator| yet; used to generate (less general)
```
pleroy: Would be good to have a bug to subclass `Integrator`, unless you plan to do it really soon.

eggrobin: Frankly, much like Adams-Moulton, I'm not sure we want to do it in the foreseeable future. This would only be useful if we wanted to play with things like the alternating Keplerian/Cartesian evolution or the like---arguably it would be interesting to see what sort of performance that has, but since we care about dense output we would probably be more interested in multistep things.

pleroy: Then you should remove "yet" from the comment...

PR #1163, opened 2016-12-08
OrthogonalizedAgainst

eggrobin:


geometry/grassmann_body.hpp
```diff
@@ -109,16 +109,20 @@ Scalar Multivector<Scalar, Frame, 3>::Norm() const {
 
 template<typename Scalar, typename Frame>
 template<typename S>
-void Multivector<Scalar, Frame, 1>::Orthogonalize(
-    Multivector<S, Frame, 1>& multivector) const {
-  coordinates_.template Orthogonalize<S>(multivector.coordinates_);
+Multivector<Scalar, Frame, 1>
+    Multivector<Scalar, Frame, 1>::OrthogonalizedAgainst(
```
pleroy: The indent is weird.  Is this clang-format?

geometry/grassmann_body.hpp
```diff
@@ -109,16 +109,20 @@ Scalar Multivector<Scalar, Frame, 3>::Norm() const {
...
+      coordinates_.OrthogonalizedAgainst(multivector.coordinates_));
 }
 
 template<typename Scalar, typename Frame>
 template<typename S>
-void Multivector<Scalar, Frame, 2>::Orthogonalize(
-    Multivector<S, Frame, 2>& multivector) const {
-  coordinates_.template Orthogonalize<S>(&multivector->coordinates_);
+Multivector<Scalar, Frame, 2>
+    Multivector<Scalar, Frame, 2>::OrthogonalizedAgainst(
```
pleroy: Same.

geometry/grassmann.hpp
```diff
@@ -37,7 +37,8 @@ class Multivector<Scalar, Frame, 1> {
   Scalar Norm() const;
 
   template<typename S>
-  void Orthogonalize(Multivector<S, Frame, 1>& multivector) const;
+  Multivector OrthogonalizedAgainst(
+      Multivector<S, Frame, 1> const& multivector) const;
```
pleroy: Do we need `<S, Frame, 1>`?

geometry/grassmann.hpp
```diff
@@ -78,7 +79,8 @@ class Multivector<Scalar, Frame, 2> {
   Scalar Norm() const;
 
   template<typename S>
-  void Orthogonalize(Multivector<S, Frame, 2>& multivector) const;
+  Multivector OrthogonalizedAgainst(
+      Multivector<S, Frame, 2> const& multivector) const;
```
pleroy: Same.

geometry/grassmann.hpp
eggrobin: well, we need template parameters, this isn't the same type as this `Multivector`. We could go with `Vector`, but I'm not sure it would make things clearer (unless we also change the return type to `Vector<Scalar, Frame>`, but then that's even more template parameters...).

geometry/grassmann_body.hpp
eggrobin: yes (both)

PR #1186, opened 2016-12-30
Explicit up cast

eggrobin:


integrators/ordinary_differential_equations_body.hpp
```diff
@@ -1,16 +1,33 @@
...
 #include "integrators/symplectic_runge_kutta_nyström_integrator.hpp"
 
 namespace principia {
 namespace integrators {
 namespace internal_ordinary_differential_equations {
 
+template<typename Position_>
+SpecialSecondOrderDifferentialEquation<Position_>::SystemState::SystemState(
+    std::vector<typename SpecialSecondOrderDifferentialEquation<
+        Position_>::Position> const& q,
```
pleroy: Can't we just write `std::vector<Position>` here?  I would expect that name to be visible.

integrators/ordinary_differential_equations_body.hpp
```diff
@@ -1,16 +1,33 @@
...
+    std::vector<typename SpecialSecondOrderDifferentialEquation<
+        Position_>::Velocity> const& v,
```
pleroy: Same.

numerics/double_precision.hpp
```diff
@@ -26,7 +26,7 @@ struct DoublePrecision final {
 
   // This constructor is not explicit to make it easy to construct an object
```
pleroy: Remove this comment.

integrators/ordinary_differential_equations_body.hpp
```diff
@@ -1,16 +1,33 @@
...
 #include "integrators/symmetric_linear_multistep_integrator.hpp"
 #include "integrators/symplectic_runge_kutta_nyström_integrator.hpp"
 
 namespace principia {
 namespace integrators {
 namespace internal_ordinary_differential_equations {
 
+template<typename Position_>
+SpecialSecondOrderDifferentialEquation<Position_>::SystemState::SystemState(
+    std::vector<typename SpecialSecondOrderDifferentialEquation<
```
pleroy: I would expect `std::vector<Position>` to work, as it seems that `Position` would be visible here.  Why is this trick needed?

integrators/ordinary_differential_equations_body.hpp
pleroy: Same for velocity.

PR #1208, opened 2017-01-23
Enable multi-processor compilation.

pleroy:


pleroy:
Turns out `/MP` is incompatible with `/Gm`.

pleroy:
Turns out we like the property sheet.

PR #1214, opened 2017-01-28
Bubble frames, and more switching to the new APIs

eggrobin:


ksp_plugin/plugin.cpp
```diff
@@ -415,6 +417,10 @@ void Plugin::FreeVesselsAndCollectPileUps() {
   }
 }
 
+void Plugin::AddPileUpToBubble(std::list<PileUp>::iterator pile_up) {
```
pleroy: const

serialization/journal.proto
```diff
@@ -593,6 +593,22 @@ message GetBufferedLogging {
   optional Return return = 3;
 }
 
+message GetCelestialDegreesOfFreedom {
```
pleroy: Could we call this CelestialDegreesOfFreedom?  That would be consistent with the other celestial-related methods.

ksp_plugin/interface.cpp
```diff
@@ -240,6 +240,16 @@ WXYZ principia__CelestialSphereRotation(Plugin const* const plugin) {
   return m.Return(ToWXYZ(plugin->CelestialSphereRotation().quaternion()));
 }
 
+QP principia__CelestialWorldDegreesOfFreedom(Plugin const* const plugin,
+                                           int const index) {
```
pleroy: Bad indent.

PR #1228, opened 2017-03-05
Through to the adapter

eggrobin:
Utterly untested of course. But it builds.


ksp_plugin/plugin.cpp
```diff
@@ -570,6 +570,19 @@ void Plugin::AdvanceTime(Instant const& t, Angle const& planetarium_rotation) {
   loaded_vessels_.clear();
 }
 
+DegreesOfFreedom<World> Plugin::GetPartActualDegreesOfFreedom(
+    PartId part_id) const {
```
pleroy: const

PR #1230, opened 2017-03-06
Test the things

eggrobin:
Fix some bug and uncover others; the C# has some traces that I'm going to need as soon as things stop crashing below it, so I'm keeping them; the INFO traces on the C++ should stay, because they're for part creation/transfer/destruction, which should not be overly spammy.

### DIFF
--- a/ksp_plugin/plugin.cpp
+++ b/ksp_plugin/plugin.cpp
@@ -705,7 +705,7 @@ not_null<std::unique_ptr<DiscreteTrajectory<World>>> Plugin::RenderedPrediction(
     Position<World> const& sun_world_position) const {
   CHECK(!initializing_);
   Vessel const& vessel = *find_vessel_by_guid_or_die(vessel_guid);
-  return RenderedTrajectoryFromIterators(vessel.prediction().Fork(),
+  return RenderedTrajectoryFromIterators(vessel.prediction().Begin(),
                                          vessel.prediction().End(),
                                          sun_world_position);
 }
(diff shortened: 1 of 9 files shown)


ksp_plugin/plugin.cpp
```diff
@@ -105,9 +105,7 @@ Plugin::Plugin(Instant const& game_epoch,
       prediction_parameters_(DefaultPredictionParameters()),
       planetarium_rotation_(planetarium_rotation),
       game_epoch_(game_epoch),
-      current_time_(solar_system_epoch) {}
-
-void Plugin::InsertCelestialAbsoluteCartesian(
+      current_time_(solar_system_epoch) {}void Plugin::InsertCelestialAbsoluteCartesian(
```
pleroy: Botched.

ksp_plugin/plugin.cpp
```diff
@@ -141,6 +139,21 @@ void Plugin::InsertCelestialAbsoluteCartesian(
                                                   initial_state);
 }
 
+Plugin::~Plugin() {
+  // This is largely awful; If we simply let the destructor of |Part| deal with
```
pleroy: Remove the awful, bad, very sad adjectives.

serialization/ksp_plugin.proto
```diff
@@ -119,14 +119,16 @@ message Vessel {
...
   repeated Part parts = 14;
   repeated uint32 kept_parts = 15;
   required DiscreteTrajectory psychohistory = 16;
   required bool psychohistory_is_authoritative = 17;
   required DiscreteTrajectory prediction = 18;
   optional FlightPlan flight_plan = 4;
 
+  reserved 13;
+  reserved "dummy_part";
+
```
pleroy: Remove.

PR #1235, opened 2017-03-09
Give the PileUp a real psychohistory so that NudgeParts doesn't lag behind

eggrobin:
And document a new bug.

### DIFF
--- a/ksp_plugin/pile_up.cpp
+++ b/ksp_plugin/pile_up.cpp
@@ -127,10 +127,16 @@ void PileUp::AdvanceTime(
     Ephemeris<Barycentric>::FixedStepParameters const& fixed_step_parameters,
     Ephemeris<Barycentric>::AdaptiveStepParameters const&
         adaptive_step_parameters) {
+  CHECK_GE(psychohistory_->Size(), 1);
+  CHECK_LE(psychohistory_->Size(), 2);
+
+  if (psychohistory_->Size() == 2) {
+    // Remove the non-authoritative point.
+    psychohistory_->ForgetAfter(psychohistory_->Begin().time());
+  }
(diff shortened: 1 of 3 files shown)


ksp_plugin/pile_up.cpp
```diff
@@ -164,13 +167,17 @@ void PileUp::AdvanceTime(
...
   for (; it != psychohistory_->End(); ++it) {
-    AppendToPartTails(it, /*authoritative=*/true);
+    AppendToPartTails(it,
+                      /*authoritative=*/it != psychohistory_->last() ||
+                                        last_point_is_authoritative);
   }
-  if (!prolongation.Empty()) {
-    CHECK_EQ(prolongation.Size(), 2);
-    AppendToPartTails(prolongation.last(), /*authoritative=*/false);
+  if (last_point_is_authoritative) {
```
pleroy: Let's have a `last_authoritative` like we have for `Vessel` and use it here.

ksp_plugin/pile_up.cpp
```diff
@@ -127,28 +127,31 @@ void PileUp::AdvanceTime(
     Ephemeris<Barycentric>::FixedStepParameters const& fixed_step_parameters,
     Ephemeris<Barycentric>::AdaptiveStepParameters const&
         adaptive_step_parameters) {
-  CHECK_EQ(psychohistory_->Size(), 1);
+  CHECK_GE(psychohistory_->Size(), 1);
+  CHECK_LE(psychohistory_->Size(), 2);
 
-  DiscreteTrajectory<Barycentric> prolongation;
+  if (psychohistory_->Size() == 2) {
```
pleroy: See my next comment: use `last_authoritative` here.

PR #1236, opened 2017-03-09
Do not ignore engines

eggrobin:
Engines are now taken into account correctly, and conservation of momentum is otherwise enforced. Stage separation appears to work, however it also seems to jettison the camera for a while. This needs to be looked into.

### DIFF
--- a/ksp_plugin/plugin.cpp
+++ b/ksp_plugin/plugin.cpp
@@ -487,8 +487,8 @@ void Plugin::IncrementPartIntrinsicForce(PartId const part_id,
   vessel->part(part_id)->increment_intrinsic_force(WorldToBarycentric()(force));
 }
 
-void Plugin::BeginReportingCollisions() {
-  for (not_null<Vessel*>vessel:loaded_vessels_) {
+void Plugin::PrepareToReportCollisions() {
+  for (not_null<Vessel*> vessel : loaded_vessels_) {
     // TODO(egg): we're taking the address of a parameter passed by reference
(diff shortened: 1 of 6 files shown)


ksp_plugin/plugin.hpp
```diff
@@ -187,6 +187,16 @@ class Plugin {
   virtual void IncrementPartIntrinsicForce(PartId part_id,
                                            Vector<Force, World> const& force);
 
+  // Calls |MakeSingleton| for all parts in loaded vessels, enabling the use of
+  // union-find for pile up construction.  This must be called after the calls
+  // to |IncrementPartIntrinsicForce|, and before the calls to
+  // |ReportCollision|.
+  virtual void BeginReportingCollisions();
```
pleroy: The name is a bit of a misnomer, it doesn't "begin" anything.  How about `PrepareToReportCollisions`?

ksp_plugin/plugin.cpp
```diff
@@ -492,6 +487,22 @@ void Plugin::IncrementPartIntrinsicForce(PartId const part_id,
   vessel->part(part_id)->increment_intrinsic_force(WorldToBarycentric()(force));
 }
 
+void Plugin::BeginReportingCollisions() {
+  for (not_null<Vessel*>vessel:loaded_vessels_) {
```
pleroy: Bad formatting.

ksp_plugin/plugin.hpp

PR #1241, opened 2017-03-10
Remove compatibility code

pleroy:


eggrobin:
https://github.com/mockingbirdnest/Principia/blob/master/geometry/permutation_test.cpp#L7-L9 should also be updated.

PR #1243, opened 2017-03-11
Fix the camera and selection bugs

eggrobin:
Today in "KSP is dumb": the FloatingOrigin system puts the origin where the root part of the vessel is (i.e. at top typically). Then the camera is offset from that so that it's at the barycentre of the vessel. Because putting the barycentre of the thing to which physics happens at 0 and the camera looking there would be too sane.

### DIFF
--- a/ksp_plugin/plugin.cpp
+++ b/ksp_plugin/plugin.cpp
@@ -600,20 +600,6 @@ void Plugin::AdvanceTime(Instant const& t, Angle const& planetarium_rotation) {
     Vessel& vessel = *pair.second;
     vessel.AdvanceTime();
   }
-  if (loaded_vessels_.empty()) {
-    bubble_barycentre_ = std::experimental::nullopt;
-  } else {
-    BarycentreCalculator<DegreesOfFreedom<Barycentric>, Mass>
-        bubble_barycentre_calculator;
-    for (not_null<Vessel*> const vessel : loaded_vessels_) {
-      vessel->ForAllParts([&bubble_barycentre_calculator](Part& part) {
-        part.clear_intrinsic_force();
-        bubble_barycentre_calculator.Add(part.degrees_of_freedom(),
-                                         part.mass());
-      });
-    }
-    bubble_barycentre_ = bubble_barycentre_calculator.Get();
-  }
 
   VLOG(1) << "Time has been advanced" << '\n'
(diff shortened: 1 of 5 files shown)


ksp_plugin/plugin.cpp
```diff
@@ -624,34 +613,36 @@ void Plugin::AdvanceTime(Instant const& t, Angle const& planetarium_rotation) {
 }
 
 DegreesOfFreedom<World> Plugin::GetPartActualDegreesOfFreedom(
-    PartId const part_id) const {
-  CHECK(bubble_barycentre_);
+    PartId const part_id,
+    PartId const part_at_origin) const {
+  auto const world_origin = FindOrDie(part_id_to_vessel_, part_at_origin)
+                                ->part(part_at_origin)
+                                ->degrees_of_freedom();
```
pleroy: We don't like arrows at the beginning of lines.

ksp_plugin/plugin.cpp
```diff
@@ -624,34 +613,36 @@ void Plugin::AdvanceTime(Instant const& t, Angle const& planetarium_rotation) {
...
-  return *bubble_barycentre_;
-}
-
 DegreesOfFreedom<World> Plugin::CelestialWorldDegreesOfFreedom(
-    Index const index) const {
+    Index const index,
+    PartId const part_at_origin) const {
+  auto const world_origin = FindOrDie(part_id_to_vessel_, part_at_origin)
+                                ->part(part_at_origin)
+                                ->degrees_of_freedom();
```
pleroy: Same.

ksp_plugin/plugin.hpp
```diff
@@ -221,17 +221,17 @@ class Plugin {
   virtual void AdvanceTime(Instant const& t, Angle const& planetarium_rotation);
 
   // Returns the degrees of freedom of the given part in |World|, assuming that
-  // the origin of |World| is fixed at |bubble_barycentre_|.  The part must be
-  // in a loaded vessel.
```
pleroy: Was it intentional to remove the last sentence?

eggrobin: It was, this invariant is not enforced, and has no fundamental reason to be. Eventually we will distinguish loaded and packed and we can think about that again.

PR #1245, opened 2017-03-11
Implement serialization of the Plugin and reinstate more tests

pleroy:
The integration test is not there yet.

### DIFF
--- a/ksp_plugin/plugin.cpp
+++ b/ksp_plugin/plugin.cpp
@@ -1051,6 +1051,13 @@ void Plugin::WriteToMessage(
     vessel->WriteToMessage(vessel_message->mutable_vessel());
     Index const parent_index = FindOrDie(celestial_to_index, vessel->parent());
     vessel_message->set_parent_index(parent_index);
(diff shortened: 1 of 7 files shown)


ksp_plugin/plugin.cpp
```diff
@@ -1106,25 +1096,81 @@ not_null<std::unique_ptr<Plugin>> Plugin::ReadFromMessage(
...
+        vessel_message.vessel(),
+        parent,
+        plugin->ephemeris_.get(),
+        [&part_id_to_vessel = plugin->part_id_to_vessel_](
+            PartId const part_id) {
+          CHECK_NE(part_id_to_vessel.erase(part_id), 0) << part_id;
+        });
+
+    // All vessels are kept.
+    plugin->kept_vessels_.emplace(vessel.get());
```
eggrobin: Let's save the plugin as we found it. This seems error-prone.

pleroy: Done.

PR #1250, opened 2017-03-12
Remove noisy traces

eggrobin:
And expose journalling in the logging UI.


ksp_plugin_adapter/ksp_plugin_adapter.cs
```diff
@@ -79,8 +79,11 @@ public partial class PrincipiaPluginAdapter
   [KSPField(isPersistant = true)]
   private int buffered_logging_ = 0;
 
+  // Whether a journal will be recorded when the plugin is next constructed.
   [KSPField(isPersistant = true)]
   private bool must_record_journal_ = false;
+  // Whether a journal is currently being recorded.
+  private static bool journalling_;
```
pleroy: `journaling` says the OED: https://en.oxforddictionaries.com/definition/journal.

PR #1251, opened 2017-03-12
Add Contains and fix clients.

pleroy:


base/map_util.hpp
```diff
@@ -43,6 +43,13 @@
 namespace principia {
 namespace base {
 
+// Returns true if and only if the given collection contains the given key.
+template <class Collection>
+bool Contains(Collection const& collection,
+              typename Collection::key_type const key) {
+  return collection.count(key) > 0;
```
eggrobin: Since this is templatized on the `Collection`, it may be fed a multimap; perhaps use `find`?

PR #1260, opened 2017-03-14
Advance parts

eggrobin:
Another attempt at putting those calls in the right order.

### DIFF
--- a/ksp_plugin/plugin.cpp
+++ b/ksp_plugin/plugin.cpp
@@ -585,9 +585,7 @@ void Plugin::SetPartApparentDegreesOfFreedom(
       part, world_to_apparent_bubble(degrees_of_freedom));
 }
 
-void Plugin::AdvanceTime(Instant const& t, Angle const& planetarium_rotation) {
-  VLOG(1) << __FUNCTION__ << '\n'
-          << NAMED(t) << '\n' << NAMED(planetarium_rotation);
+void Plugin::AdvanceParts(Instant const& t) {
   CHECK(!initializing_);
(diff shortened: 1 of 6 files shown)


ksp_plugin/plugin.cpp
```diff
@@ -585,9 +585,7 @@ void Plugin::SetPartApparentDegreesOfFreedom(
       part, world_to_apparent_bubble(degrees_of_freedom));
 }
 
-void Plugin::AdvanceTime(Instant const& t, Angle const& planetarium_rotation) {
-  VLOG(1) << __FUNCTION__ << '\n'
```
pleroy: May be keep the VLOG for the time?

serialization/journal.proto
```diff
@@ -100,6 +100,17 @@ message AdvanceTime {
...
+    optional AdvanceParts extension = 5120;
+  }
+  message In {
+    required fixed64 plugin = 1 [(pointer_to) = "Plugin", (is_subject) = true];
+    required double t = 2;
+  }
+  optional In in = 1;
+}
+
 message PrepareToReportCollisions {
```
pleroy: This message doesn't seem to be in alphabetical order.

ksp_plugin/plugin.cpp
eggrobin: We never actually use the VLOGs, we either journal or add ad hoc traces.

serialization/journal.proto
eggrobin: Done.

PR #1261, opened 2017-03-15
Issues with the flight plan

eggrobin:


ksp_plugin/interface_flight_plan.cpp
```diff
@@ -85,14 +85,31 @@ Burn GetBurn(Plugin const& plugin,
                                kExtensionFieldNumber;
     parameters.primary_index = extension.primary();
     parameters.secondary_index = extension.secondary();
-  }
-  if (message.HasExtension(
+  } else if (message.HasExtension(
```
pleroy: Preference for a series of `if` (not `else if`) each of which increments a counter `number_of_extensions` and a `CHECK` at the end that we have seen one extension.  This is stricter.

PR #1274, opened 2017-03-19
Try to fix the non-deterministic replay by making the ordering of maps and sets deterministic

pleroy:
The theory is that the computations (e.g., of barycentres) that are done in map/set order yield different results when the order changes due to vessels/parts landing at different addresses.  To confirm this we'll need to build a new journal.

### DIFF
--- a/ksp_plugin/identification.hpp
+++ b/ksp_plugin/identification.hpp
@@ -42,9 +42,9 @@ struct VesselByGUIDComparator {
 };
 
 template<typename T>
-using PartTo =std::map<not_null<Part*>,
-                       T,
-                       PartByPartIdComparator>;
+using PartTo = std::map<not_null<Part*>,
+                        T,
+                        PartByPartIdComparator>;
 using VesselSet = std::set<not_null<Vessel*>,
                            VesselByGUIDComparator>;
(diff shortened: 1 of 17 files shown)


pleroy:
retest this please

ksp_plugin/identification.hpp
```diff
@@ -14,5 +23,28 @@ using GUID = std::string;
...
+};
+
+// Comparator by GUID.  Useful for ensuring a consistent ordering in sets of
+// pointers to Vessels.
+struct VesselByGUIDComparator {
+  bool operator()(not_null<Vessel*> left, not_null<Vessel*> right) const;
+  bool operator()(not_null<Vessel const*> left,
+                  not_null<Vessel const*> right) const;
+};
+
```
eggrobin: Could we name the types
```cpp
std::map<not_null<Part*>, T, PartByPartIdComparator>
std::set<not_null<Vessel*>, VesselByGUIDComparator>
std::set<not_null<Vessel const*>, VesselByGUIDComparator>
```
they're a bit of a mouthful, and it's not like we want to mix with comparatorless maps or sets of the same.

pleroy: Done.

pleroy:
retest this please

PR #1276, opened 2017-03-21
Names

eggrobin:
- use the right part names `Part.partName` is always `"Part"`, `part.name` corresponds to `ProtoPartSnapshot.partName`;
- vessel renaming;
- remove a spammy serialization log;
- log on vessel destruction---this one is spammy, but I don't understand why? we should look into that;
- an improvised correction for `physicalObject`s;
- a `TODO` with a suggestion for the asteroid issue.

### DIFF
--- a/ksp_plugin/vessel.cpp
+++ b/ksp_plugin/vessel.cpp
@@ -46,6 +46,7 @@ Vessel::Vessel(GUID const& guid,
       prediction_(make_not_null_unique<DiscreteTrajectory<Barycentric>>()) {}
 
 Vessel::~Vessel() {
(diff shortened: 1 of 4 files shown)


ksp_plugin/vessel.cpp
```diff
@@ -58,6 +59,16 @@ GUID const& Vessel::guid() const {
   return guid_;
 }
 
+std::string const& Vessel::name() const {
+  return name_;
+}
+
+void Vessel::rename(std::string const& new_name) {
```
pleroy: set_name?

eggrobin: Done.

PR #1278, opened 2017-03-21
Fix a bug regarding old unloaded vessels

pleroy:
Such vessels can exist, and their parts are not part of a `PileUp` so they don't have a tail, which breaks one of our invariants.  The fix is to make sure that *all* parts are put in a singleton by `PrepareToReportCollisions` and to collect `PileUp`s for all the vessels in `FreeVesselsAndPartsAndCollectPileUps`.  This simplifies the code somewhat.


eggrobin:
retest this please

PR #1279, opened 2017-03-22
Don't spam vessel creations and deletions

eggrobin:
Most of the commits here are traces to investigate the (unrelated) docking bug, which I think I understand now.

### DIFF
--- a/ksp_plugin/plugin.cpp
+++ b/ksp_plugin/plugin.cpp
@@ -380,19 +380,19 @@ void Plugin::InsertOrKeepVessel(GUID const& vessel_guid,
   CHECK(!initializing_);
   not_null<Celestial const*> parent =
       FindOrDie(celestials_, parent_index).get();
-  {
-    bool& should_insert = inserted;
-    should_insert = !Contains(vessels_, vessel_guid);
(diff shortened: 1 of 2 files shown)


ksp_plugin/plugin.cpp
```diff
@@ -380,15 +380,19 @@ void Plugin::InsertOrKeepVessel(GUID const& vessel_guid,
...
       vessels_.emplace(vessel_guid,
                        make_not_null_unique<Vessel>(vessel_guid,
                                                     vessel_name,
                                                     parent,
                                                     ephemeris_.get(),
                                                     prediction_parameters_));
-  not_null<Vessel*> const vessel = it->second.get();
+    }
+  }
+  not_null<Vessel*> const vessel = FindOrDie(vessels_, vessel_guid).get();
```
pleroy: Isn't it a bit ugly that we are doing a lookup here when we just did a lookup or an insert.  I'd use vanilla STL:
```
auto it = vessels_.find(...);
if (it == vessels_.end()) {
   std::tie(...) = vessels_.emplace(...);
}
vessel = it->second;
```

PR #1304, opened 2017-04-02
Proper restartability of adaptive-step integrators

pleroy:


eggrobin:
I am confused by the comments at lines 279 and 113: if they are true, do we fail a check if an integration with last_step_is_exact produces only one (truncated) point? Otherwise, this is really the last state, possibly truncated, but it is only resumable if it does not result from a truncation. I don't think we should be dealing with `final_state` if last_step_is_exact.

pleroy:
please retest this

pleroy:
retest this please

PR #1330, opened 2017-04-15
Separate the manœuvre from its Frenet trihedron

eggrobin:


serialization/journal.proto
```diff
@@ -66,7 +66,10 @@ message NavigationManoeuvre {
   required double time_of_half_delta_v = 7;
   required double time_to_half_delta_v = 8;
   required XYZ inertial_direction = 9;
-  // Frenet trihedron at the beginning of the man�uvre.
+}
+
+// Frenet trihedron at the beginning of the man�uvre.
+message NavigationManoeuvreFrenetTrihedron {
   required XYZ binormal = 10;
```
pleroy: Tags 1, 2 , 3.

ksp_plugin_test/interface_test.cpp
```diff
@@ -1004,16 +1004,6 @@ TEST_F(InterfaceTest, FlightPlan) {
...
-  EXPECT_CALL(navigation_manœuvre, FrenetFrame())
-      .WillOnce(
-          Return(OrthogonalMap<Frenet<Navigation>, Barycentric>::Identity()));
-  EXPECT_CALL(*plugin_, CurrentTime()).WillOnce(Return(Instant() - 4 * Second));
-  EXPECT_CALL(*plugin_, GetPlottingFrame())
-      .WillOnce(Return(plotting_frame.get()));
-  EXPECT_CALL(*plotting_frame, ToThisFrameAtTime(Instant()))
-      .WillOnce(Return(barycentric_to_plotting));
-  EXPECT_CALL(*plotting_frame, FromThisFrameAtTime(Instant() - 4 * Second))
-      .WillOnce(Return(barycentric_to_plotting.Inverse()));
```
pleroy: Could we move this code to a new test for the new method?  It should be straightforward since the code is already here.

PR #1347, opened 2017-04-25
Modify SolarSystem to support keplerian elements

pleroy:


physics/solar_system_body.hpp
```diff
@@ -337,9 +394,72 @@ template<typename Frame>
...
+        CHECK(primary.empty()) << name;
+        primary = name;
+      }
+      auto owned_body = MakeMassiveBody(gravity_model_message(name));
+      unowned_bodies.emplace(name, owned_body.get());
+      owned_bodies.emplace(name, std::move(owned_body));
+    }
+
+    // Construct a hierarchical system rooted at the primary and add the other
+    // bodies layer by layer.
```
eggrobin: This seems a bit complicated; if I recall correctly `HierarchicalSystem` only requires that a body be added after its parent, and I am fine with requiring that the proto refer to a body only after it has been defined (that seems like good hygiene anyway).

physics/solar_system_body.hpp
```diff
@@ -81,26 +84,50 @@ void SolarSystem<Frame>::Initialize(
...
+      bool inserted;
+      std::tie(std::ignore, inserted) =
+          cartesian_initial_state_map_.emplace(body.name(), &body);
+      CHECK(inserted) << body.name();
+    }
+
+    // Check that the maps are consistent.
+    auto it1 = gravity_model_map_.begin();
+    auto it2 = cartesian_initial_state_map_.begin();
+    for (; it1 != gravity_model_map_.end() &&
```
pleroy: As discussed, this doesn't work as the two iterators are of different types.

physics/solar_system_body.hpp
```diff
@@ -81,26 +84,50 @@ void SolarSystem<Frame>::Initialize(
...
+           it2 != cartesian_initial_state_map_.end();
+         ++it1, ++it2) {
+      CHECK_EQ(it1->first, it2->first);
+      names_.push_back(it1->first);
+    }
+    CHECK(it1 == gravity_model_map_.end()) << it1->first;
+    CHECK(it2 == cartesian_initial_state_map_.end()) << it2->first;
```
pleroy: UB in my code?  This is actually safe, and the check as written is more useful than a size check as it prints out the offending body.

physics/solar_system_body.hpp
```diff
@@ -81,26 +84,50 @@ void SolarSystem<Frame>::Initialize(
...
+    auto it1 = gravity_model_map_.begin();
+    auto it2 = keplerian_initial_state_map_.begin();
+    for (; it1 != gravity_model_map_.end() &&
+           it2 != keplerian_initial_state_map_.end();
+         ++it1, ++it2) {
+      CHECK_EQ(it1->first, it2->first);
+      names_.push_back(it1->first);
+    }
+    CHECK(it1 == gravity_model_map_.end()) << it1->first;
+    CHECK(it2 == keplerian_initial_state_map_.end()) << it2->first;
```
pleroy: Same reply.

physics/solar_system_body.hpp
pleroy: At this point we have lost the order of the original proto, the map is key/ordered by name.  It's not that bad, and being able to order the bodies as we wish in the proto seems useful.

pleroy:
retest this please

PR #1349, opened 2017-04-30
Move the KSP tests to astronomy and rename the data files

pleroy:
No significant changes in this CL, it's renamings all the way down.


pleroy:
@cmr:  This pull request concludes a series of changes which make it possible to construct a solar system from keplerian elements given in data files.  This might be relevant for you.

PR #1353, opened 2017-05-01
Make it possible to construct a SolarSystem from protocol buffers

pleroy:
Also replace the `Initialize` method with a constructor.  This is preparatory work for using SolarSystem in the Plugin.


pleroy:
retest this please

PR #1356, opened 2017-05-05
Use the new solar system construction mechanisms in the plugin

pleroy:


serialization/astronomy.proto
```diff
@@ -17,7 +17,7 @@ message GravityModel {
     optional string reference_angle = 8;
     optional string angular_frequency = 9;
     // Oblate body.
-    optional string j2 = 5;
+    optional string j2 = 5;//TODO(phl):double!
```
eggrobin: Change or format the TODO properly.

pleroy:
retest this please

astronomy/epoch.hpp
```diff
@@ -36,17 +36,22 @@ CONSTEXPR_INFINITY Instant InfiniteFuture = J2000 + Infinity<Time>();
 // The Julian Date JD |days|. J2000.0 is JD 2451545.0. |days| is the number of
 // days since -4712-01-01T12:00:00,000 (Terrestrial Time, Julian calendar).
 constexpr Instant JulianDate(double days);
+constexpr double JulianDayNumber(Instant const& t);
```
eggrobin: OK for this CL, but in relation to #1355, mark for deprecation within Catalan: the Julian Dates should really just be literals much like the ISO 8601 dates are (and this won't be needed once #1355 is fixed).

ksp_plugin/interface.cpp
```diff
@@ -530,15 +534,36 @@ void principia__InsertCelestialJacobiKeplerian(
...
+  if (keplerian_elements != nullptr) {
+    serialization::InitialState::Keplerian::Body::Elements* elements =
+        initial_state.mutable_elements();
+    elements->set_eccentricity(keplerian_elements->eccentricity);
+    if (!std::isnan(keplerian_elements->semimajor_axis)) {
+      elements->set_semimajor_axis(
+          DebugString(keplerian_elements->semimajor_axis * Metre));
+    }
+    if (!std::isnan(keplerian_elements->mean_motion)) {
+      // s^-1 rad is inconvenient to parse.
```
eggrobin: Make a low-priority issue to have a proper parser for quantities.

ksp_plugin/plugin.cpp
```diff
@@ -149,127 +124,120 @@ Plugin::~Plugin() {
...
+
+  // If the system was constructed using keplerian elements, it may be the
+  // stock KSP system in which case it needs to be stabilized.
+  if (initial_state_.has_keplerian()) {
+    auto const hierarchical_system = solar_system.MakeHierarchicalSystem();
+    serialization::HierarchicalSystem message;
+    hierarchical_system->WriteToMessage(&message);
+    std::string const serialized_message = message.SerializeAsString();
+    uint64_t const system_fingerprint =
+        Fingerprint2011(serialized_message.c_str(), serialized_message.size());
```
eggrobin: I thought we wanted to include the gravity model in the fingerprint?

ksp_plugin/plugin.cpp
```diff
@@ -149,127 +124,120 @@ Plugin::~Plugin() {
...
+      uint64_t const system_fingerprint = Fingerprint2011(
+          serialized_message.c_str(), serialized_message.size());
+      LOG(INFO) << "System fingerprint after stabilization is " << std::hex
+                << std::uppercase << system_fingerprint;
+      CHECK_EQ(ksp_stabilized_system_fingerprint, system_fingerprint)
+          << "Attempt at stabilizing the KSP system failed!\n"
+          << gravity_model_.DebugString() << "\n"
+          << initial_state_.DebugString();
+      LOG(INFO) << "This is the stabilized KSP system, all hail retrobop!";
+    } else {
```
eggrobin: Add an `else if` clause that checks if this is already the retrobop system; this may happen when using a Kopernicus config that's equivalent to `StabilizeKSP` (in fact we should probably check that it is possible to write such a Kopernicus config, otherwise we're doing inconsistent things in `StabilizeKSP`).

eggrobin: It would also be nice to have a check for RealSolarSystem; that would not be in the `has_keplerian()` case though.

serialization/astronomy.proto
```diff
@@ -17,7 +17,7 @@ message GravityModel {
     optional string reference_angle = 8;
     optional string angular_frequency = 9;
     // Oblate body.
-    optional string j2 = 5;
+    optional string j2 = 5;  // TODO(phl): Change to a double.
```
eggrobin: It's probably more convenient to keep it a string, since these protos mirror KSP config nodes, and I wouldn't trust KSP's ability to keep floating-point numbers invariant under serialization/deserialization.

astronomy/epoch.hpp
pleroy: Added TODOs.

ksp_plugin/interface.cpp
pleroy: #1357, but I have been thinking of that, trust me, it's hard.

ksp_plugin/plugin.cpp
pleroy: The serialized form of the `HierarchicalSystem` has both the bodies and their keplerian elements, so I believe that this covers everything.

ksp_plugin/plugin.cpp
pleroy: 1. Done.
1. Yeah, but that's quite a bit of work, and what for?  Other than saying "congratulations, you are using RSS", I don't see that we would do anything useful.

serialization/astronomy.proto
pleroy: I won't change the KSP configs, it's just internal to the C++.  And I trust protobuf to know how to convert back and forth.  The PR is mostly ready anyway.

PR #1362, opened 2017-05-07
Parsing of MJD and JD literals

pleroy:


astronomy/date_time_body.hpp
```diff
@@ -718,26 +761,180 @@ constexpr Time TimeParser::ToTime() const {
...
   // Given correctness of the date and time parts of the string, this check
   // ensures that either both are in basic format or both are in extended
   // format.
   return CHECKING(
       contains(str, size, '-') == contains(str, size, ':'),
-      DateTime(
-          operator""_Date(str, index_of(str, size, 'T')),
-          operator""_Time(str + index_of(str, size, 'T') + 1,
-                          size - (index_of(str, size, 'T') + 1))).checked());
+      size >= 2 && str[0] == 'J' && str[1]
```
eggrobin: `str[1] == 'D'`

astronomy/date_time_body.hpp
```diff
@@ -718,26 +761,180 @@ constexpr Time TimeParser::ToTime() const {
...
+          ? contains(str, size - 1, '.')
```
eggrobin: I think this means that the parser doesn't support something like `"MJD1729."`. Is that correct? In any case, please comment to specify precisely what format is allowed, as is done for ISO dates.

astronomy/time_scales_test.cpp
```diff
@@ -360,6 +360,17 @@ TEST_F(TimeScalesTest, LunarEclipses) {
...
 
+TEST_F(TimeScalesTest, JulianDate) {
+  EXPECT_THAT("JD2451545"_TT, Eq(j2000_week));
+  EXPECT_THAT("JD2455201.00000"_TT, Eq("2010-01-04T12:00:00.000"_TT));
+}
+
+TEST_F(TimeScalesTest, ModifiedJulianDate) {
+  EXPECT_THAT("MJD0.0"_TT, Eq("1858-11-17T00:00:00"_TT));
+  EXPECT_THAT("MJD55200.0000"_TT, Eq("2010-01-04T00:00:00.000"_TT));
+  EXPECT_THAT("MJD55200.00000142361"_TT, Eq("2010-01-04T00:00:00.123"_TT));
```
eggrobin: In this implementation, nothing prevents MJD and JD from being used with UTC, [which poses problems](https://github.com/astropy/astropy/blob/master/cextern/erfa/ut1utc.c#L39-L42). I would not bother with supporting the ERFA quasi-JD convention, let's simply disallow those for UTC. That also makes it well-defined to go through `Time`.

astronomy/date_time_body.hpp
eggrobin: Also add a death test for that (and other formatting errors).

astronomy/date_time_body.hpp
pleroy: Done.

astronomy/time_scales_test.cpp
pleroy: Done.

astronomy/date_time_body.hpp
eggrobin: Still testing a character as a boolean.

astronomy/date_time.hpp
```diff
@@ -84,27 +76,51 @@ class DateTime final {
 
   constexpr Date const& date() const;
   constexpr Time const& time() const;
+  // Returns true iff the object was constructed from an ISO 8601 string.
+  constexpr bool iso() const;
```
eggrobin: Or from `BeginningOfDay`... Really what we care about in UTC is that it's not constructed from JD or MJD, not that it's constructed from an ISO 8601 string. Switch to `jd()` (which also encompasses any [variants](https://en.wikipedia.org/wiki/Julian_day#Variants), including MJD).

PR #1363, opened 2017-05-11
Construct the plugin with date literals

### LINKED ISSUE #1355: Plugin should be constructed with Julian dates

pleroy:
Fix #1355.


astronomy/date_time.hpp
```diff
@@ -47,6 +47,8 @@ class Time final {
  public:
   static constexpr Time hhmmss_ms(int hhmmss, int ms);
 
+  constexpr Time(int hour, int minute, int second, int millisecond);
```
eggrobin: Keep this and checked private, befriend TimeParser instead.

PR #809, opened 2016-01-10
First cut of elliptic orbits

eggrobin:
Accessors for other characteristics might be nice.

PR #1372, opened 2017-05-13
More elements

### LINKED ISSUE #1339: Expand KeplerOrbit to return all the useful numbers
See e.g., https://en.wikipedia.org/wiki/Conic_section#Conic_parameters and https://en.wikipedia.org/wiki/Orbital_elements.

eggrobin:
Fix #1339.
We way want to add a few more things (see the ifdefed declarations), and the numerics need investigation, but this should do the job for current usage.


base/optional_serialization.hpp
```diff
@@ -0,0 +1,31 @@
...
+  std::function<Message*()> mutable_field;
+  std::function<void()> clear_field;
+};
+
+#define OPTIONAL_FIELD(message, field)                      \
+  (::principia::base::OptionalField<                        \
+       std::remove_cv_t<std::remove_reference_t<            \
+           decltype(message->field())>>> {                  \
+       [m = (message)]() { return m->mutable_##field(); },  \
+       [m = (message)]() { m->clear_##field(); }})
```
pleroy: Put the last `})` on a separate line to help readability.

pleroy: This business of going through lambdas seems unnecessarily convoluted.  You could just have a 3-argument macro:
`#define WRITE_TO_OPTIONAL(message, field, value)` and it would not be less readable or less safe.  Heck, it could even have only 2 arguments if you ensure that `field` and `value` have the same name.

physics/kepler_orbit.hpp
```diff
@@ -18,16 +18,63 @@ using quantities::Angle;
...
 template<typename Frame>
 struct KeplerianElements final {
-  double eccentricity{};
+  // I. These elements determine the shape and size of the conic.  Two are
+  // needed, from two different numbered categories below.
+  // 1. eccentricity.
+  std::experimental::optional<double> eccentricity;
+  // The following two elements are NaN for elliptic orbits.
+  std::experimental::optional<Angle> asymptotic_true_anomaly;
+  std::experimental::optional<Angle> turning_angle;
```
pleroy: I wonder if the fields after section 1 could be a oneof {eliptic, hyperbolic}.  It's quite annoying to have all these fields that are specific to one shape or another and you have to remember which.

physics/kepler_orbit_body.hpp
```diff
@@ -157,94 +213,478 @@ KeplerOrbit<Frame>::KeplerOrbit(
...
+  elements.hyperbolic_mean_anomaly.reset();
+  if (e < 1) {
+    elements.mean_anomaly = *elements_at_epoch_.mean_anomaly +
+                            *elements_at_epoch_.mean_motion * (t - epoch_);
+  } else if (e == 1) {
     // Parabolic case.
     LOG(FATAL) << "not yet implemented";
-    base::noreturn();
   } else {
-    // Hyperbolic case.
```
pleroy: Keep this comment.

physics/kepler_orbit_body.hpp
```diff
@@ -157,94 +213,478 @@ KeplerOrbit<Frame>::KeplerOrbit(
...
-  Length const& a = *elements_at_epoch_.semimajor_axis;
+  double const& e = *elements_at_epoch_.eccentricity;
   Angle const& i = elements_at_epoch_.inclination;
   Angle const& Ω = elements_at_epoch_.longitude_of_ascending_node;
-  Angle const& ω = elements_at_epoch_.argument_of_periapsis;
-  Angle const mean_anomaly =
-      elements_at_epoch_.mean_anomaly +
-      *elements_at_epoch_.mean_motion * (t - epoch_);
-  if (eccentricity < 1) {
-    // Elliptic case.
```
pleroy: Keep this comment.

physics/kepler_orbit_body.hpp
```diff
@@ -157,94 +213,478 @@ KeplerOrbit<Frame>::KeplerOrbit(
...
+  auto& hyperbolic_excess_velocity = elements.hyperbolic_excess_velocity;
+  auto& semiminor_axis = elements.semiminor_axis;
+  auto& impact_parameter = elements.impact_parameter;
+  auto& semilatus_rectum = elements.semilatus_rectum;
+  auto& specific_angular_momentum = elements.specific_angular_momentum;
+  auto& periapsis_distance = elements.periapsis_distance;
+  auto& apoapsis_distance = elements.apoapsis_distance;
+  int const eccentricity_specifications = eccentricity.has_value() +
+                                          asymptotic_true_anomaly.has_value() +
+                                          turning_angle.has_value();
```
pleroy: Again, when reading this code I wonder if we could have some kind of a oneof class to help enforce these invariants.

physics/kepler_orbit_body.hpp
```diff
@@ -157,94 +213,478 @@ KeplerOrbit<Frame>::KeplerOrbit(
...
+  auto& specific_angular_momentum = elements.specific_angular_momentum;
+  auto& periapsis_distance = elements.periapsis_distance;
+  auto& apoapsis_distance = elements.apoapsis_distance;
+
+  CompleteConicParametersByCategory(elements, μ);
+
+  CHECK_EQ(eccentricity.has_value() + semimajor_axis.has_value() +
+               semiminor_axis.has_value() + semilatus_rectum.has_value() +
+               periapsis_distance.has_value() + apoapsis_distance.has_value(),
+           2) << elements;
```
pleroy: Expected value first.

physics/kepler_orbit_body.hpp
```diff
@@ -157,94 +213,478 @@ KeplerOrbit<Frame>::KeplerOrbit(
...
+    double const& e = *eccentricity;
+    Length const& r_ap = *apoapsis_distance;
+    semimajor_axis = r_ap / (1 + e);
+    semiminor_axis = *semimajor_axis * Sqrt(1 - Pow<2>(e));
+    impact_parameter = -*semimajor_axis * Sqrt(Pow<2>(e) - 1);
+    semilatus_rectum = r_ap * (1 - e);
+    periapsis_distance = *semimajor_axis * (1 - e);
+  } else if (semimajor_axis && semiminor_axis) {
+    Length const& a = *semimajor_axis;
+    auto const& b² = *semiminor_axis != *semiminor_axis
```
pleroy: Should not be a reference if it comes from a computation.  Same comment elsewhere in this function.

physics/kepler_orbit_body.hpp
```diff
@@ -157,94 +213,478 @@ KeplerOrbit<Frame>::KeplerOrbit(
...
+    Length const& a = *semimajor_axis;
+    Length const& r_ap = *apoapsis_distance;
+    eccentricity = r_ap / a - 1;
+    semiminor_axis = Sqrt((2 * a - r_ap) * r_ap);
+    impact_parameter = Sqrt((r_ap - 2 * a) * r_ap);
+    semilatus_rectum = r_ap * (2 - r_ap / a);
+    periapsis_distance = 2 * a - r_ap;
+  } else if (semiminor_axis && semilatus_rectum) {
+    auto const& b² = *semiminor_axis != *semiminor_axis
+                           ? -Pow<2>(*impact_parameter)
```
pleroy: The indent is odd.  Same comment elsewhere in this function.

physics/kepler_orbit_body.hpp
```diff
@@ -157,94 +213,478 @@ KeplerOrbit<Frame>::KeplerOrbit(
...
+
+template<typename Frame>
+void KeplerOrbit<Frame>::CompleteAnomalies(KeplerianElements<Frame>& elements) {
+  auto const& e = *elements.eccentricity;
+  auto& true_anomaly = elements.true_anomaly;
+  auto& mean_anomaly = elements.mean_anomaly;
+  auto& hyperbolic_mean_anomaly = elements.hyperbolic_mean_anomaly;
+  CHECK_EQ(true_anomaly.has_value() + mean_anomaly.has_value() +
+               hyperbolic_mean_anomaly.has_value(),
+           1);
```
pleroy: Expected value first.

physics/kepler_orbit_body.hpp
```diff
@@ -157,94 +213,478 @@ KeplerOrbit<Frame>::KeplerOrbit(
...
+              hyperbolic_eccentric_anomaly);
+    };
+    Angle const hyperbolic_eccentric_anomaly =
+        Bisect(hyperbolic_kepler_equation,
+               0 * Radian,
+               *hyperbolic_mean_anomaly / (e - 1));
+    true_anomaly =
+        2 * ArcTan(Sqrt(e + 1) * Sinh(hyperbolic_eccentric_anomaly / 2),
+                   Sqrt(e - 1) * Cosh(hyperbolic_eccentric_anomaly / 2));
+    mean_anomaly = -NaN<Angle>();
```
pleroy: Do we really want to use the sign of `NaN`?  Where is it read?  It can very easily be lost if someone does any arithmetic on this value.

serialization/physics.proto
```diff
@@ -116,13 +116,29 @@ message HierarchicalSystem {
 
 message KeplerianElements {
   required Frame frame = 1;
-  required double eccentricity = 2;
+  optional double eccentricity = 2;
```
pleroy: As mentioned elsewhere I think we should have a bit more structure here instead of 25 flat optional fields.  At a minimum, comments explaining the different sections/groups.

physics/kepler_orbit_test.cpp
```diff
@@ -54,6 +59,182 @@ class KeplerOrbitTest : public ::testing::Test {};
...
+// OM= 1.762814430848047e+02 W = 3.412521452280985e+02 Tp=  2444230.840535562020
+// N = 1.979556771581467e-06 MA= 2.337771065477673e+03 TA= 1.006849541017370e+02
+// A =-4.808470899553643e+08 AD= 9.999999000000000e+99 PR= 9.999999000000000e+99
+// X =-4.202547896125371e+09 Y =-1.982453908150731e+10 Z = 4.378406169173994e+09
+// VX=-2.067668772297011e+00 VY=-1.647515322877371e+01 VZ= 3.618493348898172e+00
+
+
+class KeplerOrbitTest : public ::testing::Test {
+ protected:
+  KeplerianElements<ICRFJ2000Equator> MoonElements() const {
```
pleroy: Make all these methods static.

physics/kepler_orbit_test.cpp
```diff
@@ -70,14 +251,7 @@ TEST_F(KeplerOrbitTest, EarthMoon) {
       AlmostEquals(
           4.0350323550225975e+05 * (Pow<3>(Kilo(Metre)) / Pow<2>(Second)), 1));
   Instant const date = JulianDate(2457397.500000000);
```
pleroy: Julian date literal?

physics/kepler_orbit_test.cpp
```diff
@@ -86,44 +260,1278 @@ TEST_F(KeplerOrbitTest, EarthMoon) {
...
+                         solar_system.gravity_model_message("Sun"));
+  MasslessBody const voyager1{};
+  constexpr Instant date = "2017-05-25T19:44:00,000"_TT;
+
+  Displacement<ICRFJ2000Equator> const expected_displacement(
+      {-4.202547896125371e+09 * Kilo(Metre),
+       -1.982453908150731e+10 * Kilo(Metre),
+        4.378406169173994e+09 * Kilo(Metre)});
+  Velocity<ICRFJ2000Equator> const expected_velocity(
+      {-2.067668772297011e+00 * (Kilo(Metre) / Second),
```
pleroy: Remove useless parentheses.

physics/kepler_orbit.hpp
eggrobin: As discussed, maybe Radical, but having a bunch of optionals is fine.

base/optional_serialization.hpp
eggrobin: Done.

physics/kepler_orbit_body.hpp
eggrobin: Done.

physics/kepler_orbit_body.hpp
eggrobin: Done.

physics/kepler_orbit_body.hpp
eggrobin: As discussed, not a good idea.

physics/kepler_orbit_body.hpp
eggrobin: As discussed, no (and we should eventually clean things up to consistently put expected values last).

physics/kepler_orbit_body.hpp
eggrobin: Done.

physics/kepler_orbit_body.hpp
eggrobin: Done.

physics/kepler_orbit_body.hpp
eggrobin: Same.

physics/kepler_orbit_body.hpp
eggrobin: It is there for consistency with the result of the computations from the true anomaly. This will become saner once we have Radical, for now it makes it much easier to test.

serialization/physics.proto
eggrobin: As discussed, optionals are fine. Comments would be redundant with those in the struct since this mirrors the struct, let's not replicate comments.

physics/kepler_orbit_test.cpp
eggrobin: Done.

physics/kepler_orbit_test.cpp
eggrobin: Done.

physics/kepler_orbit_test.cpp
eggrobin: They're only useless because it's `Second` which is `1.0` though...

eggrobin: We should decide which one we want... https://github.com/mockingbirdnest/Principia/pull/809#r49282153

PR #1379, opened 2017-05-14
Gregox and the bouncy planets

eggrobin:


ksp_plugin_adapter/ksp_plugin_adapter.cs
```diff
@@ -253,6 +253,7 @@ private enum PluginSource {
                                       (Vector3d)from_parent.p,
                                       copy.referenceBody,
                                       universal_time);
+    body.CBUpdate();
```
pleroy: Don't ask, don't tell.

PR #1395, opened 2017-05-24
Fix a bug where Forkable::LowerBound would sometimes return an incorrect iterator

pleroy:
In particular, if that iterator was at end we would be reading random memory.  This manifested itself by markers (notably for apoasides) drawn in the middle of nowhere and by the inability to replay some journals.

### DIFF
--- a/physics/forkable_body.hpp
+++ b/physics/forkable_body.hpp
@@ -204,13 +204,18 @@ It3rator Forkable<Tr4jectory, It3rator>::Find(Instant const& time) const {
 template<typename Tr4jectory, typename It3rator>
 It3rator Forkable<Tr4jectory, It3rator>::LowerBound(Instant const& time) const {
   It3rator iterator;
-
   Tr4jectory const* ancestor = that();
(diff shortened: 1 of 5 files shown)


physics/forkable_body.hpp
```diff
@@ -217,31 +222,47 @@ It3rator Forkable<Tr4jectory, It3rator>::LowerBound(Instant const& time) const {
...
+          ++fork_points_it;
+          if (ancestry_it == iterator.ancestry_.end()) {
+            // We didn't find an interesting fork in the ancestry, so we stop
+            // here and |NormalizeIfEnd| will return a proper |End|.
+            CHECK(fork_points_it == fork_points.end());
+            break;
+          }
+          if (!((*ancestry_it)->timeline_empty() ||
+                (*fork_points_it &&
+                 **fork_points_it == (*ancestry_it)->timeline_end()))) {
```
eggrobin: Use De Morgan to make the comment reflect the condition.

PR #1400, opened 2017-05-25
All NaNs are almost equal

eggrobin:
but some NaNs are less equal than others.
This fixes the Kepler orbit tests on Linux.


pleroy:
retest this please

PR #1409, opened 2017-05-28
I'm gonna have to sfinae the shit out of this!

pleroy:
* Use sfinae to declare functions that work for both `double` and `Quantity<D>`.
* Add FMA for quantities.
* Split off the testing of elementary functions.


quantities/elementary_functions.hpp
```diff
@@ -9,27 +9,30 @@ namespace quantities {
 // friendships otherwise become impossible to untangle.
```
eggrobin: Is this still the case?

pleroy: Moved to a distinct internal namespace.

PR #1418, opened 2017-05-30
A geometric transform to implement a pinhole camera perspective

pleroy:


geometry/perspective.hpp
```diff
@@ -0,0 +1,43 @@
...
+#include "geometry/point.hpp"
+#include "geometry/rp2_element.hpp"
+
+namespace principia {
+namespace geometry {
+namespace internal_perspective {
+
+// A perspective using the pinhole camera model.  It project a point of
+// |FromFrame| to an element of ℝP².  |ToFrame| is the frame of the camera.  In
+// that frame the camera is looking at the origin and looking at the positive
```
eggrobin: is at the origin and

geometry/perspective.hpp
```diff
@@ -0,0 +1,43 @@
...
+// z-axis.  The x- and y- axis of the camera correspond to those of ℝP².
+template<typename FromFrame, typename ToFrame, typename Scalar,
+         template<typename, typename> class LinearMap>
```
eggrobin: I'm really not convinced that this should be templatized on `Scalar`; can you find an application for something other than `Length`?

geometry/perspective.hpp
pleroy: located

geometry/perspective.hpp
pleroy: As discussed, just consistency with all the rest of the parameterization in `geometry`.

PR #1419, opened 2017-05-31
MacOSX support

lamont-granquist:
WIP WIP WIP WIP WIP

🚧  DO NOT MERGE 🚧 

some of the patches to the Makefile need to be made conditional or they'll most likely break Linux builds as well.

### DIFF
--- a/generate_version_translation_unit.sh
+++ b/generate_version_translation_unit.sh
@@ -16,7 +16,7 @@ char const Version[] =
 PLATFORM=$(uname -s)
 if [ "$PLATFORM" == "Darwin" ]; then
   echo -e "$VERSION_TEMPLATE" |
-  sed "s/%%DATE%%/`date -j -u -f \"%Y-%m-%d %H:%M:%S %z\" \"$(git log -1 --format=%cd --date=iso)\"`/" |
+  sed "s/%%DATE%%/`date -j -u -f \"%Y-%m-%d %H:%M:%S %z\" -u +%Y-%m-%dT%H:%M:%SZ \"$(git log -1 --format=%cd --date=iso)\"`/" |
   sed "s/%%VERSION%%/`git describe --tags --always --dirty --abbrev=40 --long`/" > $TEMPORARY_FILE
 else
   echo -e "$VERSION_TEMPLATE" |
--- a/generate_version_translation_unit.sh
+++ b/generate_version_translation_unit.sh
@@ -13,9 +13,17 @@ char const Version[] =
 }  // namespace base
 }  // namespace principia"
 
-echo -e "$VERSION_TEMPLATE" | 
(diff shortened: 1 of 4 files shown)


pleroy:

install_deps.sh
```diff
@@ -10,7 +11,7 @@ BASE_FLAGS="-fPIC -O3 -g"
 
 PLATFORM=$(uname -s)
 if [ "$PLATFORM" == "Darwin" ]; then
-    C_FLAGS="$BASE_FLAGS -mmacosx-version-min=10.7 -arch i386"
+    C_FLAGS="$BASE_FLAGS -mmacosx-version-min=10.7 -arch x86_64"
```
lamont-granquist: this is the only line in this file that is strictly necessary to fix for MacOSX -- all the rest of the diff in `install_deps.sh` here i just found more useful, but it could all be dropped.

generate_version_translation_unit.sh
```diff
@@ -13,9 +13,17 @@ char const Version[] =
 }  // namespace base
 }  // namespace principia"
 
-echo -e "$VERSION_TEMPLATE" | 
-sed "s/%%DATE%%/`date -d $(git log -1 --format=%cd --date=iso-strict) -u +%Y-%m-%dT%H:%M:%SZ`/" |
-sed "s/%%VERSION%%/`git describe --tags --always --dirty --abbrev=40 --long`/" > $TEMPORARY_FILE
+PLATFORM=$(uname -s)
+if [ "$PLATFORM" == "Darwin" ]; then
+  echo -e "$VERSION_TEMPLATE" |
+  sed "s/%%DATE%%/`date -j -u -f \"%Y-%m-%d %H:%M:%S %z\" \"$(git log -1 --format=%cd --date=iso)\"`/" |
```
lamont-granquist: mac `date` doesn't have nice iso format flags, so it has to get done manually here

eggrobin: The format seems inconsistent with the the one we are using, which is really ISO 8601 extended format, that is `%Y-%m-%dT%H:%M:%SZ`; note the T, and note the timezone written simply as Z (with no space) since we give UTC. 
Example in our Catalan build:
````cpp
char const BuildDate[] = "2017-05-26T12:44:02Z";

File: install_deps.sh (hunk 1 above)
Comment (eggrobin) [inline install_deps.sh]: The other changes look good to me too.

**Comment by eggrobin** [comment]
Looks good, as you said the Makefile changes need to be made conditional.

File: generate_version_translation_unit.sh (hunk 2 above)
Comment (lamont-granquist) [inline generate_version_translation_unit.sh]: So that string is what parses the `git log` output and its right because `--date=iso` gives that.  The problem is that it needs the proper format string for output, which I appear to have dropped

**Comment by lamont-granquist** [comment]
```
char const BuildDate[] = "2017-05-28T16:27:13Z";
```

eggrobin:
Tests pass on linux when merging this, so LGTM.

PR #1430, opened 2017-06-05
Make the beach towel a bit smaller

eggrobin:
Also removed the build date, since it's not a build date anymore (it's just the commit date), so it doesn't tell us anything that the commit number wouldn't.


pleroy:
retest this please

PR #1431, opened 2017-06-05
A function to check if a point is hidden by a sphere

pleroy:


geometry/sphere.hpp
```diff
@@ -0,0 +1,37 @@
...
+
+template<typename Scalar, typename Frame>
+class Sphere {
+ public:
+  Sphere(Point<Vector<Scalar, Frame>> const& centre,
+         Scalar const& radius);
+
+  Point<Vector<Scalar, Frame>> const& centre() const;
+  Scalar const& radius() const;
+  Product<Scalar, Scalar> const radius²() const;
```
eggrobin: Nice identifer

Issue #1414, opened 2017-05-30
Body Targeting


NathanKell:
For those of us who do not have Jebediah delusions of cowboying, the ability to target celestial bodies (and thus rely on MechJeb's relative inclination indicator and, more importantly, launch into plane of target functionality) would be useful.

eggrobin:
This could also be used for more unification of reference frame selection in the vein of #1411, though it's not clear if we want that particular behaviour (when launching to the Moon's plane, it may still be useful to see the velocities in the ECI or ECEF frame).

NathanKell:
During early launch I very definitely want to stay in ECEF if possible, yes.

eggrobin:
Upon further reflection, the inability to target the celestial orbited by the current celestial (e.g. Earth from the Moon) means unification with our reference frame selection would be very clunky and lopsided. I will allow celestial targeting, but not interact with it (besides disabling the rendering of the occasional random patched conic thing).
As a separate issue, we may want to display nodes in the MCEA and EMB frames, they make sense there (the reference plane is physically significant). They may even make sense in the ECEF frame (to aim for equatorial orbits), but this sounds less obviously useful.

eggrobin:
In the future, it may make sense for Principia to support a "warp to alignment with plane" feature, or indeed a "warp to [astronomical event here]"; this is something we can do (we have tests that look for eclipses), and it seems useful.
It is still a good idea to allow celestial targeting for the occasional conic-based mod that will depend on it however.

NathanKell:
Agree, that'd be rather useful for a bunch of reasons!

PR #1433, opened 2017-06-05
Display nodes in orbital frames

eggrobin:
As mentioned in #1414, it makes sense to display nodes in, e.g., the EMB and MCEA frames like we do in the target LVLH frame.

### DIFF
--- a/ksp_plugin_adapter/ksp_plugin_adapter.cs
+++ b/ksp_plugin_adapter/ksp_plugin_adapter.cs
@@ -1546,6 +1546,32 @@ public partial class PrincipiaPluginAdapter
             vessel    : null,
             celestial : celestial);
       }
+      var frame_type = plotting_frame_selector_.get().frame_type;
+      if (frame_type ==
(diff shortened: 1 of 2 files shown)


pleroy:
retest this please

ksp_plugin_adapter/ksp_plugin_adapter.cs
```diff
@@ -1604,6 +1630,32 @@ private System.Collections.IEnumerator
...
+            MapNodePool.NodeSource.PREDICTION,
+            vessel    : null,
+            celestial : primary);
+        map_node_pool_.RenderAndDeleteMarkers(
+            descending_nodes_iterator,
+            MapObject.ObjectType.DescendingNode,
+            MapNodePool.NodeSource.PREDICTION,
+            vessel    : null,
+            celestial : primary);
+      }
```
eggrobin: As discussed, things change subtly, so let's not do that right now; eventually this needs refactoring, but it's nontrivial.

PR #1436, opened 2017-06-07
Render the trajectory of the target in target-mobile frames

### LINKED ISSUE #1424: Plot the trajectory of the target vessel in frames that do not fix the target
Feature request by @scimas and @maccollo.
We should not plot any approach/node markers, because the actual rendez-vous should be done in a target-fixed frame, not the stock way. However, in order to get in similar orbits it helps to see things in a reference frame where there are orbits.

Example provided by @maccollo: this transfer (2 burns) from a near-ellipse to another via third-body magic. Once the vessel has reached something resembling the target orbit, the phase change should be done in the target LVLH frame, but this first part of the flight seems essentially unplannable outside of the MCI frame.

Example provided by @scimas:
> my ship is in 80km circular orbit and my target is something like 5Mm x 3Mm orbit; having a target Ap and Pe visible helps in that case, right? Or am I wrong here too and the target frame takes care of it and I just have to figure it out?

eggrobin:
Fix #1424.

### DIFF
--- a/ksp_plugin_adapter/ksp_plugin_adapter.cs
+++ b/ksp_plugin_adapter/ksp_plugin_adapter.cs
@@ -879,6 +879,12 @@ public partial class PrincipiaPluginAdapter
 
       if (ready_to_draw_active_vessel_trajectory) {
         plugin_.UpdatePrediction(active_vessel.id.ToString());
+        string target_id =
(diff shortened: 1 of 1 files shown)


eggrobin:
For illustration:
Target LVLH at Kerbin, no target trajectory:
screenshot71
KCI:
screenshot72
KCKF:
screenshot73

PR #1455, opened 2017-07-04
An API for segment hiding

pleroy:


geometry/perspective.hpp
```diff
@@ -32,9 +33,17 @@ class Perspective final {
...
       Point<Vector<Scalar, FromFrame>> const& point) const;
 
+  // Returns true iff the |point| is hidden by the |sphere| in this perspective.
   bool IsHiddenBySphere(Point<Vector<Scalar, FromFrame>> const& point,
                         Sphere<Scalar, FromFrame> const& sphere) const;
 
+  // Returns the (sub)segments of |segment| that are visible in this perspective
+  // after taking into account the hiding by |sphere|.  The returned vector has
+  // 0, 1, or 2 elements.
+  std::vector<Segment<Point<Vector<Scalar, FromFrame>>>> VisibleSegments(
```
eggrobin: if a Segment<Vector> is a pair of Point<Vector>, this should be a Segment<Vector>, otherwise you have Point<Point<Vector>> which is nonsense.

geometry/point.hpp
```diff
@@ -81,6 +81,9 @@ class Point final {
   friend class geometry::BarycentreCalculator;
 };
 
+template<typename Vector>
+using Segment = std::pair<Point<Vector>, Point<Vector>>;
```
eggrobin: I'm somewhat reluctant to put that in point.hpp; for now let's keep it in perspective which is the only place where we use it, and see how things evolve from there

geometry/perspective.hpp
eggrobin: Also, as discussed, this should *really* use Positions anyway, it's getting way too clumsy for no good reason (a job for another CL though).

pleroy: Done.

geometry/point.hpp
pleroy: Reluctantly moved.

PR #1461, opened 2017-07-09
A function to compute the parts of a segment that are hidden by a sphere

pleroy:


geometry/perspective_test.cpp
```diff
@@ -143,6 +146,259 @@ TEST_F(PerspectiveTest, IsHiddenBySphere) {
...
+        World::origin +
+        Displacement<World>({-3 * Metre, 21 * Metre, 32 * Metre});
+    Segment<Displacement<World>> segment{p1, p2};
+    EXPECT_THAT(perspective.VisibleSegments(segment, sphere),
+                ElementsAre(segment));
+  }
+
+  // A segment tangent to the sphere when seen from the camera.  Need a 1-ulp
+  // tolerance to make sure that the segment doesn't get cut in two because of
+  // rounding.
```
eggrobin: Do this test with -ε and check that it then intersects.

geometry/perspective_test.cpp
```diff
@@ -143,6 +146,259 @@ TEST_F(PerspectiveTest, IsHiddenBySphere) {
...
+  // The sphere is at the origin and has unit radius.
+  Sphere<Length, World> const sphere(World::origin,
+                                     /*radius=*/1 * Metre);
+
+  // The equation of the line KP in the plane x-z is:
+  //   x - 3 * sqrt(11) z + 10 = 0
+  // This is used to compute many of the expected values below.
+
+  // A segment away from the sphere with x > 0.
+  {
```
eggrobin: Make those individual tests within a fixture that contains the stuff you are reusing; possibly put that in a new fixture in a new file.

pleroy: Done.

geometry/perspective_test.cpp
pleroy: Done.

PR #1474, opened 2017-07-18
Unpacked pile ups are straggling

### LINKED ISSUE #1473: Unpacked vessels are off by one frame
Initially reported by @scimas, initially fused with #1416.
Especially visible in low orbit with a nearby (loaded) packed vessel in the vicinity (say about 300 m away). The vessel will appear to jump along its orbit with respect to the active vessel when going in and out of warp, which is due to the active vessel being packed and unpacked while the other vessel remains packed.

I suspect that what is happening is that unpacked vessels have their parts interpreted by KSP as being one frame in the past during `FixedUpdate`, and in the present in `WaitForFixedUpdate` (after the physics engine runs), whereas Principia assumes they are in the present in `FixedUpdate`, and in the future in `WaitForFixedUpdate`.

This requires looking at the `PileUp` logic again to reverse its timing with respect to the `Plugin`.

eggrobin:
Fix #1473.

### DIFF
--- a/journal/player_test.cpp
+++ b/journal/player_test.cpp
@@ -92,7 +92,7 @@ TEST_F(PlayerTest, DISABLED_Debug) {
   // An example of how journaling may be used for debugging.  You must set
   // |path| and fill the |method_in| and |method_out_return| protocol buffers.
   std::string path =
-      R"(C:\Users\robin\Projects\Kerbal Space Program\KSP_win64 1.2.2\glog\Principia\JOURNAL.20170719-193221)";  // NOLINT
+      R"(\\venezia.mockingbirdnest.com\Namespaces\Public\Public Mockingbird\Principia\Crashes\1441\replication\JOURNAL.20170616-195725)";  // NOLINT
   Player player(path);
   int count = 0;
   while (player.Play()) {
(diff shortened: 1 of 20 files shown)


ksp_plugin/pile_up.hpp
```diff
@@ -88,6 +88,8 @@ class PileUp {
   // |DeformPileUpIfNeeded|.
   void NudgeParts() const;
 
+  Instant const& time();
```
pleroy: Document this method, in particular wrt authoritative-ness.

ksp_plugin/plugin.cpp
```diff
@@ -503,10 +505,10 @@ void Plugin::FreeVesselsAndPartsAndCollectPileUps() {
   // the same subset.
   for (auto const& pair : vessels_) {
     Vessel& vessel = *pair.second;
-    vessel.ForSomePart([this](Part& first_part) {
+    vessel.ForSomePart([Δt, this](Part& first_part) {
```
pleroy: Capture `this` before the other parameters (just like for method calls).

ksp_plugin/plugin.cpp
```diff
@@ -1317,11 +1318,12 @@ void Plugin::AddPart(not_null<Vessel*> const vessel,
...
-                                         name,
-                                         mass,
-                                         degrees_of_freedom,
-                                         std::move(deletion_callback));
+  auto part = make_not_null_unique<Part>(
+      part_id,
+      name,
+      mass,
+      degrees_of_freedom,
+      std::move(deletion_callback));
```
pleroy: Revert this change.

ksp_plugin/plugin.hpp
```diff
@@ -213,7 +214,7 @@ class Plugin {
   // Advances time on the pile ups to |t|, filling the tails of all parts up to
```
pleroy: Fix the comment.

ksp_plugin/plugin.hpp
```diff
@@ -242,6 +243,8 @@ class Plugin {
   // |Planetarium.InverseRotAngle| is in degrees.
   virtual void AdvanceTime(Instant const& t, Angle const& planetarium_rotation);
 
+  virtual void CatchUpVessel(GUID const& vessel_guid);
```
pleroy: Comment.

serialization/journal.proto
```diff
@@ -105,13 +105,23 @@ message AdvanceTime {
   optional In in = 1;
 }
 
-message AdvanceParts {
+message CatchUpLaggingVessels {
   extend Method {
-    optional AdvanceParts extension = 5120;
+    optional CatchUpLaggingVessels extension = 5120;
```
pleroy: I'd rather use a new tag here to avoid future confusion.

ksp_plugin/pile_up.hpp
eggrobin: Done.

ksp_plugin/plugin.cpp
eggrobin: Done.

ksp_plugin/plugin.cpp
eggrobin: Done.

ksp_plugin/plugin.hpp

ksp_plugin/plugin.hpp

serialization/journal.proto
eggrobin: As discussed, it's fine.

journal/player_test.cpp
```diff
@@ -106,25 +106,22 @@ TEST_F(PlayerTest, DISABLED_Debug) {
   LOG(ERROR) << "Last successful method out/return: \n"
               << player.last_method_out_return().DebugString();
 
-#if 0
+#if 1
```
pleroy: 0

PR #1477, opened 2017-07-21
A lunar reminder to update Principia

eggrobin:
Looks like this in-game: 
screenshot104

### DIFF
--- a/ksp_plugin_adapter/ksp_plugin_adapter.cs
+++ b/ksp_plugin_adapter/ksp_plugin_adapter.cs
@@ -16,9 +16,9 @@ public partial class PrincipiaPluginAdapter
     : ScenarioModule,
       WindowRenderer.ManagerInterface {
       
-  private const String чебышёв_release_name = "Чебышёв";
-  private const int чебышёв_lunation_number = 218;
-  private DateTimeOffset чебышёв_release_date =
+  private const String next_release_name = "Чебышёв";
+  private const int next_release_lunation_number = 218;
+  private DateTimeOffset next_release_date =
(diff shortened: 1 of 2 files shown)


ksp_plugin_adapter/ksp_plugin_adapter.cs
```diff
@@ -1755,6 +1760,12 @@ private System.Collections.IEnumerator
       if (!PluginRunning()) {
         UnityEngine.GUILayout.TextArea(text : "Plugin is not started");
       }
+      if (DateTimeOffset.Now > чебышёв_release_date) {
+        UnityEngine.GUILayout.TextArea(
+            "PSA: the new moon of lunation number " + чебышёв_lunation_number +
```
pleroy: Announcement: ...

ksp_plugin_adapter/ksp_plugin_adapter.cs
```diff
@@ -15,6 +15,11 @@ namespace ksp_plugin_adapter {
 public partial class PrincipiaPluginAdapter
     : ScenarioModule,
       WindowRenderer.ManagerInterface {
+      
+  private const String чебышёв_release_name = "Чебышёв";
```
pleroy: Call this `next_release_name`, etc., so that we don't need to rename the identifiers when doing a release.

eggrobin: Done.

ksp_plugin_adapter/ksp_plugin_adapter.cs
eggrobin: Done.

Issue #973, opened 2016-03-30
Move the hiding computations and 2-D rendering to C++ code


pleroy:

pleroy:
Worked on this in #1405, #1412, #1418, #1423, #1431, #1437, #1438, #1439, #1455, #1461, #1462, #1463, #1464.

pleroy:
Closing as we have moved all the code we want to move to C++.

PR #1479, opened 2017-07-26
Changes to planetarium construction and hiding

pleroy:
This has been tested in game, but it's only the C++ code for now.

Yet another bit of #973.


geometry/perspective_body.hpp
```diff
@@ -186,39 +205,67 @@ Perspective<FromFrame, ToFrame, Scalar, LinearMap>::VisibleSegments(
...
+  // part that is on the other side of T with respect to S.
+  // For each solution δ of the above quadratic equation, this loop computes PH
+  // and obtains the values of λ and τ.
   std::vector<double> λs;
-  λs.reserve(4);
-
-  // For each solution of the above quadratic equation, compute the value of λ,
-  // if any.
-  Vector<Scalar, FromFrame> const AB = B - A;
+  λs.reserve(2);
```
pleroy: Should be 4.

PR #1482, opened 2017-07-29
Generated code for interchange of booleans

pleroy:


tools/journal_proto_processor.cpp
```diff
@@ -854,9 +856,23 @@ void JournalProtoProcessor::ProcessInterchangeMessage(
             field_cxx_assignment_fn_[field_descriptor](
                 "m.", serialize_member_name));
 
-    cs_interface_type_declaration_[descriptor] +=
-        "  public " + field_cs_type_[field_descriptor] + " " +
-        field_descriptor_name + ";\n";
+    if (field_cs_private_type_[field_descriptor].empty()) {
```
eggrobin: Use `field_cs_marshal_[descriptor].empty()` instead: if marshalling is involved, it's probably never going to be blittable, if it's blittable, it doesn't need marshalling.
Add a check that they are either both empty or both nonempty.

PR #1491, opened 2017-07-31
Avoid comparing RP2Points, compare Positions instead

pleroy:


ksp_plugin/planetarium.cpp
```diff
@@ -62,12 +63,17 @@ RP2Lines<Length, Camera> Planetarium::PlotMethod0(
       continue;
     }
 
-    // Create a new ℝP² line when two segments are not consecutive.
-    if (rp2_lines.empty() || rp2_lines.back().back() != rp2_first) {
+    // Create a new ℝP² line when two segments are not consecutive.  Don't
+    // compare ℝP² points for equality, that's expensive.
+    bool const are_consecutive =
+        previous_position && *previous_position == plottable_segment.first;
```
eggrobin: Why not just operator== on the optional and the right-hand-side, using overloads 19-30 of http://en.cppreference.com/w/cpp/utility/optional/operator_cmp?

PR #1510, opened 2017-08-06
Say goodbye to the perspective on catalytic activity

pleroy:


geometry/perspective.hpp
```diff
@@ -11,78 +11,78 @@
...
+template<typename Frame>
+using Segments = std::vector<Segment<Frame>>;
 
 // A perspective using the pinhole camera model.  It project a point of
 // |FromFrame| to an element of ℝP².  |ToFrame| is the frame of the camera.  In
 // that frame the camera is located at the origin and looking at the positive
 // z-axis.  The x- and y- axis of the camera correspond to those of ℝP².
-template<typename FromFrame, typename ToFrame, typename Scalar,
+template<typename FromFrame, typename ToFrame,
          template<typename, typename> class LinearMap>
```
eggrobin: The template parameter LinearMap is no longer used (since `RigidTransformation` implies it is orthogonal), drop it.

PR #1512, opened 2017-08-06
1.3

### LINKED ISSUE #1502: Update to KSP 1.3
> <@ferram4> Barring a tester coming back with a call of "Hold Hold Hold" there will be a FAR release before the eclipse.

RealismOverhaul intends to keep up-to-date with FAR.

We should get ourselves in a state where we're compatible with 1.3; the changes should be minor, perhaps we can even release Чебышёв for both if we use the intersection of the old and new APIs or `#if` things.

eggrobin:
Supports building for either version of KSP.
Fix #1502.


pleroy:
retest this please

Issue #1351, opened 2017-04-30
Accumulate gravitational interactions in DoublePrecision


eggrobin:
Different behaviour due to non-associativity being amplified by the uncorrected stock system's terrible Ляпунов exponent is expected, but Aardvark being stabler than Kerbol which is stabler than Sun points to a systematic error due to cancellation.

pleroy:
Tried this, and it doesn't help, the system remains sensitive to the order.

pleroy:
It turns out that using compensated summation (`Increment` or `Decrement`) is not correct: these methods are only valid if the addend is smaller that the current value.  For accelerations we don't have any guarantees to that effect, and it's clear that reordering the body will affect the final result.  Experiments with _Mathematica_ show that reordering the bodies results in an error which is of 52 bits are the beginning (fine) but which occasionally jumps to 40 bits.  When these errors accumulate the trajectories of the planets end up being affected significantly (remember that the stock system is very unstable).

Using full double precision fixes the problem, but we probably don't want to pay the cost of 10 additions/subtractions instead of 1 in the inner loop of the force computation.  For a stable solar system these inaccuracies have no effect whatsoever.

pleroy:
Code pushed to [pleroy/1351](https://github.com/pleroy/Principia/tree/1351).

oscardssmith:
Would either pairwise summation, or Kahan summation help this?

pleroy:
See the reference to "compensated summation" in the comment above, which is another name for "Kahan summation".  Kahan summation can yield large relative errors for ill-conditioned sums.  That's one of these cases.

oscardssmith:
And pairwise isn't helping enough?

pleroy:
Pairwise is less precise than Kahan and has the same conditioning properties.

The numerical analysis above was mostly done to check that the sensitivity to order doesn't hide a bug.  It doesn't.  Done.  The problem of accurately predicting the Kerbin system over long periods of time is spectacularly uninteresting as that system is extremely unstable (on a scale of months in the stock game).

PR #1521, opened 2017-08-12
Add new functions to DoublePrecision

pleroy:
A useful bit that came out of the analysis of #1351.


eggrobin:
Needs testing. We already have a test for some replicated code in https://github.com/mockingbirdnest/Principia/blob/master/numerics/double_precision_test.cpp#L173-L175.

PR #1524, opened 2017-08-12
A limit on the number of points emitted by PlotMethod2

### LINKED ISSUE #1522: We need a maximum number of points for the Чебышёв plot method 2
And therefore we should plot the history in reverse so that it times out in the distant past rather than leaving a hole before the present.

eggrobin:
Fix #1522.

### DIFF
--- a/ksp_plugin/planetarium.cpp
+++ b/ksp_plugin/planetarium.cpp
@@ -181,8 +181,8 @@ RP2Lines<Length, Camera> Planetarium::PlotMethod2(
 
   goto estimate_tan²_error;
 
-  while (direction * (previous_time - final_time) < Time{} &&
-         steps_accepted < 10'000) {
+  while (steps_accepted < 10'000 &&
+         direction * (previous_time - final_time) < Time{}) {
     do {
       // One square root because we have squared errors, another one because the
       // errors are quadratic in time (in other words, two square roots because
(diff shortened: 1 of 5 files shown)


ksp_plugin/planetarium.cpp
```diff
@@ -41,7 +42,8 @@ Planetarium::Planetarium(
 RP2Lines<Length, Camera> Planetarium::PlotMethod0(
     DiscreteTrajectory<Barycentric>::Iterator const& begin,
     DiscreteTrajectory<Barycentric>::Iterator const& end,
-    Instant const& now) const {
+    Instant const& now,
+    bool const reverse) const {
```
pleroy: Maybe `/*reverse*/` to make it clear that it's unused.

ksp_plugin/planetarium.cpp
```diff
@@ -174,7 +181,8 @@ RP2Lines<Length, Camera> Planetarium::PlotMethod2(
 
   goto estimate_tan²_error;
 
-  while (previous_time < final_time) {
+  while (steps_accepted < 10'000 &&
```
pleroy: Make `10'000` a `constexpr` in the unnamed namespace.

ksp_plugin/planetarium.cpp
eggrobin: Done.

PR #1533, opened 2017-08-19
Use forks to represent the psychohistory in PileUp

pleroy:
First step towards solving #1513.

### DIFF
--- a/ksp_plugin/pile_up.cpp
+++ b/ksp_plugin/pile_up.cpp
@@ -152,27 +152,27 @@ void PileUp::AdvanceTime(Instant const& t) {
     if (history_->last().time() < t) {
       // Do not clear the |fixed_instance_| here, we will use it for the next
       // fixed-step integration.
+      // TODO(phl): Consider not setting |last_point_only| below as we would be
+      // fine with multiple points in the |psychohistory_| once all the classes
+      // has been changed.
       CHECK(ephemeris_->FlowWithAdaptiveStep(
                 psychohistory_,
                 Ephemeris<Barycentric>::NoIntrinsicAcceleration,
                 t,
                 adaptive_step_parameters_,
                 Ephemeris<Barycentric>::unlimited_max_ephemeris_steps,
-                /*last_point_only=*/true));//TODO(phl):why?
+                /*last_point_only=*/true));
     }
   } else {
     // Destroy the fixed instance, it wouldn't be correct to use it the next
     // time we go through this function.  It will be re-created as needed.
     fixed_instance_ = nullptr;
-    //TODO(phl):comment
-    // We make the existing last point authoritative, i.e. we do not remove it.
-    // If it was already authoritative nothing happens, if it was not, we
-    // integrate on top of it, and it gets appended authoritatively to the part
-    // tails.
+    // We make the |psychohistory_|, if any, authoritative, i.e. append it to
+    // the end of the |history_|. We integrate on top of it, and it gets
+    // appended authoritatively to the part tails.
     auto const psychohistory_end = psychohistory_->End();
     auto it = psychohistory_->Fork();
-    ++it;
-    for (; it != psychohistory_end; ++it) {
+    for (++it; it != psychohistory_end; ++it) {
       history_->Append(it.time(), it.degrees_of_freedom());
     }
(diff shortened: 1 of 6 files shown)


ksp_plugin/pile_up.cpp
```diff
@@ -261,27 +266,56 @@ PileUp PileUp::ReadFromMessage(
...
+            ephemeris));
+    }
+    // Fork a psychohistory for compatibility if there is a non-authoritative
+    // point.
+    if (pile_up->history_->Size() == 2) {
+      Instant const history_begin_time = pile_up->history_->Begin().time();
+      pile_up->psychohistory_ =
+          pile_up->history_->NewForkWithCopy(history_begin_time);
+      pile_up->history_->ForgetAfter(history_begin_time);
+    }
```
eggrobin: This means it is possible to have no psychohistory, if the history size was 1; on the other hand, AdvanceTime used to check that the history size was in [1,2], and now checks that the psychohistory is non-null. Always fork a psychohistory (I don't think we can easily make psychohistory_ not_null; if we can we should).

Add a test to exercise this (deserializing a pre-Cesàro message with a history of size 1, and calling AdvanceTime on it).

ksp_plugin_test/pile_up_test.cpp
```diff
@@ -68,7 +69,7 @@ class TestablePileUp : public PileUp {
   }
 
   not_null<DiscreteTrajectory<Barycentric>*> psychohistory() const {
-    return psychohistory_.get();
+    return check_not_null(psychohistory_);
```
eggrobin: Use the implicit conversion.

ksp_plugin/pile_up.cpp
```diff
@@ -133,72 +134,76 @@ void PileUp::DeformPileUpIfNeeded() {
...
+    CHECK_LT(history_->last().time(), t);
     ephemeris_->FlowWithFixedStep(t, *fixed_instance_);
-    if (psychohistory_->last().time() < t) {
+    psychohistory_ = history_->NewForkAtLast();
+    if (history_->last().time() < t) {
       // Do not clear the |fixed_instance_| here, we will use it for the next
       // fixed-step integration.
+      // TODO(phl): Consider not setting |last_point_only| below as we would be
+      // fine with multiple points in the |psychohistory_| once all the classes
+      // has been changed.
```
pleroy: Done.

ksp_plugin_test/pile_up_test.cpp
pleroy: Done.

ksp_plugin/pile_up.cpp
pleroy: Done.

ksp_plugin_test/pile_up_test.cpp
```diff
@@ -596,6 +597,50 @@ TEST_F(PileUpTest, Serialization) {
...
+    }
+    if (part_id == part_id2_) {
+      return &p2_;
+    }
+    LOG(FATAL) << "Unexpected part id " << part_id;
+    base::noreturn();
+  };
+  auto p = PileUp::ReadFromMessage(message, part_id_to_part, &ephemeris);
+
+  EXPECT_CALL(ephemeris, FlowWithAdaptiveStep(_, _, _, _, _, _))
```
eggrobin: Are we actually *expecting* this call? it seems we should use `ON_CALL` instead of pretty much all instances of `EXPECT_CALL` in this file. That is probably a job for a later pull request though.

ksp_plugin_test/pile_up_test.cpp
```diff
@@ -596,6 +597,50 @@ TEST_F(PileUpTest, Serialization) {
...
+                         astronomy::J2000,
+                         DefaultProlongationParameters(),
+                         DefaultHistoryParameters(),
+                         &ephemeris);
+
+  serialization::PileUp message;
+  pile_up.WriteToMessage(&message);
+
+  // Clear the children to simulate pre-Cesàro serialization.
+  message.mutable_history()->clear_children();
```
eggrobin: Check that `message.history().timeline_size()` is 1 so that it is clear which code path is being tested.

PR #1535, opened 2017-08-22
Use history and psychohistory in Vessel

pleroy:


ksp_plugin/part.cpp
```diff
@@ -82,16 +82,32 @@ Part::degrees_of_freedom() const {
...
-  return *psychohistory_;
+  // Make sure that we skip the fork, which may be the point of the prehistory.
+  auto it = psychohistory_->Fork();
+  return ++it;
 }
 
-bool Part::tail_is_authoritative() const {
-  return psychohistory_->Fork() == psychohistory_->last();
+DiscreteTrajectory<Barycentric>::Iterator Part::psychohistory_end() {
+  if (psychohistory_ == nullptr) {
```
eggrobin: Why can the psychohistory be null?

ksp_plugin/part.cpp
```diff
@@ -82,16 +82,32 @@ Part::degrees_of_freedom() const {
...
+}
+
+DiscreteTrajectory<Barycentric>::Iterator Part::psychohistory_begin() {
   if (psychohistory_ == nullptr) {
     psychohistory_ = history_->NewForkAtLast();
   }
-  return *psychohistory_;
+  // Make sure that we skip the fork, which may be the point of the prehistory.
+  auto it = psychohistory_->Fork();
+  return ++it;
```
eggrobin: This seems like a very confusing API: the points in [psychohistory_begin, psychohistory_end[ are missing a bit of the psychohistory (the range between the fork and psychohistory_begin) when the psychohistory is forked in the history. When treating the trajectiories as continuous, this will cause issues.

If this is really the intended behaviour, rename it to something other than begin (first_ahistorical_iterator?).

ksp_plugin/part.cpp
pleroy: Documented in `part.hpp`, to avoid reforking each time we append to the history.

ksp_plugin/part.cpp
pleroy: Ministry of Silly Names.

eggrobin:
I think this fixed #1513.

PR #1542, opened 2017-08-27
Asynchronous catch up

eggrobin:
Use the new futures to catch up asynchronously, log the time between the first promise and the first wait.

### DIFF
--- a/ksp_plugin/interface_future.cpp
+++ b/ksp_plugin/interface_future.cpp
@@ -17,9 +17,8 @@ using quantities::si::Second;
 
 namespace {
 constexpr int window_size = 500;
-std::array<Time, window_size> Δt_window;
 int window_index = 0;
-Time average_Δt;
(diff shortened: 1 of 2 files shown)


ksp_plugin/interface_future.cpp
```diff
@@ -1,16 +1,37 @@
...
+Time max_Δt = -Infinity<Time>();
+std::chrono::steady_clock::time_point time_of_last_promise_batch;
+bool last_call_was_a_promise = false;
+}  // namespace
+
 std::future<void> const* principia__FutureCatchUpVessel(
     Plugin* const plugin,
     char const* const vessel_guid) {
   journal::Method<journal::FutureCatchUpVessel> m({plugin, vessel_guid});
+  if (!last_call_was_a_promise) {
```
pleroy: Make these separate functions and make variables as local as possible.

PR #1547, opened 2017-08-30
Make the monitor accessible from C#

eggrobin:
Only a single untitled monitor for now, we may want to have something cleaner later on.

### DIFF
--- a/ksp_plugin/interface_monitor.cpp
+++ b/ksp_plugin/interface_monitor.cpp
@@ -17,15 +17,20 @@ using quantities::si::Second;
 namespace {
 std::chrono::steady_clock::time_point monitor_start_time;
 bool is_monitor_running = false;
+}  // namespace
 
-void StartMonitoring() {
+
+// No journalling to avoid overhead from that; these functions have no side
+// effects aside from logging changing and the internal state of the monitor.
+
+void principia__MonitorStart() {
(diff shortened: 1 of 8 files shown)


ksp_plugin/interface_monitor.cpp
```diff
@@ -0,0 +1,69 @@
...
+using quantities::Infinity;
+using quantities::Time;
+using quantities::si::Nano;
+using quantities::si::Second;
+
+namespace {
+std::chrono::steady_clock::time_point monitor_start_time;
+bool is_monitor_running = false;
+
+void StartMonitoring() {
```
pleroy: Now that you have specialized interface function I would remove `StartMonitoring` and `StopMonitoring` and move the code to `...MonitorStart` and `...MonitorStop`.

ksp_plugin/interface_monitor.cpp
```diff
@@ -17,15 +17,20 @@ using quantities::si::Second;
 namespace {
 std::chrono::steady_clock::time_point monitor_start_time;
 bool is_monitor_running = false;
+}  // namespace
+
+
```
pleroy: Extra empty line.

eggrobin: Done.

Issue #1552, opened 2017-09-05
Off-by-one-frame when going on EVA


eggrobin:
The Kerbal and the vessel from which it exits are teleported one frame along their orbit in the barycentric reference frame, see screenshot.
In some cases (low orbit) this will result in a null reference exception after `WaitForFixedUpdate`, which we treat as fatal (probably because the Kerbal and vessel end up being inside the atmosphere or something like that).
This issue is not present in Чебышёв.

![screenshot105](https://user-images.githubusercontent.com/2284290/30081195-85fbb210-9286-11e7-83d7-81f6395d510d.png)

eggrobin:
This also occurs when decoupling, as well as when exiting the atmosphere (becoming manageable). I did not notice this when launching the vessels used to measure the performance increase from parallelism, so this is probably *not* due to the psychohistory changes, and it is probably related to the introduction of parallelism instead (I had not done any launches, dockings, nor EVAs  since that was added, simply timewarped vessels to measure performance).

eggrobin:
The issue persists even when using only one thread in the thread pool and waiting for the futures before calling SetBodyFrames, so it does not appear to be a thread safety issue.

eggrobin:
Confirming that the issue is not present at 15afe4b67070379e0f54e27c26aed5ad8b295f1b.

eggrobin:
The issue is not present at 9ad8f0c1297d7ad7181913217b6543415dde10c2.

eggrobin:
The issue is not present at 552943bca1186c50905541718f5f4fc6bdfc8198.

eggrobin:
The issue is not present at  4292a169c1e29eef76cd2e5f62608f81e2e2c952.

eggrobin:
The issue is not present at 286cf4ef3dc637fd708a7a50eda43a6aac434d98.

eggrobin:
The issue is not present at 5bddb355403cd2a3b4892ffbb6d65ceb55520870 either; however, by destructively colliding vessels, I am able to cause a `CHECK` failure (I do not know whether this issue is present at the commits above, or even in Чебышёв).
```
0906 22:53:57.509896 15076 plugin.cpp:364] Inserted loaded vessel Decoupling tester Debris (b465c39d-8fee-4291-b999-66c2e86591ca)
[...]

PR #1554, opened 2017-09-09
Remove an unused member variable and use a nicer output path

eggrobin:
See the last comment of #1552.


ksp_plugin_adapter/ksp_plugin_adapter.cs
```diff
@@ -26,7 +26,6 @@ public partial class PrincipiaPluginAdapter
       "principia_initial_state";
   private const String principia_gravity_model_config_name =
       "principia_gravity_model";
-  private const double Δt = 10;
```
pleroy: In light of the "Δt trainwreck" I am going to ask that all the member variables at lines 19-28 be renamed to have a trailing underscore.

eggrobin: Done.

PR #1557, opened 2017-09-16
Changes for building on Linux

pleroy:


ksp_plugin/pile_up.cpp
```diff
@@ -387,7 +387,7 @@ void PileUp::AppendToPart(DiscreteTrajectory<Barycentric>::Iterator it) const {
       pile_up_dof.velocity());
   auto const pile_up_to_barycentric = barycentric_to_pile_up.Inverse();
   for (not_null<Part*> const part : parts_) {
-    (part->*append_to_part_trajectory)(
+    (static_cast<Part*>(part)->*append_to_part_trajectory)(
```
eggrobin: Open an issue to add an operator `->*` to `not_null`.

pleroy: Let's not do that, the standard smart pointers don't have it.

Issue #1561, opened 2017-09-16
Crash when time accelerating


warmist:
Sometimes ksp crashes when time accelerating. In this instance i've captured an asteroid and was starting mining. Full time accelerate crashed ksp. 
Logs: https://gist.github.com/warmist/2b6ad7c1a4925241b318a0e79841aa5e

pleroy:
Decoded stack trace:


[`physics/discrete_trajectory_body.hpp:322`](https://github.com/mockingbirdnest/Principia/blob/c9d0610117bbfaa4633fa3f2d5b72b99547f7151/physics/discrete_trajectory_body.hpp#L322)
[`physics/discrete_trajectory_body.hpp:209`](https://github.com/mockingbirdnest/Principia/blob/c9d0610117bbfaa4633fa3f2d5b72b99547f7151/physics/discrete_trajectory_body.hpp#L209)
[`physics/body_centred_body_direction_dynamic_frame_body.hpp:99`](https://github.com/mockingbirdnest/Principia/blob/c9d0610117bbfaa4633fa3f2d5b72b99547f7151/physics/body_centred_body_direction_dynamic_frame_body.hpp#L99)
[`ksp_plugin/renderer.cpp:161`](https://github.com/mockingbirdnest/Principia/blob/c9d0610117bbfaa4633fa3f2d5b72b99547f7151/ksp_plugin/renderer.cpp#L161)
[`ksp_plugin/renderer.cpp:268`](https://github.com/mockingbirdnest/Principia/blob/c9d0610117bbfaa4633fa3f2d5b72b99547f7151/ksp_plugin/renderer.cpp#L268)
[`ksp_plugin/plugin.cpp:905`](https://github.com/mockingbirdnest/Principia/blob/c9d0610117bbfaa4633fa3f2d5b72b99547f7151/ksp_plugin/plugin.cpp#L905)
[`ksp_plugin/interface.cpp:679`](https://github.com/mockingbirdnest/Principia/blob/c9d0610117bbfaa4633fa3f2d5b72b99547f7151/ksp_plugin/interface.cpp#L679)

warmist:
Also not sure if it's relevant but the game kept lagging (i.e. stopping for half a second) every few minutes.

pleroy:
The periodic slowdown is probably due to autosaving: the Principia data occupies about 40 MiB, which is going to take a little while to write to the file.

[...]

PR #1562, opened 2017-09-17
Move the code that extends the prediction to Renderer

pleroy:
This was noticed when investigating #1561 (the new test show a case where we evaluate the prediction after its last point) but this is really unrelated because #1561 is about evaluating the prediction before its first point.

Also making `Vessel::UpdatePrediction` idempotent for performance.


eggrobin:
The number of steps of the target prediction does not appear to be set correctly when changing that of the active vessel.

PR #1576, opened 2017-09-24
Remove std::move() from return statements

aw1621107:


pleroy:

pleroy:
ok to test

pleroy:
ok to test

PR #1580, opened 2017-09-24
Turn loop variable copies to references

aw1621107:
I was less certain that these changes were right, so each file changed got its own commit to help with future bisecting.


pleroy:

eggrobin:
As discussed on IRC, we don't use `not_null<T*> const&`, on the other hand we try to make those explicit rather than `auto`, precisely so we aren't tempted to add a `&`, so it might be a good idea to change those that use auto, for consistency (e.g., PileUp has both `auto const` and `not_null<Part*> const` for iterating on `parts_`.

pleroy:
ok to test

pleroy:
ok to test

pleroy:
Thanks for all your contributions.  Much appreciated.

PR #1584, opened 2017-10-02
Correctly handle the Krakensbane and floating origin, unpacked part insertion

eggrobin:
3rd point of the 10-point plan.

### DIFF
--- a/ksp_plugin/plugin.cpp
+++ b/ksp_plugin/plugin.cpp
@@ -418,9 +418,6 @@ void Plugin::InsertOrKeepLoadedPart(
     }
   } else {
     Instant const previous_time = current_time_ - Δt;
-    enum class LocalTag { tag };
-    using MainBodyCentred =
-        geometry::Frame<LocalTag, LocalTag::tag, /*frame_is_inertial=*/false>;
     BodyCentredNonRotatingDynamicFrame<Barycentric, MainBodyCentred> const
         main_body_frame{ephemeris_.get(),
                         FindOrDie(celestials_, main_body_index)->body()};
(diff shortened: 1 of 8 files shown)


ksp_plugin/plugin.cpp
```diff
@@ -418,9 +418,8 @@ void Plugin::InsertOrKeepLoadedPart(
     }
   } else {
     Instant const previous_time = current_time_ - Δt;
-    enum class LocalTag { tag };
-    using MainBodyCentred =
-        geometry::Frame<LocalTag, LocalTag::tag, /*frame_is_inertial=*/false>;
+    auto const Δplanetarium_rotation =
```
pleroy: I'd be happier to have a type here.

ksp_plugin/plugin.cpp
```diff
@@ -418,9 +418,8 @@ void Plugin::InsertOrKeepLoadedPart(
...
+        Exp(Δt * angular_velocity_of_world_).Forget();
     BodyCentredNonRotatingDynamicFrame<Barycentric, MainBodyCentred> const
```
pleroy: I wonder if this computation could be factored with `BarycentricToWorld`.  Put a TODO to investigate.

ksp_plugin/plugin.cpp
eggrobin: Done.

ksp_plugin/plugin.cpp
eggrobin: Done.

eggrobin: Added an enum.

PR #1585, opened 2017-10-04
Ignore vessels that touch the ground or might fall underground, remove altitude thresholds

eggrobin:
Point 5 of the 10-point plan, tested as part of point 6.

### DIFF
--- a/ksp_plugin_adapter/ksp_plugin_adapter.cs
+++ b/ksp_plugin_adapter/ksp_plugin_adapter.cs
@@ -370,8 +370,6 @@ public partial class PrincipiaPluginAdapter
     return UnmanageabilityReasons(vessel) == null;
   }
 
-  const double manageability_altitude_threshold = 300 /*m*/;
-
   private string UnmanageabilityReasons(Vessel vessel) {
     List<string> reasons = new List<string>(capacity : 3);
     if (vessel.state == Vessel.State.DEAD) {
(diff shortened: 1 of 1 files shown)


ksp_plugin_adapter/ksp_plugin_adapter.cs
```diff
@@ -380,21 +378,19 @@ public partial class PrincipiaPluginAdapter
     if (!(vessel.situation == Vessel.Situations.SUB_ORBITAL ||
           vessel.situation == Vessel.Situations.ORBITING ||
           vessel.situation == Vessel.Situations.ESCAPING ||
-          (vessel.situation == Vessel.Situations.FLYING &&
-           (vessel.altitude > manageability_altitude_threshold ||
-            vessel.altitude == -vessel.mainBody.Radius)))) {
```
pleroy: Shouldn't we keep the `-Radius` condition?  Do we want to manage the vessel at the centre of the celestial.

ksp_plugin_adapter/ksp_plugin_adapter.cs
```diff
@@ -415,6 +411,27 @@ public partial class PrincipiaPluginAdapter
...
+  }
+
+  private bool TouchesTheGround(Vessel vessel) {
+    return vessel.parts
+        .Where(part =>
+               part.Modules.OfType<ModuleWheelBase>()
+                   .Where(wheel => wheel.isGrounded)
+                   .Any() ||
+               part.currentCollisions
+                   .Where(collider => collider.gameObject.layer == 15)
```
pleroy: Could we have a named constant for `15`?

eggrobin: Added an enum.

ksp_plugin_adapter/ksp_plugin_adapter.cs
eggrobin: As discussed, this is fine (🔥🐶☕🔥).

PR #1592, opened 2017-10-12
Do not use the prediction prior to its fork point

pleroy:


eggrobin:
Seems to work in-game.
However, while the history will not be plotted in the target LVLH frame, this will plot the parts of the flight plan that are in the past. I don't think we actually want that, but I'm ok with punting on it for now; in that case open an issue to track it.

PR #1600, opened 2017-10-13
Fix compilation errors emitted by Clang 5.0

aw1621107:
Errors emitted were of the form

```
    geometry/identity_test.cpp:77:19: error: qualified reference to 'Identity' is a
          constructor name rather than a type in this context
      EXPECT_THAT(Id::Identity().Inverse()(vector2).coordinates(),
                      ^
```

To be honest, not sure since this is an actual problem rather than a compiler bug, since the code appears to have always been that way...

The way I fixed the compilation error was chosen because I noticed that the code under `TEST_F(IdentityTest, Determinant)` had a "standalone" (for lack of a better phrase) `Id` object, unlike the code causing the errors.

### DIFF
--- a/geometry/identity_test.cpp
+++ b/geometry/identity_test.cpp
@@ -52,17 +52,20 @@ TEST_F(IdentityTest, Determinant) {
 }
 
 TEST_F(IdentityTest, AppliedToVector) {
-  EXPECT_THAT(Id::Identity()(vector_).coordinates(),
+  Id identity;
+  EXPECT_THAT(identity(vector_).coordinates(),
               Eq<R3>({1.0 * Metre, 2.0 * Metre, 3.0 * Metre}));
 }
 
 TEST_F(IdentityTest, AppliedToBivector) {
-  EXPECT_THAT(Id::Identity()(bivector_).coordinates(),
+  Id identity;
+  EXPECT_THAT(identity(bivector_).coordinates(),
               Eq<R3>({1.0 * Metre, 2.0 * Metre, 3.0 * Metre}));
 }
 
 TEST_F(IdentityTest, AppliedToTrivector) {
-  EXPECT_THAT(Id::Identity()(trivector_).coordinates(),
+  Id identity;
+  EXPECT_THAT(identity(trivector_).coordinates(),
               Eq(4.0 * Metre));
 }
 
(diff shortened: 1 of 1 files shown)


geometry/identity_test.cpp
```diff
@@ -71,7 +74,8 @@ TEST_F(IdentityTest, Inverse) {
   Vector<quantities::Length, World2> const vector2 =
       Vector<quantities::Length, World2>(
           R3(1.0 * Metre, 2.0 * Metre, 3.0 * Metre));
-  EXPECT_THAT(Id::Identity().Inverse()(vector2).coordinates(),
+  Id identity;
```
aw1621107: Not sure whether another `Id` object is needed with one being created a few lines below, but figured I should stay on the conservative side, since I'm not familiar with the code.

pleroy:

pleroy:
I think that adding all these intermediary objects obscures the code somewhat.  According to [this](http://lists.llvm.org/pipermail/llvm-bugs/2017-January/053481.html) bug report the fix is to remove the part after the `::`.  Thus `Id::Identity()` should become `Id()`.  Could you try that?

aw1621107:
Yep, that seems to work. I'll get a new commit pushed soon.

Should I also remove the `Id identity;` line in `TEST_F(IdentityTest, Determinant)` and `Id id;` in `TEST_F(IdentityTest, Inverse)`?

And sort of an aside, does `Id()` really count as a function-style cast, as implied by the GCC error message?

pleroy:
I'd remove `Id identity;` in `TEST_F(IdentityTest, Determinant)` as it seems to serve no purpose.  I'd keep `Id id;` in `TEST_F(IdentityTest, Inverse)` as it's used in several places so it (marginally) simplifies the code.

I must confess that I don't understand the part about a "function-style cast".  What's happening is clear, though: `Id` is a type, `Id()` is the construction of an object of that type, `Id::Identity()` is an attempt at calling the constructor as if it was a function, but a constructor is not a function, so you lose.  (`Id::Identity` may be used to designate the contructor in other contexts, e.g., in a `using` clause.)

aw1621107:
Done! And thanks for clarifying what's going on that requires this change. Still learning the darker corners of C++...

pleroy:
ok to test

pleroy:
Thanks for your contribution.

PR #1605, opened 2017-10-14
only support macosx >= 10.11

lamont-granquist:
only good build box is still 10.11 so this has to be the floor for now.


pleroy:

pleroy:
ok to test

PR #1606, opened 2017-10-14
Don't lose accuracy when computing the times in the integrators

### LINKED ISSUE #1595: Crash on camera switching from Sun to vessel with extremely large velocity/altitude
From #principia:
```
04:08 < awang> egg|work|egg: I have an interesting bug for you
04:09 < awang> Looking in the tracking station, see a sounding rocket core some absurd
               distance from the Earth
04:10 < awang> 242.189Tm, moving at 596.5Mm/s
04:10 < awang> Click "fly", go to map view, Sun isn't visible
04:10 < awang> Hit tab, camera goes to sun
04:10 < awang> Hit backquote to reset camera, crash in Principia
04:11 < awang> https://hastebin.com/raw/qelasunita
04:13 < awang> INFO file is here:
               https://www.dropbox.com/s/c8b2mfd2skq9iyp/INFO.20171011-180038.310?dl=0
04:13 < awang> Don't know if this is something to worry about, since the circumstances
               are probably somewhat uncommon
```

Save is [here](https://www.dropbox.com/s/61pii4pjlph835c/persistent.sfs?dl=0). Don't know if it'll work, since my game is heavily modded. The ship in question should be "Proc Sounding Rocket 1"

pleroy:
Fix #1595

### DIFF
--- a/physics/ephemeris_test.cpp
+++ b/physics/ephemeris_test.cpp
@@ -498,7 +498,7 @@ TEST_P(EphemerisTest, EarthProbe) {
   // The solution is a line, so the rounding errors dominate.  Different
   // compilers result in different errors and thus different numbers of steps.
   EXPECT_THAT(probe_positions.size(),
-              AnyOf(Eq(434), Eq(419), Eq(400), Eq(379), Eq(386), Eq(-1)));
+              AnyOf(Eq(400), Eq(386), Eq(379)));
   EXPECT_THAT(probe_positions.back().coordinates().x,
               AlmostEquals(1.00 * period * v_probe, 194, 235));
   EXPECT_THAT(probe_positions.back().coordinates().y,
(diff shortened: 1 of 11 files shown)


physics/ephemeris_test.cpp
```diff
@@ -497,9 +497,10 @@ TEST_P(EphemerisTest, EarthProbe) {
   }
   // The solution is a line, so the rounding errors dominate.  Different
   // compilers result in different errors and thus different numbers of steps.
-  EXPECT_THAT(probe_positions.size(), AnyOf(Eq(434), Eq(419), Eq(-1)));
+  EXPECT_THAT(probe_positions.size(),
+              AnyOf(Eq(434), Eq(419), Eq(400), Eq(379), Eq(-1)));
```
eggrobin: Does this pass on linux? Also perhaps you should remove the old expectations.
What's up with the Eq(-1)? shouldn1t vector::size be nonnegative?

pleroy:
retest this please

PR #1610, opened 2017-10-16
Indicate burn type being used by flight planner

aw1621107:
Burn type being used could have gone into
Compute[Engine|RCS]Characteristics() and/or UseTheForceLuke(), but that
wouldn't work later if/when the ability for the user to select specific
engines for a burn is added, since ComputeEngineCharacteristics would be
used when "normal" engines are used for the burn, regardless of whether
the engines used for the burn were selected manually by the user or by
the "Active Engines" button.

### DIFF
--- a/ksp_plugin_adapter/burn_editor.cs
+++ b/ksp_plugin_adapter/burn_editor.cs
@@ -55,6 +55,12 @@ class BurnEditor {
     ComputeEngineCharacteristics();
   }
 
+  enum BurnType {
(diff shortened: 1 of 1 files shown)


eggrobin:
As discussed on IRC, this information is not persisted (the principia.serialization.Manoeuvre message would have to include it, as well as the interchange principia.journal.serialization.Burn).

Closing for now; @aw1621107, feel free to reopen later if you want to do the persistence.

PR #1611, opened 2017-10-17
Ensure flight planner is always (partially) visible

### LINKED ISSUE #1609: Flight plan window can become inaccessible after deleting last maneuver
If deleting the last maneuver in the flight planner causes the bottom of the flight planning window to vanish off the top of the screen, the window is no longer accessible, since there is no way for the user to drag it back on-screen.

Going to the space center and returning appears to reset the flight planning window location, but that could be somewhat annoying to do if loading times are really long.

aw1621107:
Fixes #1609.

### DIFF
--- a/ksp_plugin_adapter/flight_planner.cs
+++ b/ksp_plugin_adapter/flight_planner.cs
@@ -39,6 +39,14 @@ class FlightPlanner : WindowRenderer {
                               func       : RenderPlanner,
                               text       : "Flight plan",
                               options    : UnityEngine.GUILayout.MinWidth(500));
+      const float min_width_on_screen = 50;
+      const float min_height_on_screen = 50;
+      window_rectangle_.x = UnityEngine.Mathf.Clamp(window_rectangle_.x,
+                                                    min_width_on_screen - window_rectangle_.width,
+                                                    UnityEngine.Screen.width - min_width_on_screen);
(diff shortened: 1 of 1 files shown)


ksp_plugin_adapter/flight_planner.cs
```diff
@@ -39,6 +39,14 @@ class FlightPlanner : WindowRenderer {
                               func       : RenderPlanner,
                               text       : "Flight plan",
                               options    : UnityEngine.GUILayout.MinWidth(500));
+      const float min_width_on_screen = 50;
+      const float min_height_on_screen = 50;
+      window_rectangle_.x = UnityEngine.Mathf.Clamp(window_rectangle_.x,
+                                                    min_width_on_screen - window_rectangle_.width,
```
eggrobin: We use 80 character lines, please fold.
I'm also going ask you to put the all the `min_something` on the same side, I find that easier to understand:
```
      window_rectangle_.x =
          UnityEngine.Mathf.Clamp(
              window_rectangle_.x,
              -window_rectangle_.width + min_width_on_screen,
              UnityEngine.Screen.width - min_width_on_screen);
      window_rectangle_.y =
          UnityEngine.Mathf.Clamp(
              window_rectangle_.y,
              -window_rectangle_.height + min_height_on_screen,
              UnityEngine.Screen.height - min_height_on_screen);
```
should be fine.

PR #1615, opened 2017-10-28
Retrobop should not use retro APIs

eggrobin:
Remove old protos that are missing required fields, use the new unified KSP system proto and stabilization function.


mathematica/retrobop_dynamical_stability.cpp
```diff
@@ -159,22 +158,15 @@ std::unique_ptr<Message> Read(std::ifstream& file) {
...
-    auto const degrees_of_freedom = Read<serialization::Pair>(input_file);
-    CHECK(degrees_of_freedom != nullptr);
-    system.bodies.push_back(MassiveBody::ReadFromMessage(*body));
-    system.degrees_of_freedom.push_back(
-        DegreesOfFreedom<Barycentric>::ReadFromMessage(*degrees_of_freedom));
-  }
-  input_file.close();
-  return system;
+HierarchicalSystem<Barycentric>::BarycentricSystem MakeStabilizedKSPSystem() {
+  static auto const& system = *[]() {
```
pleroy: The `&` worries me here.  Is the lifetime extended until the termination of the program?

eggrobin: `new` so this lives forever, no lifetime extension going on here.
Since I'm using a lambda anyway, perhaps it would be clearer to `make_unique` and `release`.

Issue #228, opened 2014-10-25
Keep the whole history


eggrobin:
For the moment we compute the histories using `sampling_period == 0`, which results in large gaps in the trajectory at high timewarp (rendered as long straight line segments, see below).  We should use `sampling_period == 1` instead, and downsample ourselves as needed when rendering.
![screenshot8](https://cloud.githubusercontent.com/assets/2284290/4780431/6b346bfe-5c68-11e4-9093-9b9af84bf963.png)

pleroy:
Blocked by #973.

eggrobin:
@aw1621107  reports impact; the visual appearance has changed with the Чебышёв interpolated plotting, we now get weird loops instead of straight segments:
![](https://i.imgur.com/Uanelz6.jpg)
![](https://i.imgur.com/bqkqtNz.jpg)

PR #1617, opened 2017-10-31
The downsampling algorithm

eggrobin:


base/ranges.hpp
```diff
@@ -0,0 +1,54 @@
...
+
+  IterableRange(Iterator begin, Iterator end);
+
+  const_iterator begin() const;
+  const_iterator end() const;
+
+  const_iterator cbegin() const;
+  const_iterator cend() const;
+
+  size_type size() const;
```
pleroy: Comment that this is O(N).

numerics/fit_hermite_spline.hpp
```diff
@@ -0,0 +1,41 @@
...
+namespace internal_fit_hermite_spline {
+
+// Given |samples| for which the arguments, values, and derivatives can be
+// obtained via the given functors, returns a sequence of iterators
+// (it₀, it₁, it₂, ..., itᵣ) such that, with it₋₁ = samples.begin(),
+// for all integers it in [0, r], the |Hermite3| interpolation of
+// (*itᵢ₋₁, *itᵢ) fits |samples| within |tolerance|, but the interpolation
+// of (*itᵢ₋₁, *(it + 1)ᵢ) does not, i.e., the iterators delimit "maximal"
+// fitting polynomials.
+// Note that it follows from the above that itᵣ < samples.end() - 1, so that at
```
pleroy: This sentence doesn't really parse.

numerics/fit_hermite_spline.hpp
```diff
@@ -0,0 +1,41 @@
...
+#include <list>
+
+#include "numerics/hermite3.hpp"
+
+namespace principia {
+namespace numerics {
+namespace internal_fit_hermite_spline {
+
+// Given |samples| for which the arguments, values, and derivatives can be
+// obtained via the given functors, returns a sequence of iterators
```
pleroy: You need to document the functors better, they are totally mysterious in the template declarations below.

In fact it would be much more readable to do something like, e.g.:
```
  typename Argument,
  typename Value,

  std::function<Argument(typename Samples::iterator)> const& get_argument,
  ... // Same for get_value.
  std::function<Derivative<Value>(typename Sample::iterator)> const& get_derivative,
```
As written, all the contract is hidden in the implementation.

numerics/fit_hermite_spline_body.hpp
```diff
@@ -0,0 +1,96 @@
...
+      std::is_same<ErrorType, typename Normed<Value>::NormType>::value,
+      "|tolerance| must have the same type as a distance between values");
+
+  if (samples.size() < 3) {
+    // With 0 or 1 points there is nothing to interpolate, with 2 we cannot
+    // estimate the error.
+    return {};
+  }
+
+  auto interpolation_error = [get_argument, get_value, get_derivative](
```
pleroy: Alphabetize the captures.

numerics/fit_hermite_spline_body.hpp
```diff
@@ -0,0 +1,96 @@
...
+    // such the cubic fitting one more sample would not fit the samples within
+    // |tolerance|.
+    // Note that there may be more than one cubic satisfying this property;
+    // ideally we would like to find the longest one, but this would be costly,
+    // and we do not expect significant gains from this in practice.
+
+    // Invariant: The Hermite interpolant on [samples.begin(), lower] is below
+    // the tolerance, the Hermite interpolant on [samples.begin(), upper] is
+    // above.
+    auto lower = samples.begin() + 1;
```
pleroy: Could this be implemented using `std::binary_search`?

numerics/fit_hermite_spline_test.cpp
```diff
@@ -0,0 +1,110 @@
...
+    Instant t;
+    Length x;
+    Speed v;
+  };
+  Instant const t0_;
+};
+
+TEST_F(FitHermiteSplineTest, Sinusoid) {
+  AngularFrequency const ω = 1 * Radian / Second;
+  auto const f = [ω, this](Instant const t) {
```
pleroy: `Instant const&` at several places.

numerics/fit_hermite_spline_test.cpp
```diff
@@ -0,0 +1,110 @@
...
+    return Cos(ω * (t - t0_)) * Metre;
+  };
+  auto const df = [ω, this](Instant const t) {
+    return -ω * Sin(ω *(t - t0_)) * Metre / Radian;
+  };
+  auto t = DoublePrecision<Instant>(t0_);
```
pleroy: Put this immediately before the loops.

numerics/fit_hermite_spline_test.cpp
```diff
@@ -0,0 +1,110 @@
...
+                       [](auto&& sample) -> auto&& { return sample.t; },
+                       [](auto&& sample) -> auto&& { return sample.x; },
+                       [](auto&& sample) -> auto&& { return sample.v; },
+                       1 * Centi(Metre));
+
+  // Note that gmock doesn't do decltypes, so we can't pass a λ directly.
+  // Also note that |Pointee| doesn't work with iterators, so
+  // |Pointee(Field(&Sample::t, _))| is not an option.
+  std::function<Instant(std::vector<Sample>::const_iterator)> const get_time =
+      [](auto it) { return it->t; };
```
pleroy: `auto const`

base/ranges.hpp
eggrobin: It is O(N), but it can be Θ(1), so this would be misleading.

numerics/fit_hermite_spline.hpp
eggrobin: Done. https://twitter.com/stephentyrone/status/902227449433513984

numerics/fit_hermite_spline.hpp
eggrobin: Done.

numerics/fit_hermite_spline_body.hpp
eggrobin: Done.

numerics/fit_hermite_spline_body.hpp
eggrobin: `std::lower_bound` and `std::binary_search` require (25.4.3.1:1, 25.4.3.4:1)
> The elements `e` of `[first,last)` shall be partitioned with respect to the expression `e <
value` or `comp(e, value)`.

and `std::upper_bound` requires (25.4.3.2:1)

> The elements `e` of `[first,last)` shall be partitioned with respect to the expression `!(value
< e)` or `!comp(value, e)`.

*partitioned with respect to an expression* is defined as (25.4:6)
> A sequence `[start,finish)` is *partitioned with respect to an expression* `f(e)` if there exists an integer `n` such that for all `0 <= i < distance(start, finish)`, `f(*(start + i))` is true if and only if `i < n`.

Another option, `std::partition_point`, requires (25.3.13:16)
> `[first,last)` shall be partitioned by `pred`, i.e. all elements that satisfy `pred` shall appear before those that do not.

This is not satisfied by the range `[samples.begin()+1, samples.end()[`, which may have several alternations of the predicate (but has at least one, and an odd number), as described in the comment.

All citations from the working draft [N4296](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2014/n4296.pdf) of JTC 1/SC 22/WG 21.

numerics/fit_hermite_spline_test.cpp
eggrobin: Done.

numerics/fit_hermite_spline_test.cpp
eggrobin: Done and scoped.

numerics/fit_hermite_spline_test.cpp
eggrobin: Done.

numerics/fit_hermite_spline_body.hpp
eggrobin: Added a proof of termination.

eggrobin:
This is part of the work on #228.

PR #1619, opened 2017-11-05
Downsampling trajectories

eggrobin:
This is part of the work on #228.

### DIFF
--- a/physics/discrete_trajectory.hpp
+++ b/physics/discrete_trajectory.hpp
@@ -194,15 +194,15 @@ class DiscreteTrajectory : public Forkable<DiscreteTrajectory<Frame>,
     Instant const& first_dense_time() const;
     // Keeps |dense_intervals_| consistent with the new
     // |start_of_dense_timeline_|.
-    void set_start_of_dense_timeline(TimelineConstIterator value,
-                                     Timeline const& timeline);
+    void SetStartOfDenseTimeline(TimelineConstIterator value,
+                                 Timeline const& timeline);
 
     // Sets |dense_intervals_| to
     // |std::distance(start_of_dense_timeline, timeline.end()) - 1|.  This is
     // linear in the value of |dense_intervals_|.
-    void recount_dense_intervals(Timeline const& timeline);
+    void RecountDenseIntervals(Timeline const& timeline);
     // Increments |dense_intervals_|.  The caller must ensure that this is
-    // equivalent to |recount_dense_intervals(timeline)|.  This is checked in
+    // equivalent to |RecountDenseIntervals(timeline)|.  This is checked in
     // debug mode.
     void increment_dense_intervals(Timeline const& timeline);
 
--- a/physics/discrete_trajectory.hpp
+++ b/physics/discrete_trajectory.hpp
@@ -182,20 +182,56 @@ class DiscreteTrajectory : public Forkable<DiscreteTrajectory<Frame>,
   std::int64_t timeline_size() const override;
 
  private:
-  struct Downsampling {
+  class Downsampling {
+   public:
+    Downsampling(std::int64_t max_dense_intervals,
+                 Length tolerance,
+                 TimelineConstIterator start_of_dense_timeline,
+                 Timeline const& timeline);
+
+    TimelineConstIterator start_of_dense_timeline() const;
+    // |start_of_dense_timeline()->first|, for readability.
+    Instant const& first_dense_time() const;
+    // Keeps |dense_intervals_| consistent with the new
+    // |start_of_dense_timeline_|.
+    void set_start_of_dense_timeline(TimelineConstIterator value,
+                                     Timeline const& timeline);
+
+    // Sets |dense_intervals_| to
+    // |std::distance(start_of_dense_timeline, timeline.end()) - 1|.  This is
+    // linear in the value of |dense_intervals_|.
+    void recount_dense_intervals(Timeline const& timeline);
+    // Increments |dense_intervals_|.  The caller must ensure that this is
+    // equivalent to |recount_dense_intervals(timeline)|.  This is checked in
+    // debug mode.
+    void increment_dense_intervals(Timeline const& timeline);
+
+    std::int64_t max_dense_intervals() const;
+    bool reached_max_dense_intervals() const;
+
+    Length tolerance() const;
+
+    void WriteToMessage(
+        not_null<serialization::DiscreteTrajectory::Downsampling*> message,
+        Timeline const& timeline) const;
+    static Downsampling ReadFromMessage(
+        serialization::DiscreteTrajectory::Downsampling const& message,
+        Timeline const& timeline);
+
+   private:
+    // The maximal value that |dense_intervals| is allowed to reach before
+    // downsampling occurs.
+    std::int64_t const max_dense_intervals_;
+    // The tolerance for downsampling with |FitHermiteSpline|.
+    Length const tolerance_;
     // An iterator to the first point of the timeline which is not the left
     // endpoint of a downsampled interval.  Not |timeline_.end()| if the
     // timeline is nonempty.
-    TimelineConstIterator start_of_dense_timeline;
+    TimelineConstIterator start_of_dense_timeline_;
     // |std::distance(start_of_dense_timeline, timeline_.cend()) - 1|.  Kept as
     // an optimization for |Append| as it can be maintained by incrementing,
     // whereas |std::distance| is linear in the value of the result.
-    std::int64_t dense_intervals;
-    // The maximal value that |dense_intervals| is allowed to reach before
-    // downsampling occurs.
-    std::int64_t max_dense_intervals;
-    // The tolerance for downsampling with |FitHermiteSpline|.
-    Length tolerance;
+    std::int64_t dense_intervals_;
   };
 
   // This trajectory need not be a root.
--- a/physics/discrete_trajectory.hpp
+++ b/physics/discrete_trajectory.hpp
@@ -175,9 +175,18 @@ class DiscreteTrajectory : public Forkable<DiscreteTrajectory<Frame>,
 
  private:
   struct Downsampling {
+    // An iterator to the first point of the timeline which is not the left
+    // endpoint of a downsampled interval.  Not |timeline_.end()| if the
+    // timeline is nonempty.
     TimelineConstIterator start_of_dense_timeline;
+    // |std::distance(start_of_dense_timeline, timeline_.cend()) - 1|.  Kept as
+    // an optimization for |Append| as it can be maintained by incrementing,
+    // whereas |std::distance| is linear in the value of the result.
     std::int64_t dense_intervals;
+    // The maximal value that |dense_intervals| is allowed to reach before
+    // downsampling occurs.
     std::int64_t max_dense_intervals;
+    // The tolerance for downsampling with |FitHermiteSpline|.
     Length tolerance;
   };
 
--- a/physics/discrete_trajectory.hpp
+++ b/physics/discrete_trajectory.hpp
@@ -174,6 +174,13 @@ class DiscreteTrajectory : public Forkable<DiscreteTrajectory<Frame>,
   std::int64_t timeline_size() const override;
 
  private:
+  struct Downsampling {
+    TimelineConstIterator start_of_dense_timeline;
+    std::int64_t dense_intervals;
+    std::int64_t max_dense_intervals;
+    Length tolerance;
+  };
+
   // This trajectory need not be a root.
   void WriteSubTreeToMessage(
       not_null<serialization::DiscreteTrajectory*> message,
@@ -191,6 +198,8 @@ class DiscreteTrajectory : public Forkable<DiscreteTrajectory<Frame>,
       Instant const& time) const;
 
   Timeline timeline_;
+
+  std::experimental::optional<Downsampling> downsampling_;
 
   template<typename, typename>
   friend class internal_forkable::ForkableIterator;
--- a/physics/discrete_trajectory.hpp
+++ b/physics/discrete_trajectory.hpp
@@ -198,7 +198,7 @@ class DiscreteTrajectory : public Forkable<DiscreteTrajectory<Frame>,
                                  Timeline const& timeline);
 
     // Sets |dense_intervals_| to
-    // |std::distance(start_of_dense_timeline, timeline.end()) - 1|.  This is
+    // |std::distance(start_of_dense_timeline_, timeline.end()) - 1|.  This is
     // linear in the value of |dense_intervals_|.
     void RecountDenseIntervals(Timeline const& timeline);
     // Increments |dense_intervals_|.  The caller must ensure that this is
--- a/physics/discrete_trajectory.hpp
+++ b/physics/discrete_trajectory.hpp
@@ -131,6 +131,14 @@ class DiscreteTrajectory : public Forkable<DiscreteTrajectory<Frame>,
   // |time|.  This trajectory must be a root.
   void ForgetBefore(Instant const& time);
 
+  // This trajectory must be root, and must not be already downsampling.
+  // Following this call, this trajectory must not have forks when calling
+  // |Append|.  Occasionally removes intermediate points from the trajectory
+  // when |Append|ing, ensuring that |EvaluatePosition| returns a result within
+  // |tolerance| of the missing points.  |max_dense_intervals| is the largest
+  // number of points that can be added before removal is considered.
+  void SetDownsampling(std::int64_t max_dense_intervals, Length tolerance);
+
   // Implementation of the interface |Trajectory|.
 
   // The bounds are the times of |Begin()| and |last()| if this trajectory is
--- a/physics/discrete_trajectory_body.hpp
+++ b/physics/discrete_trajectory_body.hpp
@@ -162,16 +162,15 @@ void DiscreteTrajectory<Frame>::Append(
       << (--timeline_.end())->first;
   if (downsampling_.has_value()) {
     if (timeline_.size() == 1) {
-      downsampling_->start_of_dense_timeline = timeline_.begin();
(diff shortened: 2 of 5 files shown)


physics/discrete_trajectory.hpp
```diff
@@ -131,6 +131,14 @@ class DiscreteTrajectory : public Forkable<DiscreteTrajectory<Frame>,
   // |time|.  This trajectory must be a root.
   void ForgetBefore(Instant const& time);
 
+  // This trajectory must be root, and must not be already downsampling.
+  // Following this call, this trajectory must not have forks when calling
+  // |Append|.  Occasionally removes intermediate points from the trajectory
+  // when |Append|ing, ensuring that |EvaluatePosition| returns a result within
+  // |tolerance| of the missing points.  |max_dense_intervals| is the largest
+  // number of points that can be added before removal is considered.
```
pleroy: Is it the number of *points* or the number of *intervals*?

physics/discrete_trajectory.hpp
```diff
@@ -174,6 +182,22 @@ class DiscreteTrajectory : public Forkable<DiscreteTrajectory<Frame>,
   std::int64_t timeline_size() const override;
 
  private:
+  struct Downsampling {
```
pleroy: I'd like to have a 2-argument constructor taking `max_dense_intervals` and `tolerance`.  Then the last two members may be made `const`.

In fact this should be a bona-fide class with actual code, see my comments below.

physics/discrete_trajectory_body.hpp
```diff
@@ -158,6 +160,44 @@ void DiscreteTrajectory<Frame>::Append(
...
       << (--timeline_.end())->first;
+  if (downsampling_.has_value()) {
+    if (timeline_.size() == 1) {
+      downsampling_->start_of_dense_timeline = timeline_.begin();
+      downsampling_->dense_intervals = 0;
+    } else {
+      CheckNoForksBefore(last().time());
+      ++downsampling_->dense_intervals;
+      if (downsampling_->dense_intervals ==
+          downsampling_->max_dense_intervals) {
```
pleroy: Would this fit on the previous line?

physics/discrete_trajectory_body.hpp
```diff
@@ -158,6 +160,44 @@ void DiscreteTrajectory<Frame>::Append(
...
+        std::vector<TimelineConstIterator> dense_iterators;
+        dense_iterators.reserve(downsampling_->max_dense_intervals + 1);
```
pleroy: Comment the `+ 1`: this is a number of points.

physics/discrete_trajectory_body.hpp
```diff
@@ -158,6 +160,44 @@ void DiscreteTrajectory<Frame>::Append(
...
+            dense_iterators,
+            [](auto&& it) -> auto&& { return it->first; },
+            [](auto&& it) -> auto&& { return it->second.position(); },
+            [](auto&& it) -> auto&& { return it->second.velocity(); },
+            downsampling_->tolerance);
+        if (right_endpoints.empty()) {
+          right_endpoints.push_back(dense_iterators.end() - 1);
+        }
+        TimelineConstIterator left = downsampling_->start_of_dense_timeline;
+        for (const auto& it_in_dense_iterators : right_endpoints) {
```
pleroy: I find it odd to call `it_in_dense_iterators` an iterator which is not in `dense_iterators`.

physics/discrete_trajectory_body.hpp
```diff
@@ -158,6 +160,44 @@ void DiscreteTrajectory<Frame>::Append(
...
+          TimelineConstIterator const right = *it_in_dense_iterators;
+          timeline_.erase(++left, right);
+          left = right;
+        }
+        downsampling_->start_of_dense_timeline = left;
+        downsampling_->dense_intervals =
+            std::distance(downsampling_->start_of_dense_timeline,
```
pleroy: You could avoid an O(N) call to `std::distance` by decrementing `dense_intervals` as you go in the "decimation" loop above.

physics/discrete_trajectory_body.hpp
```diff
@@ -168,7 +208,26 @@ void DiscreteTrajectory<Frame>::ForgetAfter(Instant const& time) {
   // entry and all the entries that follow it.  This preserves any entry with
   // time == |time|.
   auto const it = timeline_.upper_bound(time);
+  if (downsampling_.has_value()) {
+    if (it != timeline_.end() &&
+        downsampling_->start_of_dense_timeline->first >= it->first) {
```
pleroy: This seems backwards, I think you want `<=`.  You are going to shorten the dense timeline so you need to adjust the number of dense intervals.  In the case covered by your predicate the dense timeline goes away entirely so it should be reset to empty.

If my analysis is correct, a test is missing.

Also, could we give name to the `first`?

physics/discrete_trajectory_body.hpp
```diff
@@ -168,7 +208,26 @@ void DiscreteTrajectory<Frame>::ForgetAfter(Instant const& time) {
...
+      // The start of the dense timeline will be invalidated.
+      if (it == timeline_.begin()) {
+        // The timeline will be empty after erasing.
+        downsampling_->start_of_dense_timeline = timeline_.end();
```
pleroy: Are we sure that `end` is not invalidated?

physics/discrete_trajectory_body.hpp
```diff
@@ -178,10 +237,34 @@ void DiscreteTrajectory<Frame>::ForgetBefore(Instant const& time) {
...
   // the entries that precede it.  This preserves any entry with time == |time|.
   auto it = timeline_.lower_bound(time);
+  if (downsampling_.has_value() &&
+      (it == timeline_.end() ||
+       downsampling_->start_of_dense_timeline->first < it->first)) {
+    // The start of the dense timeline will be invalidated.
+    downsampling_->start_of_dense_timeline = it;
+    downsampling_->dense_intervals =
+        std::distance(downsampling_->start_of_dense_timeline,
+                      timeline_.cend()) - 1;
```
pleroy: I would factor these 4 line in a function of `Downsampling` that would probably be usable in `Append` and in other places.

physics/discrete_trajectory_body.hpp
```diff
@@ -299,6 +382,18 @@ void DiscreteTrajectory<Frame>::WriteSubTreeToMessage(
     degrees_of_freedom.WriteToMessage(
         instantaneous_degrees_of_freedom->mutable_degrees_of_freedom());
   }
+  if (downsampling_.has_value()) {
```
pleroy: This code and the next function would be much cleaner if `Downsampling` had its own read/write methods.

serialization/physics.proto
```diff
@@ -69,9 +69,16 @@ message DiscreteTrajectory {
     required Point fork_time = 1;
     repeated DiscreteTrajectory trajectories = 2;
   }
+  message Downsampling {
```
pleroy: Order the messages alphabetically.

serialization/physics.proto
```diff
@@ -69,9 +69,16 @@ message DiscreteTrajectory {
...
+    // The instant of the iterator; absent if it is the end of the timeline.
+    optional Point start_of_dense_timeline = 1;
+    required int64 max_dense_intervals = 2;
+    required Quantity tolerance = 3;
+  }
   repeated Litter children = 1;
   repeated InstantaneousDegreesOfFreedom timeline = 2;
   repeated int32 fork_position = 3;
+  optional Downsampling downsampling = 4;
```
pleroy: Comment that this was added in Chen.

eggrobin: Done.

serialization/physics.proto
eggrobin: Done.

physics/discrete_trajectory_body.hpp
eggrobin: This is what cppreference says: http://en.cppreference.com/w/cpp/container/map/erase.

physics/discrete_trajectory_body.hpp
eggrobin: Done.

physics/discrete_trajectory_body.hpp
eggrobin: Done.

physics/discrete_trajectory_body.hpp
eggrobin: I think this is correct, PTAL with the new more verbose identifiers.
Updating the number of dense intervals is done systematically, *after* `erase` is called.
This simply ensures that in the case where the dense timeline goes away, its start is a valid iterator.

physics/discrete_trajectory_body.hpp
eggrobin: OBE

physics/discrete_trajectory_body.hpp
eggrobin: It is.

physics/discrete_trajectory_body.hpp
eggrobin: Done.

physics/discrete_trajectory.hpp
eggrobin: Done.

physics/discrete_trajectory_body.hpp
eggrobin: OBE.

physics/discrete_trajectory.hpp
eggrobin: After the first round of downsampling, this is the same, because "added" (we only add right endpoints).
I'm not sure how to word that properly, we add points, not intervals.

physics/discrete_trajectory_body.hpp
eggrobin: (I will add a test tomorrow)

physics/discrete_trajectory.hpp
```diff
@@ -182,20 +182,56 @@ class DiscreteTrajectory : public Forkable<DiscreteTrajectory<Frame>,
...
+    Instant const& first_dense_time() const;
+    // Keeps |dense_intervals_| consistent with the new
+    // |start_of_dense_timeline_|.
+    void set_start_of_dense_timeline(TimelineConstIterator value,
+                                     Timeline const& timeline);
+
+    // Sets |dense_intervals_| to
+    // |std::distance(start_of_dense_timeline, timeline.end()) - 1|.  This is
+    // linear in the value of |dense_intervals_|.
+    void recount_dense_intervals(Timeline const& timeline);
```
pleroy: Should probably be called `RecountDenseIntervals` as it is expensive.

physics/discrete_trajectory_body.hpp
```diff
@@ -236,17 +233,16 @@ void DiscreteTrajectory<Frame>::ForgetBefore(Instant const& time) {
...
   // Get an iterator denoting the first entry with time >= |time|.  Remove all
   // the entries that precede it.  This preserves any entry with time == |time|.
-  auto it = timeline_.lower_bound(time);
+  auto const first_kept_in_timeline = timeline_.lower_bound(time);
+  Instant const& first_kept_time = first_kept_in_timeline->first;
   if (downsampling_.has_value() &&
-      (it == timeline_.end() ||
-       downsampling_->start_of_dense_timeline->first < it->first)) {
+      (first_kept_in_timeline == timeline_.end() ||
+       downsampling_->first_dense_time() < first_kept_in_timeline->first)) {
```
pleroy: `first_kept_time`

physics/discrete_trajectory.hpp
```diff
@@ -182,20 +182,56 @@ class DiscreteTrajectory : public Forkable<DiscreteTrajectory<Frame>,
   std::int64_t timeline_size() const override;
 
  private:
-  struct Downsampling {
+  class Downsampling {
+   public:
+    Downsampling(std::int64_t max_dense_intervals,
+                 Length tolerance,
+                 TimelineConstIterator start_of_dense_timeline,
```
pleroy: Remove this parameter and set the member to `begin()` always.

physics/discrete_trajectory_body.hpp
```diff
@@ -255,13 +251,8 @@ void DiscreteTrajectory<Frame>::SetDownsampling(
...
   CHECK(!downsampling_.has_value());
-  downsampling_.emplace();
-  downsampling_->start_of_dense_timeline = timeline_.begin();
-  downsampling_->dense_intervals =
-      std::distance(downsampling_->start_of_dense_timeline,
-                    timeline_.cend()) - 1;
-  downsampling_->max_dense_intervals = max_dense_intervals;
-  downsampling_->tolerance = tolerance;
+  downsampling_.emplace(
+      timeline_.begin(), max_dense_intervals, tolerance, timeline_);
```
pleroy: Should I be worried that the parameters are not in the right order?

physics/discrete_trajectory.hpp
```diff
@@ -182,20 +182,56 @@ class DiscreteTrajectory : public Forkable<DiscreteTrajectory<Frame>,
...
+                 Timeline const& timeline);
+
+    TimelineConstIterator start_of_dense_timeline() const;
+    // |start_of_dense_timeline()->first|, for readability.
+    Instant const& first_dense_time() const;
+    // Keeps |dense_intervals_| consistent with the new
+    // |start_of_dense_timeline_|.
+    void set_start_of_dense_timeline(TimelineConstIterator value,
```
pleroy: Should probably be called `SetStartOfDenseTimeline` as it is expensive.

physics/discrete_trajectory_body.hpp
```diff
@@ -207,26 +204,26 @@ void DiscreteTrajectory<Frame>::ForgetAfter(Instant const& time) {
...
   // entry and all the entries that follow it.  This preserves any entry with
   // time == |time|.
-  auto const it = timeline_.upper_bound(time);
+  auto const first_removed_in_timeline = timeline_.upper_bound(time);
+  Instant const& first_removed_time = first_removed_in_timeline->first;
   if (downsampling_.has_value()) {
-    if (it != timeline_.end() &&
-        downsampling_->start_of_dense_timeline->first >= it->first) {
+    if (first_removed_in_timeline != timeline_.end() &&
+        downsampling_->first_dense_time() >= first_removed_time) {
```
pleroy: I still don't understand this code.  Let's talk.

physics/discrete_trajectory_body.hpp
eggrobin: oops. Done.

physics/discrete_trajectory_body.hpp
eggrobin: Done.

physics/discrete_trajectory_body.hpp
eggrobin: Done.

eggrobin: Also added a test.

physics/discrete_trajectory.hpp
eggrobin: Done.

physics/discrete_trajectory.hpp
eggrobin: Done.

physics/discrete_trajectory.hpp
eggrobin: No, this is used for deserialization too.

physics/discrete_trajectory_test.cpp
```diff
@@ -949,6 +949,51 @@ TEST_F(DiscreteTrajectoryTest, DownsamplingSerialization) {
   }
 }
 
+TEST_F(DiscreteTrajectoryTest, DownsamplingForgetAfter) {
+  DiscreteTrajectory<World> circle;
+  DiscreteTrajectory<World> forgotten_circle;
+  circle.SetDownsampling(/*max_dense_intervals=*/50,
+                         /*tolerance=*/1 * Milli(Metre));
+  forgotten_circle.SetDownsampling(/*max_dense_intervals=*/50,
+                                     /*tolerance=*/1 * Milli(Metre));
```
pleroy: Bad indent.

physics/discrete_trajectory.hpp
```diff
@@ -194,15 +194,15 @@ class DiscreteTrajectory : public Forkable<DiscreteTrajectory<Frame>,
     Instant const& first_dense_time() const;
     // Keeps |dense_intervals_| consistent with the new
     // |start_of_dense_timeline_|.
-    void set_start_of_dense_timeline(TimelineConstIterator value,
-                                     Timeline const& timeline);
+    void SetStartOfDenseTimeline(TimelineConstIterator value,
+                                 Timeline const& timeline);
 
     // Sets |dense_intervals_| to
     // |std::distance(start_of_dense_timeline, timeline.end()) - 1|.  This is
```
pleroy: `start_of_dense_timeline_`

physics/discrete_trajectory_test.cpp
eggrobin: Done.

physics/discrete_trajectory.hpp
eggrobin: Done.

physics/discrete_trajectory_body.hpp
```diff
@@ -205,10 +205,13 @@ void DiscreteTrajectory<Frame>::ForgetAfter(Instant const& time) {
   // entry and all the entries that follow it.  This preserves any entry with
   // time == |time|.
   auto const first_removed_in_timeline = timeline_.upper_bound(time);
-  Instant const& first_removed_time = first_removed_in_timeline->first;
+  Instant const* const first_removed_time =
+      first_removed_in_timeline == timeline_.end()
+          ? nullptr
+          : &first_removed_in_timeline->first;
   if (downsampling_.has_value()) {
     if (first_removed_in_timeline != timeline_.end() &&
```
pleroy: Let's make this `first_removed_time != nullptr`, it's more explicit.

PR #1622, opened 2017-11-11
Use the downsampling

### LINKED ISSUE #228: Keep the whole history

eggrobin:
Fix #228, finally.
Also some C++ grammatical warts.


eggrobin:
With these defaults, the following 100-day orbit makes the save 14 MiB; compare with roughly 800 MiB for 3 vessels without downsampling on the same duration.
screenshot0

Issue #1447, opened 2017-06-18
KSP crash when reentering a vessel from EVA


scimas:
Not sure exactly what is causing the crash.  I'm sending an orbital station to minmus, so it has a lot of parts. There are 3 crew capacity parts, the hitchiker storage, mobile processing lab and an mk1 lander module. If I EVA out of the mk1 lander module and then try to board the ship again, ksp is crashing. The problem doesn't happen if I EVA from the other two modules and get back in. The principia logs are showing a 'System.NullReferenceException.' The logs and the save file is attached.

The vessel in question is named VCASS Minmus Station. Jebediah can be found piloting the mk1 lander module; EVA him and try to board back the mk1 lander, instant crash. Mods installed: Principia, Kerbal Engineer Redux, Kerbal Alarm Clock, MechJeb 2. MechJeb is not installed on any of the crafts though, so it can be excluded while testing.
[bug_report.zip](https://github.com/mockingbirdnest/Principia/files/1083251/bug_report.zip)

pleroy:
Decoded stack trace:

[`google/glog\src\utilities.cc:122`](https://github.com/mockingbirdnest/Principia/blob/6346e82182385798f8ee55f2b54beb0a4b3a1393/google/glog\src\utilities.cc#L122)
[`google/glog\src\logging.cc:2034`](https://github.com/mockingbirdnest/Principia/blob/6346e82182385798f8ee55f2b54beb0a4b3a1393/google/glog\src\logging.cc#L2034)
[`ksp_plugin/interface.cpp:637`](https://github.com/mockingbirdnest/Principia/blob/6346e82182385798f8ee55f2b54beb0a4b3a1393/ksp_plugin/interface.cpp#L637)

Maxence1402:
Got a related bug, but while going on EVA in orbit around Minmus, either Jeb is off the lander (metres away), either KSP crashes. I didn't see this in the Catalan version (I'm using Cauchy).

eggrobin:
The full FATAL failure:
```
F0618 00:47:19.120632  6052 interface.cpp:637] System.NullReferenceException: 

  at (wrapper managed-to-native) UnityEngine.Component:get_gameObject ()

  at principia.ksp_plugin_adapter.PrincipiaPluginAdapter+<AdvanceAndNudgeVesselsAfterPhysicsSimulation>d__93.MoveNext () [0x00000] in <filename unknown>:0 
```

eggrobin:
[...]

PR #1623, opened 2017-11-11
#1618 is the new #1447

### LINKED ISSUE #1618: crash on minmus
I got to Minmmus, and landed. When I tried to go away from the surface, the game crashed.

Here is the glog archives: https://www.dropbox.com/s/yn3i35y8vkhn4cj/Principia.rar?dl=0

And here is the KSP.log https://www.dropbox.com/s/lq9jcl4o4m72yiw/KSP.rar?dl=0

eggrobin:
Fix #1618.

We don't have a save, so this is not tested, but the trace looks similar enough to #1447 that we can reasonably guess that this is the issue.

### DIFF
--- a/ksp_plugin_adapter/ksp_plugin_adapter.cs
+++ b/ksp_plugin_adapter/ksp_plugin_adapter.cs
@@ -1206,17 +1206,17 @@ public partial class PrincipiaPluginAdapter
                 closest_physical_parent(part1).flightID);
           }
           foreach (var collider in part1.currentCollisions) {
-            if (collider.gameObject.layer == (int)UnityLayers.LocalScenery) {
-              Log.Info("Reporting collision with local scenery " +
-                       collider.name);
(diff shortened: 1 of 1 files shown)


ksp_plugin_adapter/ksp_plugin_adapter.cs
```diff
@@ -1206,18 +1206,18 @@ private System.Collections.IEnumerator
                 closest_physical_parent(part1).flightID);
           }
           foreach (var collider in part1.currentCollisions) {
-            if (collider.gameObject.layer == (int)UnityLayers.LocalScenery) {
-              Log.Info("Reporting collision with local scenery " +
-                       collider.name);
-              plugin_.ReportGroundCollision(
-                  closest_physical_parent(part1).flightID);
-            }
```
pleroy: This code was odd: let's use `collider` and later check if by any chance it was null...

PR #1636, opened 2017-11-28
Interpolation is not trivial when we are not interpolating at a node

eggrobin:
Emit the table for evaluating the 0th derivative by finite difference (aka interpolation).


pleroy:
Please put the code to compute the Cohen et al. coefficients in some `.wl` file (either the same or a separate one).  This can be done in a separate PR.

Issue #1628, opened 2017-11-22
Crash during timewarp at space center


aw1621107:
Steps to reproduce:

1. Start game, load "dev" save
2. Click "Warp to complete" for launchpad refurbishing
3. Crash around 4 hours remaining for launchpad refurbishing

Save: https://www.dropbox.com/sh/wh6qo46lbns39wn/AABZwWk_yUmXdgiy0qrWnjpQa?dl=0
Journal: https://www.dropbox.com/s/lxiyuhyn9gztzpq/JOURNAL.20171122-074220?dl=0

Required mods (as far as I can remember):

- RO
- RP-0 dev
- KCT
- Procedural parts
- Principia
- SXT
- RealFuels
- BluedogDB

Trimmed down part from macOS diagnostic report:

```
Process:               KSP [98865]
Path:                  /Users/USER/*/KSP.app/Contents/MacOS/KSP
Identifier:            unity.Squad.Kerbal Space Program
Version:               Unity Player version 5.4.0p4 (5.4.0p4)
Code Type:             X86-64 (Native)
Parent Process:        ??? [1]
Responsible:           KSP [98865]
User ID:               501

Date/Time:             2017-11-22 01:03:09.020 +0000
OS Version:            Mac OS X 10.13.1 (17B48)

System Integrity Protection: enabled

Crashed Thread:        0  Dispatch queue: com.apple.main-thread

Exception Type:        EXC_BAD_ACCESS (SIGILL)
Exception Codes:       KERN_INVALID_ADDRESS at 0x0000000000000000
Exception Note:        EXC_CORPSE_NOTIFY

Termination Signal:    Illegal instruction: 4
Termination Reason:    Namespace SIGNAL, Code 0x4
Terminating Process:   KSP [98865]

Thread 0 Crashed:: Dispatch queue: com.apple.main-thread
0   libsystem_kernel.dylib        	0x00007fff64f2ae7e __psynch_cvwait + 10
1   libsystem_pthread.dylib       	0x00007fff65066662 _pthread_cond_wait + 732
2   libc++.1.dylib                	0x00007fff62e19cb0 std::__1::condition_variable::wait(std::__1::unique_lock<std::__1::mutex>&) + 18
3   libc++.1.dylib                	0x00007fff62e1a728 std::__1::__assoc_sub_state::__sub_wait(std::__1::unique_lock<std::__1::mutex>&) + 46
[...]

PR #1644, opened 2017-12-08
Detect the vessels that have collided with a celestial and propagate them through the interface

pleroy:


pleroy:
Contributes to #1628.

PR #1664, opened 2018-01-04
Change the pile-up to be co-owned by its parts

### LINKED ISSUE #1223: Consider shared ownership of the pile ups by the parts they contain
The current scheme with `clear_pile_up` seems like a contrived way of achieving the same thing. This would mean the list of pile ups in the plugin would no longer be owning, and it would require a callback to for a pile up to remove itself from that list, but it may nevertheless be cleaner than the current solution.

pleroy:
Fix #1223.


ksp_plugin/part.cpp
```diff
@@ -132,43 +131,39 @@ void Part::ClearHistory() {
...
-    IteratorOn<std::list<PileUp>> pile_up = *containing_pile_up_;
-    for (not_null<Part*> const part : pile_up.iterator()->parts()) {
-      LOG(INFO) << "Removing part " << part->ShortDebugString()
-                << " from its pile up at " << &*pile_up.iterator();
-      part->containing_pile_up_ = std::experimental::nullopt;
-    }
-    CHECK(!is_piled_up());
-    pile_up.Erase();
-  }
+void Part::remove_from_pile_up() {
```
eggrobin: The name is a bit misleading, in that the function does not remove `this` from `containing_pile_up_->parts_` (thus it makes the `PileUp` invalid.

`reset_containing_pile_up()` for consistency with the other two functions related to this member, or `reset_and_invalidate_containing_pile_up()` to make the invalidation explicit?

ksp_plugin/plugin.cpp
```diff
@@ -1178,14 +1179,22 @@ void Plugin::WriteToMessage(
     celestial_message->set_ephemeris_index(
         ephemeris_->serialization_index_for_body(owned_celestial->body()));
   }
+
+  auto const serialization_index_for_pile_up =
```
eggrobin: As discussed, use a container.

ksp_plugin/part.cpp
pleroy: Done.

ksp_plugin/plugin.cpp
pleroy: Done.

PR #1665, opened 2018-01-06
A class to represent polynomials in the monomial basis

pleroy:


benchmarks/polynomial.cpp
```diff
@@ -0,0 +1,121 @@
...
+
+using quantities::Length;
+using quantities::Quantity;
+using quantities::SIUnit;
+using quantities::Time;
+
+namespace numerics {
+
+namespace {
+int const evaluations_per_iteration = 1000;
```
eggrobin: constexpr int

benchmarks/polynomial.cpp
```diff
@@ -0,0 +1,121 @@
...
+      break;
+    case 16:
+      EvaluatePolynomialInMonomialBasis<double, Time, 16>(state);
+      break;
+    default:
+      LOG(FATAL) << "Degree " << degree << " in BM_EvaluatePolynomialDouble";
+  }
+}
+
+void BM_EvaluatePolynomialInMonomialBasisQuantity(benchmark::State& state) {
```
eggrobin: Add a benchmark with vectors, and one with points (possibly one with degrees of freedom too).

numerics/polynomial.hpp
```diff
@@ -0,0 +1,103 @@
...
+//       Value,
+//       Quotient<Difference<Value>,
+//                Exponentiation<Difference<Argument>, order>>>;
+//
+//   template<typename Value, typename Argument>
+//   using Derivative = NthDerivative<Value, Argument, 1>;
+//
+// Unfortunately VS2015 is buggy and this interacts poorly with the
+// std::integer_sequence below (we get the wrong types).  Revisit once MSFT has
+// fixed their bugs.
```
eggrobin: I'm not sure this is going to happen unless we make a bug report, considering that https://developercommunity.visualstudio.com/content/problem/106570/incorrect-compilation-error-on-c-code-involving-te.html only kicked the bug down the road to https://developercommunity.visualstudio.com/content/problem/165535/incorrect-compilation-error-on-c-operators-and-dec.html.

Maybe we should consider switching to another compiler front-end...

quantities/quantities_body.hpp
```diff
@@ -298,13 +292,45 @@ constexpr bool Quantity<D>::operator!=(Quantity const& right) const {
...
+                "Invalid angle exponent");
   message->set_dimensions(D::representation);
   message->set_magnitude(magnitude_);
 }
 
 template<typename D>
 Quantity<D> Quantity<D>::ReadFromMessage(
     serialization::Quantity const& message) {
+  static_assert(D::length_is_serializable,
+                "Invalid length exponent");
```
eggrobin: Should we keep this stuff in the message? it makes the messages larger, adds quite a bit of ugly code, and I have never seen it fail (we never change the dimensions of a field, we add new fields instead).

benchmarks/polynomial.cpp
pleroy: Done.

benchmarks/polynomial.cpp
pleroy: This will be in a future PR.

numerics/polynomial.hpp
pleroy: I guess I should create a bug with MSFT.

quantities/quantities_body.hpp
pleroy: Let's leave it that way for now.

PR #1667, opened 2018-01-07
Custom integrators

### LINKED ISSUE #1666: Support setting the integrator parameters in the gravity_models file
@GregroxMun makes strange systems that are stable but completely confuse our default symmetric linear multistep method with its default step size.

Make it possible to set the ephemeris (and history and prolongation) parameters in the gravity_models file.

eggrobin:
Fix #1666.

### DIFF
--- a/ksp_plugin/interface.cpp
+++ b/ksp_plugin/interface.cpp
@@ -121,6 +121,24 @@ ParseAdaptiveStepSizeIntegrator(std::string const& integrator_kind) {
       Barycentric>::NewtonianMotionEquation>::ReadFromMessage(message);
 }
 
+Ephemeris<Barycentric>::FixedStepParameters MakeFixedStepParameters(
+   ConfigurationFixedStepParameters const& parameters) {
+  return Ephemeris<Barycentric>::FixedStepParameters(
+      ParseFixedStepSizeIntegrator(parameters.fixed_step_size_integrator),
+      ParseQuantity<Time>(parameters.integration_step_size));
+}
+
+Ephemeris<Barycentric>::AdaptiveStepParameters MakeAdaptiveStepParameters(
+   ConfigurationAdaptiveStepParameters const& parameters) {
+  // It is erroneous for a psychohistory integration to fail, so the |max_steps|
+  // must be unlimited.
+  return Ephemeris<Barycentric>::AdaptiveStepParameters(
+      ParseAdaptiveStepSizeIntegrator(parameters.adaptive_step_size_integrator),
+      /*max_steps=*/std::numeric_limits<std::int64_t>::max(),
+      ParseQuantity<Length>(parameters.length_integration_tolerance),
+      ParseQuantity<Speed>(parameters.speed_integration_tolerance));
+}
+
 serialization::GravityModel::Body MakeGravityModel(
     BodyParameters const& body_parameters) {
   // Logging operators would dereference a null C string.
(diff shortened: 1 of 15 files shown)


ksp_plugin/plugin.hpp
```diff
@@ -122,6 +122,15 @@ class Plugin {
       serialization::GravityModel::Body const& gravity_model,
       serialization::InitialState::Keplerian::Body const& initial_state);
 
+  virtual void SetEphemerisParameters(
```
pleroy: The word "set" is a bit of a misnomer since it seems to imply that this can be set at any time, while it's only at initialization.  I would call this `InitializeEphemerisParameters`.  Same for the history and psychohistory and for the interface methods.

ksp_plugin_adapter/ksp_plugin_adapter.cs
```diff
@@ -2362,6 +2362,39 @@ private enum UnityLayers {
     reset_rsas_target_ = true;
   }
 
+  private static void SetIntegrators(IntPtr plugin,
```
pleroy: `InitializeIntegrators`.

ksp_plugin_adapter/ksp_plugin_adapter.cs
```diff
@@ -2410,34 +2446,39 @@ private enum UnityLayers {
              Log.Fatal("missing Cartesian initial state for " + body.name);
           }
           int? parent_index = body.orbit?.referenceBody.flightGlobalsIndex;
+          // GetUniqueValue resp. GetAtMostOneValue corresponding to required
+          // resp. optional in principia.serialization.GravityModel.Body.
```
pleroy: Maybe this intent (required vs. optional == unique vs. atmost) should be documented in `config_node_extensions.cs`?

serialization/journal.proto
```diff
@@ -1478,6 +1478,32 @@ message SetBufferedLogging {
...
 }
 
+message SetEphemerisParameters {
+  extend Method {
+    optional SetEphemerisParameters extension = 5148;
+  }
+  message In {
+    required fixed64 plugin = 1 [(pointer_to) = "Plugin", (is_subject) = true];
+    // Must be the name of one of the values of FixedStepSizeIntegrator.Kind.
+    required string fixed_step_size_integrator_kind = 2;
```
pleroy: Let's move these two parameters to a message `FixedStepConfigurationParameters` near the beginning of this file and use it here and in the next entry point.

serialization/journal.proto
```diff
@@ -1518,6 +1544,20 @@ message SetPlottingFrame {
...
 }
 
+message SetPsychohistoryParameters {
+  extend Method {
+    optional SetPsychohistoryParameters extension = 5150;
+  }
+  message In {
+    required fixed64 plugin = 1 [(pointer_to) = "Plugin", (is_subject) = true];
+    // Must be the name of one of the values of AdaptiveStepSizeIntegrator.Kind.
+    required string adaptive_step_size_integrator_kind = 2;
```
pleroy: Same, let's create a message `AdaptiveStepConfigurationParameters`.

ksp_plugin/interface.cpp
```diff
@@ -89,6 +93,33 @@ namespace {
 int const chunk_size = 64 << 10;
 int const number_of_chunks = 8;
 
+FixedStepSizeIntegrator<Ephemeris<Barycentric>::NewtonianMotionEquation> const&
+ParseFixedStepSizeIntegrator(std::string const& integrator_kind) {
```
pleroy: This is called "make" for the existing methods, so `MakeFixedStepSizeIntegrator`, etc.

ksp_plugin/plugin.cpp
```diff
@@ -198,8 +216,9 @@ void Plugin::EndInitialization() {
   }
 
   // Construct the ephemeris.
-  ephemeris_ = solar_system.MakeEphemeris(default_ephemeris_fitting_tolerance,
-                                          DefaultEphemerisParameters());
+  ephemeris_ = solar_system.MakeEphemeris(
+      default_ephemeris_fitting_tolerance,
+      ephemeris_parameters_.value_or(DefaultEphemerisParameters()));
```
pleroy: Instead of using an `optional` I would just initialize `ephemeris_parameters_` in the constructor of the `Plugin` like the other parameters.

ksp_plugin_adapter/ksp_plugin_adapter.cs
```diff
@@ -2362,6 +2362,39 @@ private enum UnityLayers {
     reset_rsas_target_ = true;
   }
 
+  private static void SetIntegrators(IntPtr plugin,
+                                     ConfigNode gravity_model_config) {
+   if (gravity_model_config == null) {
+     return;
+   }
+   var ephemeris_parameters =
```
pleroy: I would actually like it if `sol_gravity_model.cfg` had explicit configurations for the integrators.  This would not be journal-incompatible as we use KSP for the journal, it would serve as an example for users, and it would make things more explicit.  Doesn't need to be in this PR, but maybe a TODO?

ksp_plugin_adapter/ksp_plugin_adapter.cs
```diff
@@ -2390,13 +2425,14 @@ private enum UnityLayers {
       try {
         ConfigNode initial_states = GameDatabase.Instance.GetConfigs(
             principia_initial_state_config_name_)[0].config;
-        plugin_ =
-            Interface.NewPlugin(initial_states.GetValue("game_epoch"),
-                                initial_states.GetValue("solar_system_epoch"),
-                                Planetarium.InverseRotAngle);
```
pleroy: Why the changes?  I preferred it this way.

ksp_plugin/plugin.cpp
eggrobin: As discussed, that doesn't work.

ksp_plugin/plugin.hpp
eggrobin: Done.

ksp_plugin/interface.cpp
eggrobin: No, since this just parses a string field. Could have `Make` functions taking the `ConfigurationParameters`.

ksp_plugin_adapter/ksp_plugin_adapter.cs
eggrobin: Done.

ksp_plugin_adapter/ksp_plugin_adapter.cs
eggrobin: Unique

ksp_plugin_adapter/ksp_plugin_adapter.cs
eggrobin: As discussed, this should actually be in a separate config node (and correspondingly in a separate proto in astronomy).

Changed the C#, opened #1669 to track this.

ksp_plugin_adapter/ksp_plugin_adapter.cs
eggrobin: Done. Not *quite* `==`, since in some circumstances we make `required` fields optional by defaulting to the parameters of the game.

serialization/journal.proto
eggrobin: Done.

serialization/journal.proto
eggrobin: Done.

ksp_plugin_adapter/ksp_plugin_adapter.cs
```diff
@@ -2362,36 +2364,46 @@ private enum UnityLayers {
     reset_rsas_target_ = true;
   }
 
-  private static void SetIntegrators(IntPtr plugin,
-                                     ConfigNode gravity_model_config) {
-   if (gravity_model_config == null) {
+  private static void InitializeIntegrators(
+      IntPtr plugin,
+      ConfigNode numerics_parameters_config) {
+   if (numerics_parameters_config == null) {
```
pleroy: The words "parameters config" read a bit funny.  I wonder if `principia_numerics_parameterization` would work better.

ksp_plugin_adapter/ksp_plugin_adapter.cs
```diff
@@ -2400,38 +2412,33 @@ private enum UnityLayers {
...
     plugin_construction_ = DateTime.Now;
     Dictionary<String, ConfigNode> name_to_gravity_model = null;
-    var gravity_model_configs =
-        GameDatabase.Instance.GetConfigs(principia_gravity_model_config_name_);
-    var cartesian_configs =
-        GameDatabase.Instance.GetConfigs(principia_initial_state_config_name_);
-    ConfigNode gravity_model_config = null;
-    if (gravity_model_configs.Length == 1) {
-      gravity_model_config = gravity_model_configs[0].config;
+    ConfigNode gravity_model_config = GameDatabase.Instance.GetAtMostOneNode(
```
pleroy: Alternatively, remove `_config` from all these variables, this smells of Hungarian.

eggrobin: Done.

ksp_plugin_adapter/ksp_plugin_adapter.cs
eggrobin: As discussed, `principia_numerics_blueprint`.

PR #1671, opened 2018-01-07
Null check the right way

eggrobin:
#1667 had an incorrect null check that prevented the stock game from loading. Fix forward and catch the exception so the error isn't reported as a null plugin in a call to a plugin member.


pleroy:
retest this please

PR #1699, opened 2018-01-28
Polynomial serialization, part 1: writing

pleroy:
This requires a change in the way that quantities exponents are serialized, because we need to support large negative exponent.  Hopefully this won't be save-incompatible as we probably never used large positive exponents.

### DIFF
--- a/quantities/quantities_body.hpp
+++ b/quantities/quantities_body.hpp
@@ -31,7 +31,7 @@ class ExponentSerializer : not_constructible {
       std::int64_t position);
 
  private:
-  static constexpr std::int64_t min_exponent = -31;
+  static constexpr std::int64_t min_exponent = -24;
   static constexpr std::int64_t max_exponent = 7;
   static constexpr std::int64_t exponent_mask = 0x1F;
   static constexpr std::int64_t exponent_bits = 5;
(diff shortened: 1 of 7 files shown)


quantities/quantities_body.hpp
```diff
@@ -202,6 +201,22 @@ struct NthRootGenerator<
...
+}
+
+constexpr std::int64_t ExponentSerializer::Representation(
+    std::int64_t const exponent,
+    std::int64_t const position) {
+  // For exponents in [-16, 7] this returns the representations
+  // 0x10, 0x11, ... 0x00, ... 0x07.  For exponents in [-24, -17] this returns
+  // the representations 0x08, 0x09, ... 0x0F.  The latter used to be reserved
+  // for exponents in the range [8, 15] but we believe that we never used them,
+  // and with polynomials in the monomial base we need large negative exponents.
```
eggrobin: basis

quantities/quantities_body.hpp
```diff
@@ -9,11 +9,34 @@
...
+  // 0-based position of the dimension in the representation.
+  static constexpr std::int64_t Representation(
+      std::int64_t exponent,
+      std::int64_t position);
+
+ private:
+  static constexpr std::int64_t min_exponent = -24;
+  static constexpr std::int64_t max_exponent = 7;
+  static constexpr std::int64_t exponent_mask = 0x1F;
+  static constexpr std::int64_t exponent_bits = 5;
```
eggrobin: Do we know why it is 5 and not 6? if so, document the historical reasons; perhaps we should migrate to 8 bits in the future.

serialization/numerics.proto
```diff
@@ -39,3 +39,25 @@ message DoublePrecision {
   required Value value = 1;
   required Error error = 2;
 }
+
+// Polynomials follow.
```
eggrobin: Why this comment?

quantities/quantities_body.hpp
pleroy: I have no idea.

quantities/quantities_body.hpp
pleroy: Done.

serialization/numerics.proto
pleroy: We do that in the other protos to indicate that all the subclasses/extensions are grouped together.

PR #1701, opened 2018-01-28
Benchmark the physics DLL

eggrobin:


physics/solar_system.hpp
```diff
@@ -32,11 +35,13 @@ class SolarSystem final {
   // format for SolarSystemFile protocol buffers.
   SolarSystem(
       std::experimental::filesystem::path const& gravity_model_filename,
-      std::experimental::filesystem::path const& initial_state_filename);
+      std::experimental::filesystem::path const& initial_state_filename,
+      bool ignore_frame = false);
```
pleroy: I'd prefer to avoid the defaulted parameter by having both a 2- and a 3-argument constructor, one delegating to the other.

testing_utilities/solar_system_factory.hpp
```diff
@@ -84,6 +84,10 @@ class SolarSystemFactory : not_constructible {
     AllBodiesAndOblateness,
   };
 
+  template<typename Frame>
```
pleroy: It's odd/tasteless that only this method is parameterized by the Frame.  Shouldn't the entire class be templated?

benchmarks/dynamic_frame.cpp
```diff
@@ -85,14 +85,16 @@ void FillLinearTrajectory(Position<F> const& initial,
 }
 
 // This code is derived from Plugin::RenderTrajectory.
-std::vector<std::pair<Position<ICRFJ2000Equator>,
-                      Position<ICRFJ2000Equator>>> ApplyDynamicFrame(
+std::vector<std::pair<Position<ksp_plugin::Barycentric>,
```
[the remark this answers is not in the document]
eggrobin: Done.

benchmarks/ephemeris.cpp
```diff
@@ -73,30 +76,41 @@ namespace physics {
 
 namespace {
 
-using Flow =
-    void(not_null<DiscreteTrajectory<ICRFJ2000Equator>*> const trajectory,
-         Instant const& t,
-         Ephemeris<ICRFJ2000Equator>& ephemeris);
+using Flow = void(
+    not_null<DiscreteTrajectory<ksp_plugin::Barycentric>*> const trajectory,
```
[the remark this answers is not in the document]
eggrobin: Done.

physics/solar_system.hpp
eggrobin: As discussed, leaving as-is.

testing_utilities/solar_system_factory.hpp
eggrobin: The other methods really should return an ICRFJ2000Equator system, ignore_frame is a bit of a hack that I'd like to keep in one place.

eggrobin:
retest this please

eggrobin:
retest this please (we need to do something about the flakiness induced by races when copying the DLL...)

PR #1702, opened 2018-01-29
Fix error reported by Clang

aw1621107:
Error message:

    ./numerics/fixed_arrays_body.hpp:122:46: error: use 'template' keyword to treat 'Row' as a dependent
          template name
    typename FixedMatrix<Scalar, rows, columns>::Row<r>
                                                 ^
                                                 template

### DIFF
--- a/numerics/fixed_arrays.hpp
+++ b/numerics/fixed_arrays.hpp
@@ -72,7 +72,7 @@ class FixedMatrix final {
   };
 
   template<int r>
-  typename FixedMatrix<Scalar, rows, columns>::template Row<r> row() const;
+  typename FixedMatrix::template Row<r> row() const;
 
  private:
   std::array<Scalar, rows * columns> data_;
--- a/numerics/fixed_arrays.hpp
+++ b/numerics/fixed_arrays.hpp
@@ -72,7 +72,7 @@ class FixedMatrix final {
   };
 
   template<int r>
-  Row<r> row() const;
+  typename FixedMatrix<Scalar, rows, columns>::template Row<r> row() const;
 
  private:
   std::array<Scalar, rows * columns> data_;
--- a/numerics/fixed_arrays_body.hpp
+++ b/numerics/fixed_arrays_body.hpp
@@ -119,7 +119,7 @@ FixedMatrix<Scalar, rows, columns>::Row<r>::operator*(
 
 template<typename Scalar, int rows, int columns>
 template<int r>
-typename FixedMatrix<Scalar, rows, columns>::Row<r>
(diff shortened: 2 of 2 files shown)


pleroy:

eggrobin:
retest this please

eggrobin:
@aw1621107, it seems that this confuses MSVC:
`numerics/fixed_arrays_body.hpp(125): error C2244: 'principia::numerics::internal_fixed_arrays::FixedMatrix<Scalar,rows,columns>::row': unable to match function definition to an existing declaration`.

Maybe it is one of these cases where we need a TEMPLATE macro that expands to template for clang only? We used to have a TYPENAME like this...

eggrobin:
An alternative would be to use a trailing return type `-> Row<r>`, but this would be inconsistent with our surrounding style. @pleroy, what do you think?

pleroy:
[Here](https://github.com/mockingbirdnest/Principia/commit/6ba8b0f7df5742a369a1dba4fd2e02a08d2276f1#diff-65e887111aaadf81349bb571c869ea8b) is the commit that removed `TEMPLATE`, let's add it back.

aw1621107:
Should I close this PR then?

eggrobin:
No, add `TEMPLATE` back to macros.hpp and use it here; we still want things to build on Macintosh and Linux.

eggrobin:
Another alternative is to change the declaration to `typename FixedMatrix::template Row<r>`, which should deconfuse MSVC; see https://godbolt.org/g/awXGmV.

pleroy:
I interpret the last section of [this article on cppreference](http://en.cppreference.com/w/cpp/language/dependent_name) as saying that (1) `template` is required on the definition and (2) `template` is permitted on the declaration, where it is unnecessary.  I vote for @eggrobin 's fix.

aw1621107:
Alright, I think I implemented @eggrobin's change?

numerics/fixed_arrays.hpp
```diff
@@ -72,7 +72,7 @@ class FixedMatrix final {
   };
 
   template<int r>
-  Row<r> row() const;
+  typename FixedMatrix<Scalar, rows, columns>::template Row<r> row() const;
```
eggrobin: You don't need the parameters on `FixedMatrix`: within the definition, `FixedMatrix` without parameters refers to the class being defined. The following should be enough:
```c++
template<int r>
typename FixedMatrix::template Row<r> row() const;

```

aw1621107: Ah, didn't know that. Fixed!

pleroy:
As usual, thanks for the contribution.

PR #1705, opened 2018-02-01
Use macros for the repetitive code in Newhall

pleroy:
Lo and behold!  This removes code!

### DIFF
--- a/numerics/newhall_body.hpp
+++ b/numerics/newhall_body.hpp
@@ -96,8 +96,6 @@ struct NewhallAppromixator {
       Vector& error_estimate);
 };
 
-// TODO(phl): Use macros everywhere in this file.  It will be less error-prone.
-
 #define PRINCIPIA_NEWHALL_APPROXIMATOR_SPECIALIZATION(degree)                  \
   template<typename Vector,                                                    \
            template<typename, typename, int> class Evaluator>                  \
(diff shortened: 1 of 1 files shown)


numerics/newhall_body.hpp
```diff
@@ -96,187 +96,43 @@ struct NewhallAppromixator {
...
-    return newhall_c_matrix_monomial_degree_16_divisions_8_w04 * qv;
+#define PRINCIPIA_NEWHALL_APPROXIMATOR_SPECIALIZATION(degree)                  \
+  template<typename Vector,                                                    \
+           template<typename, typename, int> class Evaluator>                  \
+  struct NewhallAppromixator<Vector, (degree), Evaluator> {                    \
+    static FixedVector<Vector, ((degree) + 1)> HomogeneousCoefficients(        \
+        FixedVector<Vector, 2 * divisions + 2> const& qv,                      \
+        Vector& error_estimate) {                                              \
+      error_estimate =                                                         \
+          newhall_c_matrix_чебышёв_degree_##degree##_divisions_8_w04.row<      \
```
eggrobin: I wonder whether we could generate std::arrays or tuples or something to obviate the need for these specializations...

pleroy: Let's not generate code when we have a perfectly good solution in the language.

eggrobin: No, I mean since we generate newhall_c_matrix_чебышёв_degree_7_divisions_8_w04 etc. already, perhaps we could generate those as a tuple so that the token joining here becomes unnecessary and it can just be done via template magic. We still need the macros for the cases, where the degree is not a constant expression.

PR #1713, opened 2018-02-04
Make sure that all windows stay on screen (a bit)

### LINKED ISSUE #1677: Principia window got saved off-screen, no way to recover
Not sure how, but the main Principia window got saved off the screen in one of my games.  i.e., the relevant part of the .sfs file looked like the below.  That puts the window off the screen.  So when I load the game, I can't see the window.  Switching between space center, tracking station, map view, vessel view, etc. does not bring the window back, nor does quitting KSP and restarting and reloading the game.  Only way to get the window back is to manually edit the .sfs file in a text editor and change main_window_x and main_window_y and reload the game.  

Suggest maybe a check when the game is first loaded that the main window is visible.  

SCENARIO
{
	name = PrincipiaPluginAdapter
	scene = 5, 7, 8
	show_main_window_ = True
	main_window_x_ = 1407
	main_window_y_ = 911

Using KSP 1.2.2 with Cristoffel on Debian Linux.

pleroy:
Fix #1677.

### DIFF
--- a/ksp_plugin_adapter/flight_planner.cs
+++ b/ksp_plugin_adapter/flight_planner.cs
@@ -39,18 +39,7 @@ class FlightPlanner : WindowRenderer {
                               func       : RenderPlanner,
                               text       : "Flight plan",
                               options    : UnityEngine.GUILayout.MinWidth(500));
-      const float min_width_on_screen = 50;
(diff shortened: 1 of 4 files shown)


pleroy:
Tested in the game.

PR #1727, opened 2018-02-18
Unify scalar multiplications and divisions of R3Elements and Multivectors

pleroy:


quantities/quantities.hpp
```diff
@@ -142,21 +143,17 @@ template<>
...
 template<>
 __m128d ToM128D(double x);
-
-// A type trait for testing if a type is a quantity.
-template<typename T>
-struct is_quantity : std::is_arithmetic<T>, not_constructible {};
-template<typename D>
-struct is_quantity<Quantity<D>> : std::true_type, not_constructible {};
+template<>
+__m128d ToM128D(int x);
```
eggrobin: Declare two templatized overloads, one for arbitrary types (with an is_arithmetic `static_assert`) and one for quantities.

Since the `Quantity` one actually takes a `Quantity` (and not `Wide<Quantity>` or something like that when that appears, or at least will have a separate overload for that), there will be no need for SFINAE tricks here.

geometry/grassmann.hpp
```diff
@@ -255,33 +257,23 @@ Multivector<Scalar, Frame, rank> operator-(
...
-Multivector<Scalar, Frame, rank> operator/(
-    Multivector<Scalar, Frame, rank> const& left,
-    double right);
-
-template<typename LDimension, typename RScalar, typename Frame, int rank>
-Multivector<Product<Quantity<LDimension>, RScalar>, Frame, rank>
-operator*(Quantity<LDimension> const& left,
+template<typename LScalar, typename RScalar, typename Frame, int rank,
+         typename = std::enable_if_t<is_quantity<LScalar>::value>>
+Multivector<Product<LScalar, RScalar>, Frame, rank>
```
eggrobin: Note that if you want this to work with wide quantities, a `Quantity operator*(Wide<Quantity>, Quantity)` has to be declared.

This is OK: we can declare the operators, which allows their usage in a `decltype`, and put a failing dependent `static_assert` in their definition to prevent them from being used, e.g., https://godbolt.org/g/DbUP24.

It makes things easier to understand and gives us good diagnostics, so this is probably the way to go.

quantities/quantities.hpp
pleroy: Done.

geometry/grassmann.hpp
pleroy: Ack.

PR #1734, opened 2018-02-25
 Implement the Cohen-Hubbard-Oesterwinter method

pleroy:
This brings significant improvements in accuracy and in speed:
Before:
```
---------------------------------------------------------------------------------
Benchmark                                          Time           CPU Iterations
---------------------------------------------------------------------------------
BM_EphemerisSolarSystemMajorBodiesOnly/-3 90719773589 ns 90371379300 ns          1 +9.92179030813603147e-01 ua
```
After:
```
---------------------------------------------------------------------------------
Benchmark                                          Time           CPU Iterations
---------------------------------------------------------------------------------
BM_EphemerisSolarSystemMajorBodiesOnly/-3 55156286341 ns 54912352000 ns          1 +9.92179030813603147e-01 ua
```

### DIFF
--- a/integrators/symmetric_linear_multistep_integrator_test.cpp
+++ b/integrators/symmetric_linear_multistep_integrator_test.cpp
@@ -245,8 +245,8 @@ void TestConvergence(Integrator const& integrator,
 
 #if !defined(_DEBUG)
   EXPECT_THAT(RelativeError(integrator.order, q_convergence_order),
-              Lt(0.05));
-  EXPECT_THAT(q_correlation, IsNear(1.0, /*tolerance=*/1.01));
+              Lt(0.02));
+  EXPECT_THAT(q_correlation, IsNear(1.0, /*tolerance=*/1.0005));
 #endif
   double const v_convergence_order = Slope(log_step_sizes, log_p_errors);
   double const v_correlation =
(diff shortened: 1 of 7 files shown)


integrators/symmetric_linear_multistep_integrator_test.cpp
```diff
@@ -245,17 +245,17 @@ void TestConvergence(Integrator const& integrator,
...
 #endif
   double const v_convergence_order = Slope(log_step_sizes, log_p_errors);
   double const v_correlation =
       PearsonProductMomentCorrelationCoefficient(log_step_sizes, log_p_errors);
   LOG(INFO) << "Convergence order in p : " << v_convergence_order;
   LOG(INFO) << "Correlation            : " << v_correlation;
 #if !defined(_DEBUG)
-  EXPECT_THAT(RelativeError(integrator.order, v_convergence_order), Lt(0.035));
-  EXPECT_THAT(v_correlation, IsNear(1.0, /*tolerance=*/1.02));
+  EXPECT_THAT(RelativeError(integrator.order, v_convergence_order), Lt(0.02));
```
eggrobin: Should the matcher be IsNear rather than Lt here and above?

pleroy: As discussed, this doesn't work too well.

PR #1735, opened 2018-02-25
Implement Sqrt using intrinsics

pleroy:
Also remove the usages of `Pow` in the Newhall code.
Before:
```
---------------------------------------------------------------------------------
Benchmark                                          Time           CPU Iterations
---------------------------------------------------------------------------------
BM_EphemerisSolarSystemMajorBodiesOnly/-3 55156286341 ns 54912352000 ns          1 +9.92179030813603147e-01 ua
```
After:
```
---------------------------------------------------------------------------------
Benchmark                                          Time           CPU Iterations
---------------------------------------------------------------------------------
BM_EphemerisSolarSystemMajorBodiesOnly/-3 52191586623 ns 51745531700 ns          1 +9.92179030813603147e-01 ua
```

### DIFF
--- a/base/macros.hpp
+++ b/base/macros.hpp
@@ -163,6 +163,9 @@ inline void noreturn() { std::exit(0); }
 #  error "What compiler is this?"
 #endif
 
+// We assume that the processor is at least a Nehalem.
+#define PRINCIPIA_USE_SSE2_INTRINSICS 1
+
 // Thread-safety analysis.
 #if PRINCIPIA_COMPILER_CLANG || PRINCIPIA_COMPILER_CLANG_CL
 #  define THREAD_ANNOTATION_ATTRIBUTE__(x) __attribute__((x))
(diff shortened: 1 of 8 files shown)


base/macros.hpp
```diff
@@ -163,6 +163,9 @@ inline void noreturn() { std::exit(0); }
 #  error "What compiler is this?"
 #endif
 
+// We assume that the processor is at least a Nehalem.
```
eggrobin: Rename to PRINCIPIA_USE_SSE3_INTRINSICS, use pmmintrin (we only use SSE2, which means we require a Pentium 4, but we also require x86-64, so we actually require Prescott or newer).

base/macros.hpp
```diff
@@ -163,6 +163,9 @@ inline void noreturn() { std::exit(0); }
...
+#define PRINCIPIA_USE_SSE2_INTRINSICS 1
```
eggrobin: Define this to !_DEBUG, so that we check that both compile.

base/macros.hpp
pleroy: Done.

base/macros.hpp
pleroy: Done.

PR #1746, opened 2018-03-08
Change the default integrator

### LINKED ISSUE #1741: Bizarre trajectories in frames tied to Pol
Reported by [Eriksonn](https://forum.kerbalspaceprogram.com/index.php?/profile/177612-eriksonn/) on the forum:

There is a rather embarrassing bug in Cohen (and, we suspect, Clifford and Christoffel) where trajectories drawn in a reference frame involving both Pol and Jool (either Pol-Centred-Jool-Aligned or Jool-Pol-Barycentric) appears horribly wiggly:

screenshot2

Interestingly enough, this is already fixed by #1734.  More interestingly, we don't have a real understanding of what's happening here other than "something is ill-conditioned".  Clearly the backward difference computation is part of the problem, but there may be deeper issues.

pleroy:
This PR does a number of changes to address the instabilities in the orbits of Pol and Bop:

* It adds a benchmark for integrating the KSP system as we were only looking at the performance for the real solar system.
* It changes the default integrator from a Quinlan-Tremaine to a Blanes-Moan.  The integration step is chosen to yield similar performance.  The accuracy is degraded a bit (from 20 to 100 m) but the stability is greatly improved and the spurious oscillations caused by Quinlan-Tremaine are gone.
* It fixes a buglet in the local error analysis.
* It changes the retrobop stability analysis to use the Blanes-Moan integrator.

Fix #1741.

### DIFF
--- a/astronomy/ksp_system_test.cpp
+++ b/astronomy/ksp_system_test.cpp
@@ -35,6 +35,7 @@ using geometry::Sign;
 using geometry::Vector;
 using integrators::FixedStepSizeIntegrator;
 using integrators::BlanesMoan2002SRKN11B;
+using integrators::BlanesMoan2002SRKN14A;
 using integrators::McLachlanAtela1992Order5Optimal;
 using integrators::Quinlan1999Order8A;
 using integrators::QuinlanTremaine1990Order10;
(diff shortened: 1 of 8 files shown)


astronomy/ksp_system_test.cpp
```diff
@@ -472,10 +473,13 @@ INSTANTIATE_TEST_CASE_P(
     AllKSPSystemConvergenceTests,
     KSPSystemConvergenceTest,
     ::testing::Values(
+        // This is our preferred integrator.  For a step of 2100 s it gives a
+        // position error of about 111 m on Laythe and takes about 0.44 s of
+        // elapsed time.
```
eggrobin: Over which duration?

pleroy: Done.

PR #1750, opened 2018-03-11
Get ksp_plugin_adapter compiling with KSP 1.4.1

aw1621107:


aw1621107:
I'm stupid and forgot to switch branches before recompiling :(

pleroy:
I am guessing that the extra DLL references will cause our continuous build to protest because we are set up with warning == failure.  We'll see.

pleroy:
Ah, the continuous build is unhappy for a different reason, because the 1.4.1 DLLs don't exist on the build machine.  I'll look into this tomorrow (it's getting late here).

pleroy:
retest this please

ksp_plugin_adapter/ksp_plugin_adapter.csproj
```diff
@@ -57,16 +57,36 @@
...
       <Private>False</Private>
     </Reference>
     <Reference Include="System" />
     <Reference Include="UnityEngine">
-      <HintPath>..\..\KSP Assemblies\1.4.0\UnityEngine.dll</HintPath>
+      <HintPath>..\..\KSP Assemblies\1.4.1\UnityEngine.dll</HintPath>
+      <Private>False</Private>
+    </Reference>
+    <Reference Include="UnityEngine.CoreModule">
+      <HintPath>..\..\KSP Assemblies\1.4.1\UnityEngine.CoreModule.dll</HintPath>
```
pleroy: So as expected these references are generating warnings which cause the continuous build to fail.  There does not seem to be any way to selectively ignore linker warnings in MSBuild.

Ideally, we would want these references to be only active on Mac (and possibly Linux).  According to [this page](http://www.mono-project.com/archived/porting_msbuild_projects_to_xbuild/) it is possible to have references that are conditioned by the operating system.  Could you try this?

pleroy: FYI, on Windows, `$(OS)` is `Windows_NT`.

aw1621107: Hopefully fixed now

pleroy:
Thanks for your contribution.

PR #1761, opened 2018-03-17
More ordinary differential equations

eggrobin:


integrators/ordinary_differential_equations.hpp
```diff
@@ -34,9 +34,137 @@ using base::Status;
 using geometry::Instant;
 using numerics::DoublePrecision;
 using quantities::Difference;
+using quantities::Quotient;
 using quantities::Time;
 using quantities::Variation;
 
+// A differential equation of the form y′ = f(y, t).
+// |State| is the type of y.
+template<typename... StateTypes>
```
pleroy: `StateTypes` sounds like Hungarian to me.  Maybe `StateElements`?

integrators/ordinary_differential_equations.hpp
```diff
@@ -34,9 +34,137 @@ using base::Status;
...
+struct ExplicitFirstOrderOrdinaryDifferentialEquation final {
+  using State = std::tuple<std::vector<StateTypes>...>;
+  // NOTE(eggrobin): we cannot use |Variation| here because of a bug in MSVC
+  // when parameter packs interact with defaulted template parameters (and
+  // possibly decltype), but this is just
+  //   std::tuple<std::vector<Variation<StateTypes>>...>.
```
pleroy: Is this still true with VS2017?

integrators/ordinary_differential_equations.hpp
```diff
@@ -34,9 +34,137 @@ using base::Status;
...
+  using StateVariation = std::tuple<
+      std::vector<Quotient<Difference<StateTypes, StateTypes>, Time>>...>;
+
+  using RightHandSideComputation =
+      std::function<Status(Instant const& t,
+                           State const& state,
+                           StateVariation& derivatives)>;
```
pleroy: Pick either "variation" or "derivative" and use only one consistently.

integrators/ordinary_differential_equations.hpp
```diff
@@ -34,9 +34,137 @@ using base::Status;
...
+  struct SystemState final {
+    SystemState() = default;
+    SystemState(State const& y, Instant const& t);
+
+    std::tuple<std::vector<DoublePrecision<StateTypes>>...> y;
+    DoublePrecision<Instant> time;
+
+    friend bool operator==(SystemState const& lhs, SystemState const& rhs) {
+      return lhs.y == rhs.y && lhs.time == rhs.time;
+    }
```
pleroy: Put the definition in the `_body.hpp` file.

integrators/ordinary_differential_equations.hpp
```diff
@@ -34,9 +34,137 @@ using base::Status;
...
+      std::function<Status(Instant const& t,
+                           State const& state,
+                           StateVariation& derivatives)>;
+
+  struct SystemState final {
+    SystemState() = default;
+    SystemState(State const& y, Instant const& t);
+
+    std::tuple<std::vector<DoublePrecision<StateTypes>>...> y;
+    DoublePrecision<Instant> time;
```
pleroy: Should this be called `t`?

integrators/ordinary_differential_equations.hpp
```diff
@@ -34,9 +34,137 @@ using base::Status;
...
+
+  // We cannot use |Difference<StateTypes>| here for the same reason.  For some
+  // reason |DoublePrecision<StateTypes>| above works...
+  using SystemStateError =
+      std::tuple<std::vector<Difference<StateTypes, StateTypes>>...>;
+
+  // A functor that computes f(y, t) and stores it in |derivatives|.
+  // This functor must be called with |std::get<i>(derivatives).size()| equal to
+  // |std::get<i>(state).size()| for all i, but there is no requirement on the
+  // values in |derivatives|.
```
pleroy: Move this comment to the type declaration.

integrators/ordinary_differential_equations.hpp
```diff
@@ -34,9 +34,137 @@ using base::Status;
...
+    DoublePrecision<Instant> time;
+
+    friend bool operator==(SystemState const& lhs, SystemState const& rhs) {
+      return lhs.y == rhs.y && lhs.time == rhs.time;
+    }
+  };
+
+  // We cannot use |Difference<StateTypes>| here for the same reason.  For some
+  // reason |DoublePrecision<StateTypes>| above works...
+  using SystemStateError =
```
pleroy: What is this used for?

integrators/ordinary_differential_equations.hpp
```diff
@@ -34,9 +34,137 @@ using base::Status;
...
+  // This functor must be called with |std::get<i>(derivatives).size()| equal to
+  // |std::get<i>(state).size()| for all i, but there is no requirement on the
+  // values in |derivatives|.
+  RightHandSideComputation compute_derivative;
+};
+
+// A differential equation of the form X′ = A(X, t) + B(X, t), where exp(hA) and
+// exp(hB) are known.  |State| is the type of X.
+template<typename... StateTypes>
+struct Splitting final {
```
pleroy: I would prefer a longer, more explicit name, e.g., `SplitExplicitFirstOrderOrdinaryDifferentialEquation`.

integrators/ordinary_differential_equations.hpp
```diff
@@ -34,9 +34,137 @@ using base::Status;
...
+  struct SystemState final {
+    SystemState() = default;
+    SystemState(State const& y, Instant const& t);
+
+    std::tuple<std::vector<DoublePrecision<StateTypes>>...> y;
+    DoublePrecision<Instant> time;
+
+    friend bool operator==(SystemState const& lhs, SystemState const& rhs) {
+      return lhs.y == rhs.y && lhs.time == rhs.time;
+    }
```
pleroy: Same.

integrators/ordinary_differential_equations.hpp
```diff
@@ -34,9 +34,137 @@ using base::Status;
...
+
+  struct SystemStateError final {
+    std::vector<Displacement> position_error;
+    std::vector<Velocity> velocity_error;
+  };
+
+  // A functor that computes f(q, t) and stores it in |accelerations|.
+  // This functor must be called with |accelerations.size()| equal to
+  // |positions.size()|, but there is no requirement on the values in
+  // |acceleration|.
```
pleroy: Comment the type, not the member.  Also, `accelerations`.

integrators/ordinary_differential_equations.hpp
```diff
@@ -80,10 +208,10 @@ struct SpecialSecondOrderDifferentialEquation final {
...
   };
 
-  // A functor that computes f(q, t) and stores it in |*accelerations|.
-  // This functor must be called with |accelerations->size()| equal to
-  // |positions->size()|, but there is no requirement on the values in
-  // |*acceleration|.
+  // A functor that computes f(q, t) and stores it in |accelerations|.
+  // This functor must be called with |accelerations.size()| equal to
+  // |positions.size()|, but there is no requirement on the values in
+  // |acceleration|.
```
pleroy: Move to the type.

integrators/ordinary_differential_equations.hpp
eggrobin: It now works!

integrators/ordinary_differential_equations.hpp
eggrobin: Done.

integrators/ordinary_differential_equations.hpp
eggrobin: Done.

integrators/ordinary_differential_equations.hpp
eggrobin: I'd rather have a better name for y instead, but I cannot find one...

integrators/ordinary_differential_equations.hpp
eggrobin: Adaptive stepsize.

integrators/ordinary_differential_equations.hpp
eggrobin: No, `Splitting` needs to have the comment on the member variable declarations because of the difference between `left_flow` and `right_flow`, let's keep this consistent.

integrators/ordinary_differential_equations.hpp
eggrobin: `Splitting` is not a good name, but at least it resembles "splitting method"; "split ODE" is not really a thing.

eggrobin: `DecomposableFirstOrderDifferentialEquation`, maybe? see https://en.wikipedia.org/wiki/Strang_splitting.

integrators/ordinary_differential_equations.hpp
eggrobin: Same as above.

integrators/ordinary_differential_equations.hpp
eggrobin: Same as above.

integrators/ordinary_differential_equations.hpp
pleroy: Sounds good.

eggrobin: Done.

integrators/ordinary_differential_equations.hpp
eggrobin: Done, but I'm not convinced that it is more readable to have the declarations at namespace scope with a `typename` salad (this is a nested class).

eggrobin: This does not work, somehow MSVC gets confused and tries to use the wrong `operator==`. Reverting.

pleroy:
retest this please

PR #1774, opened 2018-03-31
Serialization of integrators

pleroy:


integrators/integrators_body.hpp
```diff
@@ -12,10 +12,86 @@
...
+// this we will need to specialize ReadFromMessage somehow.
+#define PRINCIPIA_CASE_SPRK(kind, method)                                 \
+  case serialization::FixedStepSizeIntegrator::kind:                      \
+    do {                                                                  \
+      CHECK(message.has_composition_method());                            \
+      return SprkAsSrknDeserializer<                                      \
+          ODE,                                                            \
+          methods::method,                                                \
+          methods::method::first_same_as_last>::ReadFromMessage(message); \
+    } while (true)
```
eggrobin: Do we really want `while (true)`?
It doesn't matter because of the return, but it seems odd.

integrators/integrators_body.hpp
```diff
@@ -12,10 +12,86 @@
...
+      typename ODE::Position>();
+}
+
+template<typename ODE, typename Method>
+FixedStepSizeIntegrator<ODE> const&
+SprkAsSrknDeserializer<ODE, Method, true>::ReadFromMessage(
+    serialization::FixedStepSizeIntegrator const& message) {
+  switch (message.composition_method()) {
+    case serialization::FixedStepSizeIntegrator::ABA:
+      // TODO(phl): This is using BAB for the ABA case.  Fix once we use C++17.
```
eggrobin: Just `LOG(FATAL)` here when trying to deserialize something unsupported, instead of incorrectly deserializing.

integrators/methods.hpp
```diff
@@ -79,19 +84,20 @@ struct SymplecticPartitionedRungeKutta : not_constructible {
...
-                    : composition_ == SymplecticRungeKuttaNyström::BA,
-                "requested |composition| inconsistent with the properties of "
-                "this method");
+  static_assert(
+      SymplecticPartitionedRungeKuttaMethod::first_same_as_last
+          ? composition_ == serialization::FixedStepSizeIntegrator::ABA ||
+                composition_ == serialization::FixedStepSizeIntegrator::BAB
+          : composition_ == serialization::FixedStepSizeIntegrator::BA,
+      "requested |composition| inconsistent with the properties of "
+      "this method");
```
eggrobin: Put the "unsupported" `static_assert` after the correctness assertions.
Also consider splitting this static assert into two, so that the message is more explicit (in C++17 the split could be made more legible with `if constexpr`, but in the meantime logic works).
```C++
  static_assert(
      !SymplecticPartitionedRungeKuttaMethod::first_same_as_last ||
      composition_ == serialization::FixedStepSizeIntegrator::ABA ||
      composition_ == serialization::FixedStepSizeIntegrator::BAB
      "requested |composition| must be ABA or BAB for this first-same-as-last method");
  static_assert(
      SymplecticPartitionedRungeKuttaMethod::first_same_as_last ||
      composition_ == serialization::FixedStepSizeIntegrator::BA
      "requested |composition| must be BA for this method which is not first-same-as-last");
  static_assert(composition_ != serialization::FixedStepSizeIntegrator::ABA,
                "ABA not supported until C++17");
```

integrators/integrators_body.hpp
pleroy: Silly me.

integrators/integrators_body.hpp
pleroy: Done.

integrators/methods.hpp
pleroy: Done.

PR #1785, opened 2018-04-01
A benchmark for PlotMethod2

eggrobin:
It turns out that we mostly spend time in `lower_bound`:

profile


benchmarks/benchmarks.vcxproj
```diff
@@ -9,13 +9,15 @@
...
     <ClCompile Include="..\base\status.cpp" />
+    <ClCompile Include="..\ksp_plugin\planetarium.cpp" />
     <ClCompile Include="dynamic_frame.cpp" />
     <ClCompile Include="embedded_explicit_runge_kutta_nyström_integrator.cpp" />
     <ClCompile Include="ephemeris.cpp" />
     <ClCompile Include="hexadecimal.cpp" />
     <ClCompile Include="main.cpp" />
     <ClCompile Include="newhall.cpp" />
     <ClCompile Include="perspective.cpp" />
+    <ClCompile Include="planetarium_benchmarks.cpp" />
```
pleroy: In general we don't call benchmarks `_benchmarks`.

benchmarks/planetarium_benchmarks.cpp
```diff
@@ -0,0 +1,245 @@
...
+            goes_8_orbit.StateVectors(goes_8_epoch));
+    auto goes_8_instance = ephemeris_->NewInstance(
+        {&goes_8_trajectory_},
+        Ephemeris<Barycentric>::NoIntrinsicAccelerations,
+        HistoryParameters());
+    CHECK_OK(ephemeris_->FlowWithFixedStep(goes_8_epoch + 100 * Day,
+                                           *goes_8_instance));
+  }
+
+  DiscreteTrajectory<Barycentric> const& GOES8Trajectory() const {
```
pleroy: Should this function use snake_case?

benchmarks/planetarium_benchmarks.cpp
```diff
@@ -0,0 +1,245 @@
...
+}  // namespace
+
+void RunBenchmark(benchmark::State& state,
+                  Perspective<Navigation, Camera> const& perspective) {
+  Satellites satellites;
+  Planetarium planetarium = satellites.MakePlanetarium(perspective);
+  RP2Lines<Length, Camera> lines;
+  int total_lines = 0;
+  int iterations = 0;
+  constexpr Instant now = "2000-01-21T04:41:30,5"_TT;
```
pleroy: Where does this date come from?

benchmarks/planetarium_benchmarks.cpp
```diff
@@ -0,0 +1,245 @@
...
+                 std::to_string(total_lines / iterations) + " lines within [" +
+                 DebugString(min_x) + ", " + DebugString(max_x) + "] × [" +
+                 DebugString(min_y) + ", " + DebugString(max_y) + "]");
+}
+
+void BM_PlanetariumPlotMethod2NearPolarPerspective(benchmark::State& state) {
+  RunBenchmark(state, PolarPerspective(near));
+}
+
+BENCHMARK(BM_PlanetariumPlotMethod2NearPolarPerspective);
```
pleroy: Move all the macros together at the end.

benchmarks/planetarium_benchmarks.cpp
eggrobin: It's an eclipse that happened to be in January of 2000.

benchmarks/planetarium_benchmarks.cpp
eggrobin: Done.

benchmarks/benchmarks.vcxproj
eggrobin: Yes, but if there are two translation units with the same name in the same project MSVC complains that the object files have the same name.

benchmarks/planetarium_benchmarks.cpp
eggrobin: Done.

PR #1788, opened 2018-04-02
Improvements to named_quantities

pleroy:
Now that some bugs have been fixed we can:
* Move `NthDerivative` to `named_quantities` and use it to define `Derivative`.
* Make `Inverse` more like `Sqrt`, `Square`, etc.

### DIFF
--- a/quantities/named_quantities.hpp
+++ b/quantities/named_quantities.hpp
@@ -17,6 +17,9 @@ using Product = decltype(std::declval<Left>() * std::declval<Right>());
 template<typename Left, typename Right>
 using Quotient = decltype(std::declval<Left>() / std::declval<Right>());
 
+template<typename Q>
+using Inverse = Quotient<double, Q>;
+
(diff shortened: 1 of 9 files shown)


quantities/named_quantities.hpp
```diff
@@ -32,10 +35,16 @@ using SquareRoot = NthRoot<Q, 2>;
 template<typename Q>
 using CubeRoot = NthRoot<Q, 3>;
 
-// The result type of the derivative of a |Value|-valued function with respect
-// to its |Argument|-valued argument.
+// The result type of the N-th derivative of a |Value|-valued function with
+// respect to its |Argument|-valued argument.
+template<typename Value, typename Argument, int order>
```
eggrobin: Could we have just one type alias with `int order=1`, calling it `Derivative`?

pleroy: Done, also for the plural form (`NthDerivatives`).

Issue #1777, opened 2018-03-31
Migrate to C++17


pleroy:
We seem to have support in all the compilers now.  It would have a number of advantages:
* Constant expressions for building `std::array`.
* Get rid of our own `optional`.
* Get rid of our own `filesystem`.

pleroy:
Also `shared_mutex`.

pleroy:
Now that the date/time code has been moved to C++17, I am closing this issue.

PR #1789, opened 2018-04-02
Use void_t instead of rolling our own

pleroy:


pleroy:
Part of addressing #1777.

Issue #1760, opened 2018-03-17
Deterministic floating-point arithmetic


eggrobin:
- find the right set of flags to prevent floating-point contraction while still allowing optimizations that make assumptions about the rounding mode and ignore floating-point exceptions;
- use the same (correctly-rounded) libm on all platforms.

pleroy:
Someone asked on Discord, so I thought I would clarify what this issue does.  The important part is the second bullet, "use the same (correctly-rounded) libm on all platforms".

As things stand, our saves are not portable from one platform to another or from one version of a platform to the next.  It all comes from the elementary functions libraries (the "libm" of the issue description) which behave differently between platforms and versions.  C++ doesn't mandate what functions like `std::sin` does, and various implementation return different results (macOS and Windows are rather good, Linux is pretty lame).

This makes our tests unnecessarily complicated and brittle (see for instance the tolerances [here](https://github.com/mockingbirdnest/Principia/blob/aac8f3c90045371c46e61c425073501329f41b2d/astronomy/orbit_analysis_test.cpp#L376-L379)) but more importantly it actually means that saves will change behaviour when copied to another system, or when the underlying operating system is upgraded.  This is not only a theoretical issue: a recent Windows upgrade caused the [Reach flight plan](https://www.youtube.com/watch?v=7BDxZV7UD9I) to break (#4188).

The solution is to use a library that is "correctly rounded", i.e., that returns, for each value of the argument *x* in `double`, the machine number closest to the exact mathematical value of sin(*x*).  This is, however, a **hard** problem, see [this paper](https://hal.science/hal-04474530/document) for details.

There exists one correctly-rounded open-source library, the [CORE-MATH project](https://gitlab.inria.fr/core-math/core-math).  Unfortunately, its performance is rather disappointing (5-10× slower than the native libraries) so our strategy is to have our own implementation (starting with sin and cos, which are the functions we use the most) that is reasonably fast, but detects if it is at risk of not returning a correctly-rounded result, and falls back to CORE-MATH in that case.

We've been working on this on and off since 2018, and it's not clear if it will come to fruition.  One of the main difficulties is that it's not possible to check correctness for all the values of the `double` argument, so we need careful mathematical proofs that our implementation is indeed doing what it purports to do.

PR #1802, opened 2018-04-29
A cube root.

eggrobin:
This is part of the effort on #1760.

### DIFF
--- a/numerics/cbrt.cpp
+++ b/numerics/cbrt.cpp
@@ -1,0 +1,70 @@
+﻿
+#include "numerics/cbrt.hpp"
+
+#include <pmmintrin.h>
+
+#include <cstdint>
+#include <limits>
+
+namespace principia {
+namespace numerics {
+
+constexpr std::uint64_t C = 0x2A9F7893782DA1CE;
+static const __m128d sign_bit =
+    _mm_castsi128_pd(_mm_cvtsi64_si128(0x8000'0000'0000'0000));
+static const __m128d sign_exponent_and_sixteen_bits_of_mantissa =
+    _mm_castsi128_pd(_mm_cvtsi64_si128(0xFFFF'FFF0'0000'0000));
+// No overflow or underflow occurs in intermediate computations for y ∈ [y₁, y₂].
+// NOTE(egg): the σs do not rescale enough to put the least normal or greatest
+// finite magnitudes inside the non-rescaling range; for very small and very
+// large values, rescaling occurs twice.
+constexpr double y₁ = 0x1p-225;
+constexpr double σ₁ = 0x1p-154;
+constexpr double σ₁⁻³ = 1 / (σ₁ * σ₁ * σ₁);
+constexpr double y₂ = 0x1p237;
+constexpr double σ₂ = 0x1p154;
+constexpr double σ₂⁻³ = 1 / (σ₂ * σ₂ * σ₂);
+static_assert(σ₁⁻³ * y₁ == y₂, "Incorrect σ₁");
+static_assert(σ₂⁻³ * y₂ == y₁, "Incorrect σ₂");
+double cbrt(double const y) {
+  __m128d const y_0 = _mm_set_sd(y);
+  __m128d const sign = _mm_and_pd(sign_bit, y_0);
+  __m128d const abs_y_0 = _mm_andnot_pd(sign_bit, y_0);
+  double const abs_y = _mm_cvtsd_f64(abs_y_0);
+  // TODO(egg): we take the absolute value two or three times when going through
+  // the rescaling paths; consider having a cbrt_positive function, or a
+  // cbrt_positive_unscaled function and four rescaling paths.
+  if (abs_y < y₁) {
+    if (abs_y == 0) {
+      return y;
+    }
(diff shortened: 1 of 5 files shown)


numerics/cbrt.cpp
```diff
@@ -0,0 +1,71 @@
...
+
+#include <cstdint>
+#include <limits>
+
+namespace principia {
+namespace numerics {
+
+constexpr std::uint64_t C = 0x2A9F7893782DA1CE;
+static const __m128d sign_bit =
+    _mm_castsi128_pd(_mm_cvtsi64_si128(0x8000'0000'0000'0000));
```
pleroy: In other places we condition the use of intrinsics on some preprocessor symbol and use a slow path otherwise.  Should we do the same in this function?

numerics/cbrt.cpp
```diff
@@ -0,0 +1,71 @@
...
+
+#include <pmmintrin.h>
+
+#include <cstdint>
+#include <limits>
+
+namespace principia {
+namespace numerics {
+
+constexpr std::uint64_t C = 0x2A9F7893782DA1CE;
```
pleroy: No idea what this constant does and how it was computed.  Could we have references to some literature in this code?

numerics/cbrt.cpp
```diff
@@ -0,0 +1,71 @@
...
+  // One round of 6th order Householder.
+  double const x³ = x * x * x;
+  double const x⁶ = x³ * x³;
+  double const y² = y * y;
+  double const numerator =
+      x * (x³ - abs_y) * ((5 * x³ + 17 * abs_y) * x³ + 5 * y²);
+  double const denominator =
+      (7 * x³ + 42 * abs_y) * x⁶ + (30 * x³ + 2 * abs_y) * y²;
+  return _mm_cvtsd_f64(
+      _mm_or_pd(_mm_set_sd(x - numerator / denominator), sign));
```
pleroy: Is this faster than multiplying by sign?

numerics/cbrt_test.cpp
```diff
@@ -0,0 +1,67 @@
...
+
+namespace principia {
+namespace numerics {
+
+using ::testing::Eq;
+using ::testing::Truly;
+
+class CubeRootTest : public ::testing::Test {
+ protected:
+  static std::uint64_t bits(double const x) {
```
pleroy: Should this be `Bits`?

numerics/cbrt.hpp
```diff
@@ -0,0 +1,11 @@
+﻿#pragma once
+
+namespace principia {
+namespace numerics {
+
+// Computes ∛y with a maximal error in [0.50005, 0.50022] ULPs; the result is
+// incorrectly rounded for approximately 5 inputs per million.
+double cbrt(double y);
```
pleroy: Why not `Cbrt`?

numerics/cbrt_test.cpp
```diff
@@ -0,0 +1,67 @@
...
+TEST_F(CubeRootTest, SpecialValues) {
+  EXPECT_THAT(bits(cbrt(0)), Eq(0));
+  EXPECT_THAT(bits(cbrt(-0.0)), Eq(0x8000'0000'0000'0000));
+  EXPECT_THAT(cbrt(std::numeric_limits<double>::infinity()),
+              Eq(std::numeric_limits<double>::infinity()));
+  EXPECT_THAT(cbrt(-std::numeric_limits<double>::infinity()),
+              Eq(-std::numeric_limits<double>::infinity()));
+  EXPECT_THAT(cbrt(std::numeric_limits<double>::quiet_NaN()),
+              Truly(&std::isnan<double>));
+  EXPECT_THAT(cbrt(-std::numeric_limits<double>::quiet_NaN()),
```
pleroy: What about signalling NaNs?

numerics/cbrt_test.cpp
```diff
@@ -0,0 +1,67 @@
...
+  EXPECT_THAT(cbrt(0x1.F'FFFF'FFFF'FFFFp236),
+              Eq(0x1p79 * cbrt(0x1.F'FFFF'FFFF'FFFFp-1)));
+}
+
+TEST_F(CubeRootTest, Sign) {
+  EXPECT_THAT(cbrt(-2), Eq(-cbrt(2)));
+}
+
+TEST_F(CubeRootTest, ParticularlyBadRounding) {
+  EXPECT_THAT(cbrt(0x1.14E35E87EA5DFp0), Eq(0x1.06C80FCCA8E18p0));
```
pleroy: How do we know that it's bad?  Could we log some indication of error?

numerics/cbrt.cpp
eggrobin: Yes, but we could gain a cycle by moving this or away from the critical path.

pleroy:
retest this please

numerics/cbrt.hpp
eggrobin: Done.

numerics/cbrt_test.cpp
eggrobin: Done.

PR #1806, opened 2018-05-02
Use our cube root.

eggrobin:
Following up on #1802.
This is part of the effort on #1777.


astronomy/ksp_resonance_test.cpp
```diff
@@ -254,29 +255,29 @@ TEST_F(KSPResonanceTest, MSVC_ONLY_TEST(Stock)) {
   auto const periods_at_epoch =
       ComputePeriods(*ephemeris, ephemeris->t_min());
   EXPECT_THAT(RelativeError(periods_at_epoch.at(laythe_),
-                            expected_periods_.at(laythe_)), Lt(1.5e-3));
+                            expected_periods_.at(laythe_)), IsNear(1.36e-3));
```
pleroy: Use the same number of digits as in the original code.

PR #1808, opened 2018-05-05
Decoding of base 32768 data

pleroy:


base/base32768_body.hpp
```diff
@@ -241,7 +262,47 @@ UniqueArray<std::uint8_t> Base32768Encode(Array<std::uint8_t const> input,
...
+      repertoire = &seven_bits;
+    }
+
+    // Align |data| on the output bit index.
+    data = repertoire->Decode(code_point);
+    data <<= shift;
+
+    // Fill the output with the parts of the code point belonging to each byte.
+    if (output_bit_index == 0) {
+      output.data[0] = (data >> (2 * bits_per_byte));
```
eggrobin: Could we just 0 `output` at the beginning to simplify the logic here?

base/base32768_body.hpp
```diff
@@ -241,7 +262,47 @@ UniqueArray<std::uint8_t> Base32768Encode(Array<std::uint8_t const> input,
...
+    } else {
+      output.data[0] |= (data >> (2 * bits_per_byte));
+    }
+    if (shift < 2 * bits_per_byte && output_end - output.data > 1) {
+      output.data[1] = (data >> bits_per_byte) & ((1 << bits_per_byte) - 1);
+      if (shift < bits_per_byte && output_end - output.data > 2) {
```
eggrobin: I would like to get rid of those conditionals, but the checks on `output_end - output.data` can only be removed once we have a way to compute the size of the result.

base/base32768_body.hpp
```diff
@@ -55,12 +59,11 @@ class CachingRepertoire : public Repertoire {
...
   // and must decode characters which lie within the blocks of this repertoire.
   std::array<char16_t, block_count * block_size> encoding_cache_;
-  std::array<std::conditional_t<
-                 (CeilingLog2(block_size) + CeilingLog2(block_count) > 8),
-                 std::uint16_t,
-                 std::uint8_t>,
-             std::numeric_limits<char16_t>::max()>
-      decoding_cache_;
+  // Using a C array because MSVC gets confused with an std::array.
+  std::conditional_t<(CeilingLog2(block_size) + CeilingLog2(block_count) > 8),
```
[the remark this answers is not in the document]
pleroy: Done.

base/base32768_body.hpp
pleroy: As discussed, not easily.

base/base32768_body.hpp
pleroy: Maybe later.

PR #1811, opened 2018-05-07
Do not link cbrt in random projects

eggrobin:


eggrobin:
It turns out we do need it.

PR #1814, opened 2018-05-10
Add functions to compute the length of encoded/decoded forms

pleroy:


base/base32768_body.hpp
```diff
@@ -298,13 +302,30 @@ void Base32768Decode(Array<char16_t const> input, Array<std::uint8_t> output) {
...
+    // See the comment below.
+    return ((input.size - 1) * bits_per_code_point +
+            bits_per_final_code_point) /
+           bits_per_byte;
+  }
+  else {
+    // No need to add anything before doing the division, because either we have
+    // a multiple of 15 bits, in which case the division is exact; or there is
+    // padding and the truncation has the right effect.
+    return input.size * bits_per_code_point / bits_per_byte;
```
eggrobin: Factor out the division by `bits_per_byte` and the comment, *i.e.*,
```c++
std::int64_t encoded_bits;
if (input.size > 0 && seven_bits.CanEncode(input.data[input.size - 1])) {
  encoded_bits = (input.size - 1) * bits_per_code_point + bits_per_final_code_point;
} else {
  encoded_bits = input.size * bits_per_code_point / bits_per_byte;
}
// Either we have a multiple of 15 bits, in which case the division is exact;
// or there is padding, and truncation has the right effect.
return encoded_bits / bits_per_byte;
```

pleroy: Done.

pleroy:
retest this please

PR #1815, opened 2018-05-10
More bytes

eggrobin:


testing_utilities/serialization_body.hpp
```diff
@@ -2,24 +2,24 @@
...
 #include <fstream>
 #include <iomanip>
 #include <string>
+#include <vector>
 
 namespace principia {
 namespace testing_utilities {
 
-// TODO(egg): This should use |UniqueArray<std::uint8_t>| or
-// |Array<std::uint8_t>| rather than |std::string| for non-text bytes.
```
pleroy: So, why not a `UniqueArray`?

eggrobin: Because we `push_back`, and we don't need too `release`.

PR #1816, opened 2018-05-10
Support for UTF-16 interchange

pleroy:
No implementation of base 32768 serialization/deserialization yet.


pleroy:
retest this please

journal/profiles.cpp
```diff
@@ -49,11 +49,25 @@ T DeserializePointer(Player::PointerMap const& pointer_map,
   }
 }
 
+char16_t const* DeserializeUtf16(std::string const& serialized) {
+  // The result is char16_t-null-terminated because of the way the serialized
+  // string was built in SerializeUtf16.
+  return reinterpret_cast<char16_t const*>(serialized.c_str());
```
eggrobin: Accessing a value through `*DeserializeUtf16` is explicitly undefined, see [N4659](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2017/n4659.pdf) 6.10(8).

journal/profiles.cpp
```diff
@@ -49,11 +49,25 @@ T DeserializePointer(Player::PointerMap const& pointer_map,
   }
 }
```
eggrobin: Add a comment saying that the following two functions use `std::string`s to store non-UTF-8 (UTF-16 specifically) because proto is silly and uses `std::string` for `bytes` as well as `string`.

journal/profiles.cpp
pleroy: Done.

journal/profiles.cpp
pleroy: Done.

journal/profiles.cpp
```diff
@@ -49,17 +49,23 @@ T DeserializePointer(Player::PointerMap const& pointer_map,
...
 
-char16_t const* DeserializeUtf16(std::string const& serialized) {
+// This function uses a |std::string| to store non-UTF-8 data (UTF-16
+// specifically) because proto is silly and uses |std::string| for bytes as well
+// as string.
+std::u16string DeserializeUtf16(std::string const& serialized) {
+  std::u16string result(serialized.size() / sizeof(char16_t), u'\0');
   // The result is char16_t-null-terminated because of the way the serialized
   // string was built in SerializeUtf16.
```
eggrobin: This is not needed anymore, as the `u16string` provides null termination.

PR #1823, opened 2018-05-12
1.4.3

eggrobin:
1.4.1 and 1.4.2 are still supported when targeting 1.4.3, under the assumption that they are ABI compatible.


pleroy:
retest this please

PR #1824, opened 2018-05-12
 A benchmark for plugin serialization and an optimization

pleroy:
Before:
```
-------------------------------------------------------------------------
Benchmark                                  Time           CPU Iterations
-------------------------------------------------------------------------
BM_PluginSerializationBenchmark   3366190962 ns 2683217200 ns          1   14.3157MB/s
```
After:
```
-----------------------------------------------------------------------
Benchmark                                Time           CPU Iterations
-----------------------------------------------------------------------
BM_PluginSerializationBenchmark 2367293571 ns 1638010500 ns          1   23.4505MB/s
```

### DIFF
--- a/ksp_plugin_test/benchmark.cpp
+++ b/ksp_plugin_test/benchmark.cpp
@@ -38,7 +38,7 @@ namespace ksp_plugin {
 
 // The caller takes ownership of the result, but it's inconvenient to express
 // with |std::unique_ptr|.
-Plugin const* DeserializePluginFromLines(
+std::unique_ptr<Plugin const> DeserializePluginFromLines(
     std::vector<std::string> const& lines,
     char const* const compressor,
     int& bytes_processed) {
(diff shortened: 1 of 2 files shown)


ksp_plugin_test/benchmark.cpp
```diff
@@ -76,26 +134,15 @@ void BM_PluginDeserializationBenchmark(benchmark::State& state) {
...
-    }
-    principia__DeserializePluginHexadecimal("",
-                                            0,
-                                            &deserializer,
-                                            &plugin,
-                                            compressor);
+    Plugin const* plugin = DeserializePluginFromLines(gipfeli_plugin,
+                                                      compressor,
+                                                      bytes_processed);
     principia__DeletePlugin(&plugin);
```
eggrobin: Should this be after a `PauseTiming` so we don't benchmark destruction?

ksp_plugin_test/benchmark.cpp
```diff
@@ -34,6 +36,30 @@ using testing_utilities::ReadLinesFromHexadecimalFile;
 
 namespace ksp_plugin {
 
+// The caller takes ownership of the result, but it's inconvenient to express
+// with |std::unique_ptr|.
```
eggrobin: It's not inconvenient.

pleroy: Done.

ksp_plugin_test/benchmark.cpp
pleroy: Irrelevant after the other change.

PR #1840, opened 2018-05-22
Blizzy's toolbar support

formicant:
Hi guys!

I've added some features to the mod's GUI. Please check.

Changes:
- The _Settings (main)_ window can now be hidden separately from the other mod windows using the Blizzy's toolbar button.
- The AppLauncher button can be hidden using the radio button in the _Settings_ window.
- Blizzy's toolbar buttons added:
  - top row in the screenshot below (from left to right):
    - switch Principia GUI on/off (doubles the AppLauncher button);
    - show/hide the _Settings_ (_main_) window;
    - show/hide the _Plotting frame selection_ window;
    - show/hide the _Flight plan_ window;
  - bottom row in the screenshot below (from left to right):
    - select target celestial;
    - select target vessel;
    - toggle patched conics;
    - toggle sun lens flare.

screenshot


eggrobin:
[Replying to the forum post since it provides a wider picture and a design rationale; doing it here since to keep the discussion about this change in one place.]

> While the Principia UI looks good on a 2560 × 1440 screen like in the screenshots of the OP, on smaller screens, it eats just too much space off the screen.

Honestly, while it is usable on my screen, I would not say it looks *good*. This UI has been in need of a redesign (or a design, really) for a long time.

> I’m trying now to add an option to make the mod’s UI more compact.

Thanks for doing this.

>﻿ Step 1 I’ve done is adding the Blizzy’s toolbar support. Using the Blizzy’s toolbar, the player can keep the main Principia window hidden most of the time, and thus, save some screen space.

I don’t think using Blizzy’s toolbar is the best approach here; one ends up with a button redundant with the AppLauncher’s (as you remark), and since the toolbar accommodates multiple mods, you end up having to add this visually noisy `Pr` to the Principia icons to mark them as being Principia’s.
I think it would be better to simply make the main window itself compact, instead of the giant tax form it currently looks like.

Said window would contain the following buttons:
> select target celestial;
> select target vessel;
> show/hide the Plotting frame selection window;
> show/hide the Flight plan window;

These are functions that we expect the user to interact with pretty much all the time.
Moreover, it would have a button to
> show/hide the Settings ~(main)~ window;

The settings window would then contain the logging settings, as well as weird diagnostic toggles such as the patched conics and lens flare. Those toggles are there for weird historical reasons: the lens flare one is from our earliest testing of solar eclipses in 2015, and the patched conics are for the educational purpose of showing the difference between Keplerian approximations and the real thing; they have no business being in easily-accessible main UI.

> A possible Step 2 (not implemented yet) could be compactification of the Plotting frame selection window, because this

is too sparse.

It is *far* too sparse. We had plans from the beginning to use icons instead of the lengthy descriptions, but like other UI issues, we kept working on other things instead.

> I believe,  it could be as tiny as in this sketch:

(with the mode descriptions in tooltips).

I like this, but there is a caveat: there are a lot of celestial bodies in, e.g., RealSolarSystem, so that selector for `Moon` is likely to get unwieldy, especially if you frequently change between two frames with different reference bodies (e.g. between EMB, ECI, and ECEF while planning your return from the Moon).

What do you think about keeping the left-hand-side a tree (possibly with some changes for compactness, e.g. if the Jovian system is expanded, the user probably isn't interested in Earth, so the rest of the child bodies of the Sun could be collapsed) so that it is easy to switch between reference bodies, but replacing the four toggles and their wall of text by four buttons with icons, thus roughly halving the size of the window?

> You could just keep it on the screen not bothering to hide it every now and then.

That would indeed be nice.

rsparkyc:
You may also want to look into this for hiding the principia button from the stock toolbar:
https://forum.kerbalspaceprogram.com/index.php?/topic/169509-141-toolbar-controller-for-modders/

rsparkyc:
@pleroy: Out of curiosity, what part of the Lesser GPL is most objectionable?

eggrobin:
@rsparkyc:
> You may also want to look into this for hiding the principia button from the stock toolbar:
https://forum.kerbalspaceprogram.com/index.php?/topic/169509-141-toolbar-controller-for-modders/

Independently from any licensing questions, as outlined in my reply above, we probably don't want to use the Blizzy toolbar at all (instead we'd like to make the main window very compact).
Further, we avoid depending on other mods, in particular since we are not on CKAN (so if we wanted to use Blizzy's toolbar, we would be much happier add the code than to have a dependency).

formicant:
> I don’t think using Blizzy’s toolbar is the best approach here…
I think it would be better to simply make the main window itself compact, instead of the giant tax form it currently looks like.

I just was a bit afraid to change a lot of things at once so I chose a method that was minimally invasive.

However, I agree with everything you said. So, let’s design the UI from scratch.

> …and the patched conics are for the educational purpose of showing the difference between Keplerian approximations and the real thing; they have no business being in easily-accessible main UI.

I used patched conics for interacting with other mods, e. g. displaying nodes created by MechJeb. Moreover, this is the only(?) way to get AN/DN relative to the equator. So I’d prefer to have the button at hand.

> I like this, but there is a caveat: there are a lot of celestial bodies in, e.g., RealSolarSystem, so that selector for Moon is likely to get unwieldy, especially if you frequently change between two frames with different reference bodies (e.g. between EMB, ECI, and ECEF while planning your return from the Moon).
What do you think about keeping the left-hand-side a tree (possibly with some changes for compactness…

I actually implied that there will be a tree inside the drop-down, not a plain list of bodies.

Maybe, it also worth to add a button changing the current celestial to the previous one, so having selected Earth and then Moon, we could switch between them with a single button click?

I’ll try to draw a sketch then.

I also have a couple of questions on the _Flight plan_ window:
- Can _‘max step per segment’_ and _‘tolerance’_ of the flight plan be moved to the settings window?
- ‘Create flight plan’—‘Delete flight plan’ buttons. Is there any reason why should it be done separately from manœuvre creation and deleting?

eggrobin:
> I used patched conics for interacting with other mods, e. g. displaying nodes created by MechJeb. Moreover, this is the only(?) way to get AN/DN relative to the equator. So I’d prefer to have the button at hand.

@lamont-granquist is working on making MechJeb aware of our manoeuvres, and enabling it to guide to them rather than to the stock/MechJeb ones. Mixing and matching the MechJeb- and Principia-created nodes is likely to lead to confusion, so I don't want to encourage it.

As for nodes relative to the equator, that sounds like a good feature request, I can add that for the body-centred inertial frames. Opened #1841.

> Can ‘max step per segment’ and ‘tolerance’ of the flight plan be moved to the settings window?

They must often be tuned for long flight plans, and they are per flight plan, rather than global settings, so they should not be in the main settings window. In any case, if the flight plan window is on the screen, we already have  alot of screen real estate going into the actual manoeuvre editors, so we can afford those settings.

> ‘Create flight plan’—‘Delete flight plan’ buttons. Is there any reason why it should be done separately from manœuvre creation and deleting?

The logic is coupled with the C++ side, so messy as it is, this gets really nontrivial to change. We'd like to rethink the append-only model of the flight plan, but this will take a long time (doing anything useful other than append-only probably involves having some level of trajectory optimization).
Let's not go there for now.

formicant:
> this gets really nontrivial to change.

What if a flight plan would be created automatically when a manœuvre is created by the player, and deleted automatically when all manœuvres are deleted?

eggrobin:
> What if a flight plan would be created automatically when a manœuvre was created by the player, and deleted automatically when all manœuvres were deleted?

The flight plan is stateful: it knows the mass of the vessel when the flight plan is created, it is anchored to the actual trajectory when it was created. Deleting and recreating an empty flight plan has an effect.

This is not a very good model, but it is beyond the scope of this issue.

scimas:
> The settings window would then contain the prediction steps & tolerance settings, the history length setting

May I suggest leaving those in the main window? For those of us with not that great CPUs or living in particularly hot regions, where both high and low end CPUs throttle anyway, the prediction steps and tolerance settings need to be tweaked frequently to maintain good FPS.
For example, I usually keep tolerance at 1e-2 and prediction steps at 1e+2 in low Kerbin orbit, but increase the tolerance to 1e0 for Mun / Minmus manoeuvre executions. As you can guess, even more tweaking is required when executing interplanetary burns. But I don't want to keep those tweaked at high values for long time either, for FPS reasons, making it a frequent task. IMO frequent enough to stay in the main window.

formicant:
Will this be OK?

Principia main window sketch up

formicant:
Or, maybe:

more compact variant

pleroy:

formicant:
I thought we had decided not to merge it.

I started implementing the new UI in a new branch, but haven’t finished yet.

eggrobin:
@formicant: yes, @pleroy's comment above is actually the continuous integration impersonating @pleroy (we started the build server yesterday).

We should reword those messages to make them less confusing.

DocRockwell:
Could the size/scale of the UI and fonts be made adjustable? Presently it's barely legible at 4k resolution.

maccollo:
>This looks good. Is the curved left-pointing arrow the "previous celestial" button? I wonder whether it would be more legible to have the name of the celestial there (or possibly two buttons, for the "recent celestials").

Perhaps a context sensitive button would work well? You could have it so that it switches based on what the camera is focused on. If it's a planet, swap the reference frame to that planet, and if you press it again while still looking at that planet you cycle through the types of reference frame. 
If you press it while focusing on your current vessel it could the same based on the current SOI that the vessel is in.

madman2003:
@formicant

Out of curiosity, what is the status of new UI? I’m personally affected by tiny fonts on 4K resolution. It would be silly for me to look into this if you’re already making something.

PR #1842, opened 2018-06-01
Improvements to the construction of solar systems

pleroy:


astronomy/time_scales_test.cpp
```diff
@@ -97,7 +97,8 @@ static_assert(j2000_tai - J2000 == 32.184 * Second, "");
     (_MSC_FULL_VER == 191326215 || \
```
eggrobin: Replace the todo by a link to the neverending bug...

physics/solar_system_body.hpp
```diff
@@ -427,31 +457,34 @@ void SolarSystem<Frame>::RemoveOblateness(std::string const& name) {
   body->clear_reference_radius();
 }
 
+#define PRINCIPIA_SET_FIELD_FROM_OPTIONAL(field)                \
```
eggrobin: I think we have something similar elsewhere, we should centralize those proto-optional-to-C++-optional macros somewhere.

Not today.

pleroy:
retest this please

PR #1859, opened 2018-06-19
TRAPPIST-1 rescalings

eggrobin:
We may also need to do something about EVE clouds if @gregroxmun uses those.

### DIFF
--- a/tools/generate_configuration.cpp
+++ b/tools/generate_configuration.cpp
@@ -75,12 +75,29 @@ void GenerateConfiguration(std::string const& game_epoch,
       (directory / gravity_model_stem).replace_extension(cfg));
   CHECK(gravity_model_cfg.good());
   gravity_model_cfg << "principia_gravity_model:NEEDS[" << needs << "] {\n";
+
+  // Find the star.  This is needed to construct Kepler orbits below.
+  std::optional<serialization::GravityModel::Body> star;
+  for (std::string const& name : solar_system.names()) {
+    serialization::GravityModel::Body const& body =
+        solar_system.gravity_model_message(name);
+    if (solar_system.has_keplerian_initial_state_message(name)) {
+      bool const is_star =
+          !solar_system.keplerian_initial_state_message(name).has_parent();
+      if (is_star) {
+        star = body;
+      }
+    }
+  }
+
   for (std::string const& name : solar_system.names()) {
     serialization::GravityModel::Body const& body =
         solar_system.gravity_model_message(name);
     gravity_model_cfg << "  body {\n";
     gravity_model_cfg << "    name                    = "
-                      << (name == "Trappist-1" ? "Sun" : name) << "\n";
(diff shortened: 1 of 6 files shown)


tools/generate_kopernicus.cpp
```diff
@@ -206,6 +225,25 @@ void GenerateKopernicusForSlippist1(
     kopernicus_cfg << "  }\n";
   }
   kopernicus_cfg << "}\n";
+  for (std::string const& name : solar_system.names()) {
+    bool const is_star =
+        !solar_system.keplerian_initial_state_message(name).has_parent();
+    if (is_star) {
```
pleroy: Preference for `if (!is_star) { ... }`. `continue` is a polite spelling of `goto`.

tools/generate_configuration.cpp
```diff
@@ -79,7 +79,8 @@ void GenerateConfiguration(std::string const& game_epoch,
     serialization::GravityModel::Body const& body =
         solar_system.gravity_model_message(name);
     gravity_model_cfg << "  body {\n";
-    gravity_model_cfg << "    name                    = " << name << "\n";
+    gravity_model_cfg << "    name                    = "
+                      << (name == "Trappist-1" ? "Sun" : name) << "\n";
```
pleroy: Let's not hard-wire names.  Let's copy the code at lines 100-110 of `generate_kopernicus.cpp` to find the star (only if given by Keplerian elements) and let's use it here and below.

tools/generate_kopernicus.cpp
```diff
@@ -148,6 +146,18 @@ void GenerateKopernicusForSlippist1(
...
+      for (char const* const curve :
+           {"ScaledIntensityCurve", "IntensityCurve", "IVAIntensityCurve"}) {
+        kopernicus_cfg << "        @" << curve << " {\n";
+        kopernicus_cfg << "          @key,*[0, ] *= 10\n";
+        kopernicus_cfg << "        }\n";
+      }
+      kopernicus_cfg << "      }\n";
+      kopernicus_cfg << "    }\n";
+    }
     if (!is_star) {
```
pleroy: `else`?

eggrobin: Done.

tools/generate_configuration.cpp
```diff
@@ -163,7 +181,9 @@ void GenerateConfiguration(std::string const& game_epoch,
       auto const& body = barycentric_system.bodies[i];
       auto const& dof = barycentric_system.degrees_of_freedom[i];
       initial_state_cfg << "  body {\n";
-      initial_state_cfg << "    name = " << body->name() << "\n";
+      initial_state_cfg << "    name = "
+                        << (body->name() == "Trappist-1" ? "Sun" : body->name())
```
pleroy: This needs fixing too.

PR #1596, opened 2017-10-12
Handle collisions with scenery as collisions

eggrobin:
Not sure whether we want this in Chasles, but it seems cleaner and faster, and should provide better logs.

Note that I could not reproduce the mysteriously ground-colliding Kerbal, so at least we would have more detailed traces if it came back.

Also note that this should not spam about vessels that are firmly on the ground, as those have the state LANDED, so will be unmanageable regardless of collisions.

PR #1586, opened 2017-10-07
Add support for colliding with the ground in the C++ code

pleroy:

PR #1865, opened 2018-06-28
Reunification of the egg and ferram4 sector

### LINKED ISSUE #848: Keep track of the vessels inside the atmosphere

eggrobin:
This is a follow-up of the work done in [Chasles](https://github.com/mockingbirdnest/Principia/wiki/Change-Log#chasles) on #1413,
see #1583, #1584, #1585, #1586, #1588, #1589, #1590, #1596, and #1601.

Specifically, this reverts #1589 now that Principia supports only versions of KSP supported by versions of FAR that add forces to the parts, rather than the rigid bodies.
This fixes #848.

In order for histories to be visible in the atmosphere, the radius multiplier is set to 1; this will result in histories behind a mountain being drawn in front.


pleroy:
Fix #848.

PR #1879, opened 2018-07-07
Bulletin C 56

eggrobin:
See Bulletin C 56:
[IERS](https://datacenter.iers.org/eop/-/somos/5Rgv/getTX/16/bulletinc-056.txt)
[Observatoire de Paris](https://hpiers.obspm.fr/iers/bul/bulc/bulletinc.56)

Note that the values of the EOP C 04 series were refined from 1996 to the present day; in particular `2000-01-01T00:00:00 UTC` is now estimated to be
`2000-01-01T00:00:00,3553880 UT1` rather than
`2000-01-01T00:00:00,3554730 UT1`; *cf.* #1587.


pleroy:
retest this please

PR #1884, opened 2018-07-29
New parameter for SetPartApparent...

pleroy:


ksp_plugin/plugin.cpp
```diff
@@ -603,13 +603,14 @@ void Plugin::SetPartApparentDegreesOfFreedom(
   // |Barycentric| centred on the given reference body.
   auto const reference_body =
       FindOrDie(celestials_, reference_body_index)->body();
+  auto const reference_rotation = PlanetariumRotation(*reference_body);
```
eggrobin: No, this should use the existing PlanetariumRotation; the new overload should not exist.

PR #1885, opened 2018-08-05
A fast-but-inaccurate implementation of Sin and Cos

pleroy:
```
Run on (4 X 3310 MHz CPU s)
08/05/18 16:36:45
--------------------------------------------------------------
Benchmark                       Time           CPU Iterations
--------------------------------------------------------------
BM_FastSinCos2?              8038 ns       8000 ns     345165
BM_FastSinCos2?              8035 ns       8045 ns     345165
BM_FastSinCos2?              8035 ns       8045 ns     345165
BM_FastSinCos2?              8139 ns       7955 ns     345165
BM_FastSinCos2?              8024 ns       8000 ns     345165
BM_FastSinCos2?              8018 ns       8045 ns     345165
BM_FastSinCos2?              8006 ns       7955 ns     345165
BM_FastSinCos2?              8009 ns       8000 ns     345165
BM_FastSinCos2?              8019 ns       8000 ns     345165
BM_FastSinCos2?              8010 ns       8045 ns     345165
BM_FastSinCos2?_mean         8033 ns       8009 ns     345165
BM_FastSinCos2?_median       8021 ns       8000 ns     345165
BM_FastSinCos2?_stddev         39 ns         36 ns     345165
```


numerics/fast_sin_cos_cycle.cpp
```diff
@@ -0,0 +1,62 @@
...
+namespace principia {
+namespace numerics {
+
+namespace {
+
+using P3 = PolynomialInMonomialBasis</*Value=*/double,
+                                     /*Argument=*/double,
+                                     /*degree=*/3,
+                                     /*Evaluator=*/EstrinEvaluator>;
+P3 sin_polynomial(P3::Coefficients{6.28316404405113818577981340506,
```
eggrobin: Document the polynomials, i.e, something like "the \<k>th order polynomial that minimizes \<error function> on \<function> under \<constraints> on \<domain>".

numerics/fast_sin_cos_cycle.hpp
```diff
@@ -0,0 +1,11 @@
+﻿
+#pragma once
+
+namespace principia {
+namespace numerics {
+
+//TODO(phl):comment
+void FastSinCosCycle(double x, double& sin, double& cos);
```
eggrobin: I'm not sure I like the word "fast" here, after all we always try to write a fast implementation for the precision that we target; we wouldn't call a correctly-rounded Pochhammer symbol "SlowPoch".
Here we are aiming for a precision that is essentially @stephentyrone's definition of a ["fistful of ULPs"](https://twitter.com/stephentyrone/status/975813418736340993)...

numerics/fast_sin_cos_cycle_test.cpp
```diff
@@ -0,0 +1,150 @@
...
+    }
+    double const cos_error = std::abs(std::cos(2 * π * x) - cos);
+    if (cos_error > max_cos_error) {
+      max_cos_error = cos_error;
+      max_cos_error_x = x;
+    }
+  }
+  // These numbers come from the Mathematica minimax computation.
+  EXPECT_LT(max_sin_error, 5.90e-7);
+  EXPECT_LT(max_cos_error, 5.28e-8);
```
eggrobin: Does mathematica tell us where the worst errors are, and if so, can we test them?

numerics/fast_sin_cos_cycle_test.cpp
```diff
@@ -0,0 +1,150 @@
...
+      max_sin_error = sin_error;
+      max_sin_error_x = x;
+    }
+    double const cos_error = std::abs(std::cos(2 * π * x) - cos);
+    if (cos_error > max_cos_error) {
+      max_cos_error = cos_error;
+      max_cos_error_x = x;
+    }
+  }
+  // Slightly larger errors here because of argument reduction.
```
eggrobin:
```c++
  // Slightly larger errors here because of argument reduction in std::sin and std::cos.
```
since our argument reduction is exact.

Perhaps it would be better to test the argument reduction by taking arbitrary double x in [-1, 1], integer n in [-1e6, 1e6], and checking the equality of the values for the arguments x+n-n and x+n, as that would check that the argument reduction is exact.

numerics/fast_sin_cos_cycle.hpp
eggrobin: SinCos2π(double cycles, etc.)?

cf. the IEEE 754-2008 sinPi etc.

numerics/fast_sin_cos_cycle_test.cpp
pleroy: Done.

numerics/fast_sin_cos_cycle.hpp
pleroy: Done.

numerics/fast_sin_cos_2π.cpp
```diff
@@ -0,0 +1,73 @@
...
+                                   -70.9934281315300026308830925659});
+P3 cos_polynomial(P3::Coefficients{-19.7391820689166533085010275514,
+                                   64.9352211775039525420259682190,
+                                   -85.2540004035261113433497714557,
+                                   56.3405940928237075549636782571});
+
+}  // namespace
+
+void FastSinCos2π(double cycles, double& sin, double& cos) {
+  // Argument reduction.  Since the unit of x is the cycle, we just drop the
```
eggrobin: There is no x.

numerics/fast_sin_cos_2π.hpp
```diff
@@ -0,0 +1,13 @@
+﻿
+#pragma once
+
+namespace principia {
+namespace numerics {
+
+// An implementation of sin and cos that favours performance at the expense of
+// accuracy.  It has an absolute error of roughly 6e-7 and takes about 25 CPU
+// cycles.
```
eggrobin: Benchmark with 64-bit argument reduction, if that's measurably slower, document the range limitation.

Consider DCHECKing and returning NaN outside the correct range (returning NaN when out of range should be free with branch prediction).

numerics/fast_sin_cos_2π.cpp
pleroy: Done.

numerics/fast_sin_cos_2π.hpp
pleroy: OBE.

PR #1903, opened 2018-08-19
First step towards a ring of polynomials: left multiplication by a scalar

pleroy:


quantities/tuples.hpp
```diff
@@ -0,0 +1,37 @@
...
+// This struct has a |Type| member which is a tuple obtained by applying
+// |Transform| to each element type in |Tuple| (which must be a tuple or an array
+// or a pair).
+template<template<typename> class Transform,
+         typename Tuple,
+         typename = std::make_integer_sequence<int, std::tuple_size_v<Tuple>>>
+struct ApplyGenerator;
+
+// This struct has a |Type| member which is a tuple made of successive
+// derivatives of |Value| with respect to |Argument| from 0 to |n|, exclusively.
```
eggrobin: [...] which is an n-element tuple of successive derivatives of |Value| with respect to |Argument|; the first element is |Value|.

numerics/polynomial_body.hpp
```diff
@@ -14,6 +15,31 @@ namespace internal_polynomial {
...
+  Multiply(Scalar const& left, Tuple const& right);
+};
+
+template<typename Scalar, typename Tuple, int... indices>
+constexpr auto
+TupleArithmetic<Scalar, Tuple, std::integer_sequence<int, indices...>>::
+    Multiply(Scalar const& left, Tuple const& right) ->
+    typename Apply<ScalarLeftMultiplier, Tuple> {
+  return {left * std::get<indices>(right)...};
+}
```
eggrobin: Would `std::apply` work?

eggrobin: Nevermind, that's completely unrelated.

quantities/tuples.hpp
pleroy: Done.

numerics/polynomial_body.hpp
pleroy: No.

PR #1906, opened 2018-08-22
Cartesian product of additive groups...

pleroy:
... and their usage to add/subtract polynomials.

### DIFF
--- a/geometry/cartesian_product.hpp
+++ b/geometry/cartesian_product.hpp
@@ -12,11 +12,13 @@ namespace internal_cartesian_product {
 // This struct exports two functions:
 //
 //   static constexpr ... Add(LTuple const& left, RTuple const& right);
-//   static constexpr... Subtract(LTuple const& left, RTuple const& right);
+//   static constexpr ... Subtract(LTuple const& left, RTuple const& right);
 //
 // where the return types are appropriate for the operation.  The two tuples may
 // be of different types (including different sizes) but corresponding element
-// types must have an addition and a subtraction.
+// types must have an addition and a subtraction.  In case of different sizes,
+// the smaller tuple is zero-extended, where zero is the neutral element of
+// addition on the relevant types.
 template<typename LTuple,
          typename RTuple,
          typename = std::make_integer_sequence<
--- a/geometry/cartesian_product.hpp
+++ b/geometry/cartesian_product.hpp
@@ -1,0 +1,34 @@
+
+#pragma once
+
+#include <algorithm>
+#include <tuple>
+#include <utility>
+
+namespace principia {
+namespace geometry {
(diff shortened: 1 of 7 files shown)


geometry/cartesian_product.hpp
```diff
@@ -0,0 +1,34 @@
...
+namespace internal_cartesian_product {
+
+// This struct exports two functions:
+//
+//   static constexpr ... Add(LTuple const& left, RTuple const& right);
+//   static constexpr... Subtract(LTuple const& left, RTuple const& right);
+//
+// where the return types are appropriate for the operation.  The two tuples may
+// be of different types (including different sizes) but corresponding element
+// types must have an addition and a subtraction.
```
eggrobin: Document the semantics of adding tuples of different sizes (the smaller one is zero-extended, where zero is the neutral element of addition on the relevant types).

geometry/cartesian_product_body.hpp
```diff
@@ -0,0 +1,80 @@
...
+namespace geometry {
+namespace internal_cartesian_product {
+
+using quantities::Apply2;
+
+template<typename LTuple, typename RTuple, int... indices>
+struct CartesianProductAdditiveGroup<LTuple, RTuple,
+                                     std::integer_sequence<int, indices...>> {
+  // The types of the result of addition and subtraction, with suitable
+  // specializations for the void case of Apply2.
```
eggrobin: Should we give those semantics to quantities::Sum so that we don't need those generators?

Alternatively, would the following work?
```c++
template<typename L, typename R>
using Sum = std::conditional_t<
    std::is_void<L>{}, R,
    std::conditional_t<std::is_void<R>{}, L,
                       quantities::Sum<L, R>>>;
```

geometry/cartesian_product.hpp
```diff
@@ -0,0 +1,34 @@
...
+#include <utility>
+
+namespace principia {
+namespace geometry {
+namespace internal_cartesian_product {
+
+// This struct exports two functions:
+//
+//   static constexpr ... Add(LTuple const& left, RTuple const& right);
+//   static constexpr... Subtract(LTuple const& left, RTuple const& right);
```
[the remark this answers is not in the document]

geometry/cartesian_product.hpp

geometry/cartesian_product_body.hpp
pleroy: Let's not go there.  These `void` are a bit hacky anyway.  An alternative would be to pass two extra 1-argument metafunctions `LTransform` and `RTransform` to be applied to the elements of the longer tuple.  (This might work better at the usage site as it would avoid specializations with `void`: we'd just pass the identity metafunction.)

pleroy: Done.

PR #1910, opened 2018-08-25
Ring structure on tuples

pleroy:
Paddock calls!  (Macbeth, I, 1)

### DIFF
--- a/geometry/cartesian_product.hpp
+++ b/geometry/cartesian_product.hpp
@@ -40,18 +40,25 @@ template<typename Scalar,
 struct CartesianProductVectorSpace;
 
 //TODO(phl):comment
-template<typename LTuple, typename RTuple,
-         int lsize_ = std::tuple_size_v<LTuple>,
-         int rsize_ = std::tuple_size_v<RTuple>>
-struct CartesianProductAlgebra;
+//template<typename LTuple, typename RTuple,
+//         int lsize_ = std::tuple_size_v<LTuple>,
+//         int rsize_ = std::tuple_size_v<RTuple>>
+//struct CartesianProductAlgebra;
 
 }  // namespace internal_cartesian_product
 
 using internal_cartesian_product::CartesianProductAdditiveGroup;
-using internal_cartesian_product::CartesianProductAlgebra;
+//using internal_cartesian_product::CartesianProductAlgebra;
 using internal_cartesian_product::CartesianProductVectorSpace;
 
 }  // namespace geometry
 }  // namespace principia
 
 #include "geometry/cartesian_product_body.hpp"
+
+
+namespace principia {
+namespace geometry {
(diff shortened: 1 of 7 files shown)


geometry/cartesian_product.hpp
```diff
@@ -37,11 +37,23 @@ struct CartesianProductAdditiveGroup;
...
+// This class exports one function:
+//
+//   static constexpr ... Multiply(LTuple const& left, RTuple const& right);
+//
+// where the return types are appropriate for the operation.  The element types
+// of the tuples must have a structure of multiplicative ring.
+template<typename LTuple, typename RTuple,
+         int lsize_ = std::tuple_size_v<LTuple>,
+         int rsize_ = std::tuple_size_v<RTuple>>
+class CartesianProductRing;
```
eggrobin: CartesianProductAdditiveGroup and CartesianProductVectorSpace make sense as names (they are the direct product of groups and direct product of vector spaces respectively), but this has to be called PolynomialRing; it is emphatically *not* the product ring.

geometry/cartesian_product.hpp
```diff
@@ -37,11 +37,23 @@ struct CartesianProductAdditiveGroup;
...
          typename = std::make_integer_sequence<int, std::tuple_size_v<Tuple>>>
-struct CartesianProductVectorSpace;
+class CartesianProductVectorSpace;
+
+// This class exports one function:
+//
+//   static constexpr ... Multiply(LTuple const& left, RTuple const& right);
+//
+// where the return types are appropriate for the operation.  The element types
+// of the tuples must have a structure of multiplicative ring.
```
eggrobin: "The element types of the tuples must have a structure of multiplicative ring" is not a good description of the requirements (it sounds like the requirement for the product ring, which this is not).

geometry/cartesian_product.hpp

geometry/cartesian_product.hpp

geometry/cartesian_product.hpp
```diff
@@ -21,19 +21,30 @@ template<typename LTuple, typename RTuple>
...
 constexpr auto operator/(Tuple const& left, Scalar const& right);
 
 }  // namespace cartesian_product
+
+namespace polynomial_ring {
+
+template<typename LTuple, typename RTuple,
+         typename = std::enable_if_t<quantities::is_tuple_v<LTuple>>,
+         typename = std::enable_if_t<quantities::is_tuple_v<RTuple>>>
+constexpr auto operator*(LTuple const& left, RTuple const& right);
```
eggrobin: Should there be a comment that the cartesian product is interpreted as the coordinates in the monomial basis (or should it say monomial basis in the namespace name)?

pleroy: Added a comment.

PR #1911, opened 2018-08-25
Operators for cartesian product

pleroy:


quantities/tuples.hpp
```diff
@@ -5,10 +5,23 @@
...
 
+#include "base/not_constructible.hpp"
+
 namespace principia {
 namespace quantities {
 namespace internal_tuples {
 
+using base::not_constructible;
+
+// A trait for finding if something is a tuple.
```
eggrobin: The property we care about is probably is_cartesian_product, which is true for tuple, array, and pair. Let's bikeshed that another day, but add a todo.

geometry/cartesian_product.hpp
```diff
@@ -5,45 +5,34 @@
...
+namespace cartesian_product {
+
+template<typename LTuple, typename RTuple>
+constexpr auto operator+(LTuple const& left, RTuple const& right);
+
+template<typename LTuple, typename RTuple>
+constexpr auto operator-(LTuple const& left, RTuple const& right);
+
+// The scalar is passed by copy in the two operators below so that the two
+// definitions can be disambiguated.
```
eggrobin: typename = void?
https://github.com/mockingbirdnest/Principia/blob/a28e05c5372005daee013e90b5a3166158fe592e/geometry/rotation.hpp#L96-L108

quantities/tuples.hpp
pleroy: Done.

geometry/cartesian_product.hpp

PR #1913, opened 2018-08-26
Add inlining to polynomial operations

pleroy:


numerics/legendre_body.hpp
```diff
@@ -3,12 +3,17 @@
...
 #include "numerics/legendre.hpp"
 
+#include "base/macros.hpp"
+
 namespace principia {
 namespace numerics {
 namespace internal_legendre {
 
+// Apparently, FORCE_INLINE has to be on the definition for it to work on
+// namespace-level functions.
```
eggrobin: If I recall correctly, FORCE_INLINE has no effect on the declaration; we should omit it there.

pleroy: Created #1914 to track this.

PR #1915, opened 2018-08-26
Derivatives of polynomials

pleroy:


numerics/combinatorics_body.hpp
```diff
@@ -0,0 +1,18 @@
+
+#pragma once
+
+#include "numerics/combinatorics.hpp"
+
+namespace principia {
+namespace numerics {
+
+constexpr int Binomial(int n, int k) {
+  return k == 0 ? 1 : n == k ? 1 : Binomial(n - 1, k - 1) + Binomial(n - 1, k);
```
eggrobin: Do we trust the compiler to do memoization on constexpr evaluation?

pleroy: I think there should be a FORCE_INLINE here.

pleroy: Hmm, this triggers C4714.  Needs more work.

PR #1924, opened 2018-09-02
Linear computation of the Legendre polynomials

pleroy:


numerics/combinatorics_test.cpp
```diff
@@ -0,0 +1,121 @@
...
+  static_assert(Binomial(3, 0) == 1, "");
+  static_assert(Binomial(3, 1) == 3, "");
+  static_assert(Binomial(3, 2) == 3, "");
+  static_assert(Binomial(3, 3) == 1, "");
+
+  static_assert(Binomial(4, 0) == 1, "");
+  static_assert(Binomial(4, 1) == 4, "");
+  static_assert(Binomial(4, 2) == 6, "");
+  static_assert(Binomial(4, 3) == 4, "");
+  static_assert(Binomial(4, 4) == 1, "");
```
eggrobin: We use C++17, so we don't empty strings.

numerics/combinatorics_test.cpp
```diff
@@ -0,0 +1,121 @@
...
+}
+
+TEST(Combinatorics, DoubleFactorial) {
+  // https://oeis.org/A000165, https://oeis.org/A001147.
```
eggrobin: http://oeis.org/A006882.

pleroy: Done.

numerics/combinatorics_test.cpp
pleroy: TIL.

PR #1925, opened 2018-09-02
Documentation for the derivation of the Legendre polynomials

pleroy:


documentation/Legendre.tex
```diff
@@ -0,0 +1,99 @@
...
+    {\DoubleFactorial{\pa{n + k - 1}}\DoubleFactorial{\pa{n - k - 1}}}
+    {\Factorial{k}\Factorial{\pa{n - k}}}
+\end{equation*}
+And finally:
+\begin{equation*}
+P_n(x) = \sum{\substack{k = 0\\ k \equiv n \pmod 2}}[n]
+  x^k \pa{-1}^{\frac{n - k}{2}}
+  \frac
+    {\DoubleFactorial{\pa{n + k - 1}}\DoubleFactorial{\pa{n - k - 1}}}
+    {\Factorial{k}\Factorial{\pa{n - k}}}
```
eggrobin: As discussed, (n-k-1)‼/(n-k)! simplifies to 1/(n-k)‼.

We could have a dedicated function for (n+k)‼/(n-k)‼, which is the double factorial version of the falling factorial (`FactorialPower[n + k, k, 2]` in *Mathematica*, with the notation (n+k)<sup>(k,2)</sup>), to avoid overflows from (n+k)‼.

Once the coefficient is in the form (n+k)<sup>(k,2)</sup>/k!, we could switch the combinatorial functions to using double, since 21! and 22! are representable in binary64 floating-point, and even higher values of n would gracefully add rounding errors to the numerator and denominator instead of causing an integer overflow.

pleroy: Done.

PR #1932, opened 2018-09-08
Geopotential data for oblate bodies

pleroy:


physics/solar_system_body.hpp
```diff
@@ -46,6 +47,7 @@ using geometry::Position;
 using geometry::RadiusLatitudeLongitude;
 using geometry::Vector;
 using geometry::Velocity;
+using numerics::FixedStrictlyLowerTriangularMatrix;
```
eggrobin: Is this using used?

serialization/astronomy.proto
```diff
@@ -23,11 +24,10 @@ message GravityModel {
     optional string reference_angle = 8;
     optional string angular_frequency = 9;
     // Oblate body.
-    optional double j2 = 11;
-    optional double c22 = 14;
-    optional double s22 = 15;
-    optional double j3 = 16;
     optional string reference_radius = 6;
+    // At most one of these two fields may be present.
```
eggrobin: Then make it a oneof?

numerics/legendre_body.hpp
```diff
@@ -45,6 +48,13 @@ LegendrePolynomial() {
           std::make_index_sequence<degree_ + 1>>::coefficients);
 }
 
+inline double LegendreNormalizationFactor(int degree, int order) {
+  int const n = degree;
+  int const m = order;
+  return Sqrt(Factorial(n - m) * (n + n + 1) * (2 - (m == 0 ? 1 : 0)) /
+              static_cast<double>(Factorial(n + m)));
```
eggrobin: (n + n + 1) * (2 - (m == 0 ? 1 : 0)) / static_cast<double>(FallingFactorial(n + m, 2 * m))

We should probably make Factorial and FallingFactorial return `double` to avoid those static_casts.

physics/solar_system_body.hpp
pleroy: Done.

numerics/legendre_body.hpp
pleroy: Done.  Logically it's an integer, let's stick to that.

serialization/astronomy.proto
pleroy: Done.

serialization/astronomy.proto
```diff
@@ -26,8 +26,10 @@ message GravityModel {
     // Oblate body.
     optional string reference_radius = 6;
     // At most one of these two fields may be present.
```
eggrobin: Redundant comment.

PR #1954, opened 2018-10-07
Documentation of the implementation details of the geopotential

pleroy:


documentation/Geopotential.tex
```diff
@@ -270,5 +270,78 @@ \subsection*{Singularities}
...
+and for $\gl$:
+\[
+\begin{dcases}
+\cos \gl = \frac{x}{\sqrt{x^2 + y^2}} \\
+\sin \gl = \frac{y}{\sqrt{x^2 + y^2}}
+\end{dcases}
+\]
+In the case where $\sqrt{x^2 + y^2}$ is $0$, any value will do for $\gl$, so we can for instance pick $\cos \gl = 1$ and $\sin \gl = 0$.
+
+The value of $\cos^m \gb$ may be computed by recurrence in a way that entails $\BigO\of{log_2 m}$ rounding errors.  If $m$ is even, we compute:
```
eggrobin: `log` is the product of 𝑙, 𝑜, and 𝑔.
Since we are in a big O, there is no need to specify the base.
The number of rounding errors is probably less interesting than the number of ULPs; of course the latter needs to incorporate the effect of condition numbers: squaring will double errors.

documentation/Geopotential.tex
```diff
@@ -270,5 +270,78 @@ \subsection*{Singularities}
...
+h_1 = \Floor{\frac m 2} \\
+h_2 = m - h_1 \\
+\cos^m \gb = \pa{\cos^{h_1} \gb}\pa{\cos^{h_2} \gb}
+\end{dcases}
+\]
+The value of $\cos m \gl$ and $\sin m \gl$ may similarly be computed using elementary trigonometric identities with $\BigO\of{log_2 m}$ rounding errors.  If $m$ is even, we compute:
+\[
+\begin{dcases}
+h = \frac m 2 \\
+\cos m \gl = \cos^2 h \gl - \sin^2 h \gl \\
```
eggrobin: Rewrite as (cos ℎλ + sin ℎλ) (cos ℎλ - sin ℎλ) and change the code accordingly.

Unfortunately (cos ℎλ - sin ℎλ) is still problematic because of the rounding errors in cos ℎλ and sin ℎλ.

pleroy: Changed the formula and the code.

FTR: The errors grow like `b[m_] := If[m <= 1, 0, 
  If[Mod[m, 2] == 0, 3 + 4 b[m/2], 
   3 + 2 b[(m - 1)/2] + 2 b[(m + 1)/2]]]` which is about `m^2`.  This is better than going from `m - 1` to `m` which would cause errors to grow like `a[m_] := If[m <= 1, 0, 3 + 2 a[m - 1]]` which is about `2^m`.

documentation/Geopotential.tex
```diff
@@ -270,5 +270,78 @@ \subsection*{Singularities}
 \[
 {\mathfrak B\of \gb}{\grad \mathfrak L\of \gl}_{m = 0} = 0
 \]
+\subsection*{Implementation considerations}
+A naïve implementation of the preceding formul{\ae} yields poor performance because of the computation of trigonometric functions, powers and polynomials.  It is advantageous
```
eggrobin: You can just write `formulæ` in the source code, that's more readable and XeTeX knows about Unicode.

Issue #1895, opened 2018-08-12
Our matchers are not flexible enough


pleroy:
For instance, the following, in `geopotential_test.cpp`, doesn't compile because of some obscure problem in `_`'s `DescribeTo`:
```
  EXPECT_THAT(acceleration,
              Componentwise(VanishesBefore(1 * Pow<-2>(Metre), 0),
                            VanishesBefore(1 * Pow<-2>(Metre), 0),
                            _));
```

pleroy:
Note that `An<T>()` works.

PR #1969, opened 2018-11-03
A new pattern for componentwise peeling

pleroy:
This is part of upgrading to a newer gmock.

This also ties with #1895.


pleroy:
retest this please

Issue #1955, opened 2018-10-09
Still, Too slow warp speed in MacOS


EthanWang706:
Principia Version: Descartes
MacOS Version: 10.13.6
KSP Version: 1.3.1
It has about 30-50% improvement after last update, but still, with only 3 spacecraft, RSS+RO+16K texture, I can only get 7 fps on KSC page with no time warping, before the update, it's 4-5fps. Warping to 10,000x is the maximum speed you can get without getting less than 1fps. In Windows 10, the same computer, same GameData, same KSP version, it could reach as high fps as I set in setting.

EthanWang706:
#1908

pleroy:
Added support for benchmarking on Linux and macOS in #1960 and #1961.
On Windows:
```
Run on (4 X 3310 MHz CPU s)
10/14/18 14:04:33
-------------------------------------------------------------------------------
Benchmark                                        Time           CPU Iterations
-------------------------------------------------------------------------------
BM_EphemerisMultithreadingBenchmark/3/1   97897958 ns          0 ns        100 +9.98765338665172644e+06 m +1.99941207398748249e+07 m +2.99993670217528194e+07 m
BM_EphemerisMultithreadingBenchmark/3/2   72153857 ns          0 ns        100 +9.98765338665172644e+06 m +1.99941207398748249e+07 m +2.99993670217528194e+07 m
BM_EphemerisMultithreadingBenchmark/3/3   44194468 ns          0 ns        100 +9.98765338665172644e+06 m +1.99941207398748249e+07 m +2.99993670217528194e+07 m
BM_EphemerisMultithreadingBenchmark/3/4   44230655 ns          0 ns        100 +9.98765338665172644e+06 m +1.99941207398748249e+07 m +2.99993670217528194e+07 m
BM_EphemerisMultithreadingBenchmark/3/5   43977030 ns          0 ns        100 +9.98765338665172644e+06 m +1.99941207398748249e+07 m +2.99993670217528194e+07 m
```
On Linux:
```
Run on (4 X 3361.62 MHz CPU s)
2018-10-14 12:34:57
-------------------------------------------------------------------------------
Benchmark                                        Time           CPU Iterations
-------------------------------------------------------------------------------
BM_EphemerisMultithreadingBenchmark/3/1  105878364 ns      77017 ns        100 +9.98765338665172644e+06 m +1.99941207398748249e+07 m +2.99993670217528194e+07 m
[...]

PR #1978, opened 2018-11-11
Replace all usages of std::mutex by absl::Mutex

pleroy:


pleroy:
This might help with #1955.

PR #1983, opened 2018-11-17
A faster build

pleroy:
This follows the recommendations from https://blogs.msdn.microsoft.com/vcblog/2016/10/26/recommendations-to-speed-c-builds-in-visual-studio/ (except for precompiled headers, which we might want to look into separately).


principia.props
```diff
@@ -98,14 +99,16 @@
...
       <PreprocessorDefinitions Condition="!$(PrincipiaCompilerClangLLVM)">
         _ENABLE_EXTENDED_ALIGNED_STORAGE;NOGDI;NOMINMAX;%(PreprocessorDefinitions)
       </PreprocessorDefinitions>
       <TreatWarningAsError Condition="!$(PrincipiaCompilerClangLLVM)">true</TreatWarningAsError>
       <SDLCheck>true</SDLCheck>
     </ClCompile>
     <Link>
-      <GenerateDebugInformation>true</GenerateDebugInformation>
+      <GenerateDebugInformation>DebugFastLink</GenerateDebugInformation>
+      <LinkTimeCodeGeneration>UseFastLinkTimeCodeGeneration</LinkTimeCodeGeneration>
```
eggrobin: Where is that from? I don't see it in https://blogs.msdn.microsoft.com/vcblog/2016/10/26/recommendations-to-speed-c-builds-in-visual-studio/.

pleroy: Changed to "Default".  Also moved incremental linking at a place where it actually has an effect.

PR #1985, opened 2018-11-18
Remove tesseral damping

eggrobin:
It does not appear to improve the performance of anything interesting, and it makes the code a lot more complex.

Benchmarks before and after: https://gist.github.com/eggrobin/d3a65b0f01bb57386afcef677978316d

### DIFF
--- a/physics/geopotential_body.hpp
+++ b/physics/geopotential_body.hpp
@@ -118,8 +118,7 @@ struct Geopotential<Frame>::Precomputations {
   // not stored.
   FixedLowerTriangularMatrix<double, size> DmPn_of_sin_β{uninitialized};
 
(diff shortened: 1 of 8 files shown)


physics/geopotential_body.hpp
```diff
@@ -540,14 +506,8 @@ Geopotential<Frame>::Geopotential(not_null<OblateBody<Frame> const*> body,
   harmonic_thresholds.push({Infinity<Length>(), 0, 0});
   harmonic_thresholds.push({Infinity<Length>(), 1, 0});
 
-  bool tesseral = false;
   while (!harmonic_thresholds.empty()) {
```
pleroy: Maybe explain what's happening in this loop (monotonic thresholds).

PR #1993, opened 2018-11-28
Change the generator to allow passing arrays across the interface

pleroy:
Note that the changes to `journal.proto` are experimental at this point, but they illustrate the effect on the generated code.  Also note that the commits at the end let you see the generated code.


tools/journal_proto_processor.cpp
```diff
@@ -957,11 +988,24 @@ void JournalProtoProcessor::ProcessInterchangeMessage(
     cxx_interface_type_declaration_[descriptor] +=
         "  " + field_cxx_type_[field_descriptor] + " " + field_descriptor_name +
         ";\n";
+
+    // If this field has a size, generate it now.
+    if (Contains(size_member_name_, field_descriptor)) {
```
eggrobin: Could we use something similar to `field_cs_private_type_` with a `T[]` property, so that the size is correctly set by the setter, and generate a const member function on the C++ side that wraps the pointer and size in a `base::Array<T>`?

pleroy: Good idea for the C# side, done, you can see the diff in the generated code at https://github.com/mockingbirdnest/Principia/compare/80eea50..a9df548.
Bad idea for the C++ code, we would stuff the pointer and length in an `Array` and extract them immediately for looping over the data.  Some day (and that day may never come), we shall use `std::span`.

PR #2005, opened 2018-12-04
Make it build

madman2003:
It's not easy to build principia. But i finally managed.

I don't expect this to accepted with no discussion, rather I hope it provokes the right discussion and the necessary fixes.

Most surprising was the fact that principia does not build against the latest C++ compiler. I don't really see it as my place to report, because the codebase is a bit over my head. But if I were in your shoes, this is something I wouldn't ignore for too long.


pleroy:

pleroy:
Thank you for your interest in building Principia.

We are well aware that our version of Visual Studio is a bit behind (it was released in June 2018).  Upgrading is a fairly complex affair for us as we need to (1) update a variety of computers at more-or-less the same time; and (2) work around, analyze and report to MSFT any bugs that they may have introduced.  Given that MSFT spins out a [release per week](https://docs.microsoft.com/en-us/visualstudio/releasenotes/vs2017-relnotes#visual-studio-2017-version-159-releases) these days, keeping up with every release is infeasible.  We just upgrade once every few months.

This being said I was planning to look into upgrading to 15.9.x after we've cut the next release, Erdős.  I guess I'll have to investigate the compilation error you mention, since we must move forward: we cannot afford to get stuck to an old version.

madman2003:
@pleroy In case it wasn't obvious, also try to build against the community version of MSVC, not just the enterprise one ;-)

pleroy:
@madman2003: Sure, but you cannot do that by preventing `coverage_analyser` from building, because that would break our workflow.  If you are only interested in building the mod, you just need to build `ksp_plugin_adapter`, not the entire solution.  If you want to do more you'll have to fork at least `Principia.sln`.

madman2003:
@pleroy i understand that this is not the solution, but i also didn’t want to invest in new build targets (building a solution in a non standard way is hardly the way to go). Consider it appropriate provocation, give people a bad answer and they’ll come up with a better one.

Principia is harder than any other mod to begin to imagine to contribute to. Basic issues like not being able to build are the last thing you want. And we haven’t even talked about cross compiling the whole thing from Linux to windows.

madman2003:
@pleroy probably a weird place to say it, but check that abseil-cpp fork, when you introduced the NOWINDOWS variable it stopped building on linux ;-)

Issue #1931, opened 2018-09-08
Clean up guided burn support in the ephemeris and flight plan.


eggrobin:
See the TODOs in #1930.

pleroy:
Also addressed by #2020 and #2022.  Closing as it seems that the TODOs added in #1930 have all been addressed.

PR #2010, opened 2018-12-06
Serialization of the Fine integrator

pleroy:
Another contribution to #1931.

### DIFF
--- a/integrators/integrators_body.hpp
+++ b/integrators/integrators_body.hpp
@@ -42,18 +42,16 @@ namespace integrators {
 namespace internal_integrators {
 
 template<typename ODE, typename Method, bool first_same_as_last>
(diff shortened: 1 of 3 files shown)


integrators/integrators_body.hpp
```diff
@@ -92,6 +90,71 @@ SprkAsSrknDeserializer<ODE, Method, true>::ReadFromMessage(
   }
 }
 
+template<typename RightHandSideComputation>
+struct AdaptiveStepSizeIntegratorDeserializer;
```
eggrobin: Invest in a comment here explaining that the `RightHandSideComputation` determines the type of the integrator when the method may be used for several kinds of integrators (this may not hold in general as we add more integrators, and is certainly not obvious).

PR #2012, opened 2018-12-06
Linux build fixes attempt 2

madman2003:
I can see the impression I left with attempt 1. This time I've done my best not to break anything in your workflow, if it still happens it is truly is an accident.

I've tried to make something half decent of the commit history, it may not be a 100% perfect.

My reasons for doing this are because I want to see what I can about the principia UI, which as a user is far too tiny. Hence i need it to build for me.

P.S. the new geopotential method has problems with real solar system, it asserts on the Vesta celestial body, i will try to write up a more detailed bug report about that


pleroy:

madman2003:
@pleroy GeodesyTest.LAGEOS2 test fails, is that expected?

madman2003:
As promised: https://github.com/mockingbirdnest/Principia/issues/2014

madman2003:
#2014 is indeed gone, but can someone at least confirm it was the potato roids that have no gravity field (and is that the case for all of them?)

Oh, and the LAGEOS2 test case still fails FYI.

PR #2022, opened 2018-12-08
Remove unused initial_direction from C#

### LINKED ISSUE #2021: crash when adding maneuver
When I add a new maneuver in the flight planner KSP crashes instantly.
Thanks in advance.

[Principia.zip](https://github.com/mockingbirdnest/Principia/files/2659623/Principia.zip)

pleroy:
Fix #2021.

Also remove `Plugin::prediction_parameters_`, and propagate the split of `AdaptiveStepParameters` and `GeneralizedAdaptiveStepParameters` to the flight plan, unifying them into a new `FlightPlanAdaptiveStepParameters` in the interface.


serialization/ksp_plugin.proto
```diff
@@ -121,6 +120,9 @@ message Plugin {
   // Pre-Bourbaki.
   reserved 2;
   reserved "/*celestial*/";
+  // Pre-Erdos.
```
eggrobin: Erdős.

ksp_plugin/interface_body.hpp
```diff
@@ -197,6 +208,49 @@ FromAdaptiveStepParameters(
...
+    physics::Ephemeris<Barycentric>::AdaptiveStepParameters,
+    physics::Ephemeris<Barycentric>::GeneralizedAdaptiveStepParameters>
+FromFlightPlanAdaptiveStepParameters(FlightPlanAdaptiveStepParameters const&
+                                         flight_plan_adaptive_step_parameters) {
+  serialization::AdaptiveStepSizeIntegrator message1;
+  CHECK(serialization::AdaptiveStepSizeIntegrator::Kind_IsValid(
+      flight_plan_adaptive_step_parameters.integrator_kind));
+  message1.set_kind(
+      static_cast<serialization::AdaptiveStepSizeIntegrator::Kind>(
+          flight_plan_adaptive_step_parameters.integrator_kind));
```
eggrobin: Move the definitions of max_steps, length_integration_tolerance, and speed_integration_tolerance to the beginning, and scope message1 and message2 to avoid mixing them up (note: if they are scoped, they can both be called message).

ksp_plugin/plugin.cpp
```diff
@@ -855,7 +855,7 @@ void Plugin::CreateFlightPlan(GUID const& vessel_guid,
   FindOrDie(vessels_, vessel_guid)->CreateFlightPlan(
       final_time,
       initial_mass,
-      prediction_parameters_,
+      DefaultPredictionParameters(),
```
eggrobin: We could use the prediction parameters of the relevant vessel, but then we would need to construct corresponding burn parameters; since this is consistent with the existing (unintended) behaviour, OK with punting on it.

ksp_plugin/interface_body.hpp
pleroy: As discussed, no.

PR #1846, opened 2018-06-09
Physical characteristics of Trappist bodies and KSP configuration files

pleroy:

PR #2024, opened 2018-12-09
Actually implement IAU 29 resolution B3

eggrobin:
While #1846 introduced a reference to the IAU resolution, it kept the existing constants.

These are ancient, introduced by https://github.com/mockingbirdnest/Principia/commit/3084f819b5bd3a88c2ddbbf0fb37ccddd846b620 in March 2014, predating the relevant IAU resolution by more than a year; being masses, they are fundamentally unsuitable as nominal values, as gravitational parameters are known with much greater accuracy than 𝐺 is.

Switch to the system of nominal solar and planetary conversion constants.

### DIFF
--- a/astronomy/trappist_gravity_model.proto.txt
+++ b/astronomy/trappist_gravity_model.proto.txt
@@ -1,5 +1,5 @@
 gravity_model {
-  solar_system_frame : SKY
+  solar_system_frame    : SKY
   # Right ascensions from a normal distribution with μ=-90, σ=0.05.
   # Declinations from a normal distribution with μ=0, σ=0.05.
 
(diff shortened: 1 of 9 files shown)


astronomy/trappist_gravity_model.proto.txt
```diff
@@ -10,7 +10,7 @@ gravity_model {
   # for life?", Vida et al., 2017.
   body {
     name                 : "Trappist-1"
-    mass                 : "0.0890 M☉"
+    mass                 : "0.0890 GM☉"
```
eggrobin: Should this be `GM☉`, `(GM)☉`, `(GM)ᴺ☉`, `(𝒢ℳ)ᴺ☉`?

pleroy: It's fine the way you wrote it.

pleroy: But it should be `gravitational_parameter`, not `mass`.  How does it even parse?

quantities/astronomy.hpp
```diff
@@ -15,25 +15,42 @@ namespace astronomy {
...
+// understood that nominal terrestrial [or jovian] radius refers specifically to
+// [the nominal equatorial radius], following common usage.”
+constexpr Length TerrestrialEquatorialRadius = 6.3781e6 * si::Metre;
+constexpr Length TerrestrialPolarRadius      = 6.3568e6 * si::Metre;
+constexpr Length JovianEquatorialRadius      = 7.1492e7 * si::Metre;
+constexpr Length JovianPolarRadius           = 6.6854e7 * si::Metre;
+constexpr GravitationalParameter TerrestrialGravitationalParameter =
+    3.986'004e14 * (Pow<3>(si::Metre) / Pow<2>(si::Second));
+constexpr GravitationalParameter JovianGravitationalParameter      =
+    1.266'865'3e17 * (Pow<3>(si::Metre) / Pow<2>(si::Second));
```
pleroy: I would have grouped all the parameters for a planet together: equatorial radius/polar radius/gravitational parameter separated by a blank line.  I would also have alphabetized the planets (thus Jovian < Terrestrial).

eggrobin: No, let's keep the order used by the IAU tables.

astronomy/trappist_gravity_model.proto.txt
```diff
@@ -1,117 +1,120 @@
...
-    name                 : "Trappist-1"
-    mass                 : "0.0890 M☉"
-    reference_instant    : "JD2457000.000000000"
-    mean_radius          : "0.1210 R☉"
-    axis_right_ascension : "-89.9408 deg"
-    axis_declination     : "0.0497342 deg"
-    reference_angle      : "0 deg"
-    angular_frequency    : "109.256 deg / d"
+    name                    : "Trappist-1"
+    gravitational_parameter : "0.0890 GM☉"
```
eggrobin: Note that this pull request does *not* regenerate the `cfg` files, as the tools would copy over `0.0890 GM☉` and the like instead of converting to SI units, which would cause interchange issues for strings passed in structs.

We want to rerun the optimization soon anyway.

PR #2004, opened 2018-12-04
InBodyParametersMarshaler

eggrobin:

PR #2026, opened 2018-12-14
[Request for early feedback] New GUI

madman2003:
As mentioned earlier, the existing principia GUI is unreadable for me, requiring me to put my nose within 20 cm of the screen. Which I obviously won't do. To resolve this practical challenge to using principia, I have re-implemented the GUI in a scaling aware system. Which at the same time brings font quality and overall appearance on par with KSP stock interface.

I've reached a point where the GUI is functional, or at least gives that appearance to someone like myself. I still have to do some final sanity checks on both the code and the GUI functionality, but as this is not a 1:1 re-implementation of the old GUI i'd like some feedback if the user experience is OK for the both of you (@pleroy and @eggrobin).

Reference frame selection is implemented in a way that looks different, because creating a replica was very impractical. The differential sliders were equally impractical or even impossible to exactly recreate due to the event driven nature of the slider that is available, so a button system replaces it.

Most other changes are restructuring to have a cleaner workflow, and fewer windows, and not waste so much space.

Currently you will see two buttons in the application, one for settings and one for the planner. They currently still share the same artwork.


eggrobin:
In light of this pull request and your previous ones (#2012, #2005), I think some things need to be clarified.

Reviewing code to the standard of quality that we maintain in this repository is actual work; even between @pleroy and I, who talk to each other on a daily basis, you will see that many pull requests involve repeated back-and-forth in the comments.

You will also see that the pull requests tend to be small changes, and what you cannot see is that behind the scene, a lot design work goes on so that we know where we are going.

While I appreciate that you went through the effort of adding a design rationale in the documentation directory, the design was not developed in concertation with us; the pull request is a behemoth at upward of two thousand lines, full of gratuitous changes (including, again, to the build system), in a style completely different from the existing code; it even comes with its own idiosyncratic copyright notices.

It would take us weeks of full-time work to get this in a mergeable state; we work on this project outside of our day jobs, for fun; we are not interested in spending our free time reviewing something that does not even attempt to conform to the surrounding code style.

While you have found the occasional of actionable issues while working on your changes (for instance #2004, or the bug in our abseil fork), we have to ask you to stop sending these changes; at this point these are a waste of both our time and yours.

Perhaps at some point we will rework the UI; it may well be that DialogGUI is the right framework to use for such a rewrite; but an extensive redesign must be planned and discussed before it is written; a pull request is not a design review.

PR #2028, opened 2018-12-15
An encoder for base 64

pleroy:


base/base64.hpp
```diff
@@ -0,0 +1,48 @@
...
+ public:
+  // Encodes |input| into |output|, which must be large enough to hold the
+  // encoded form.
+  inline void Encode(Array<std::uint8_t const> input,
+                     Array<char> output) override;
+
+  // Same as above but the storage is allocated by the callee.
+  inline UniqueArray<char> Encode(Array<std::uint8_t const> input) override;
+
+  // Length of the encoded form, in char16_t.
```
eggrobin: Do not add comments that are redundant with the base class, let alone contradictory (`char16_t`?).

base/base64_body.hpp
```diff
@@ -0,0 +1,80 @@
...
+namespace internal_base64 {
+
+constexpr std::int64_t bits_per_byte = 8;
+constexpr std::int64_t bits_per_char = 6;
+
+template<bool null_terminated>
+void principia::base::internal_base64::Base64Encoder<null_terminated>::Encode(
+    Array<std::uint8_t const> input,
+    Array<char> output) {
+  std::string_view const input_view(reinterpret_cast<const char*>(input.data),
```
eggrobin: Do we have string_view on mac?

base/base64_body.hpp
```diff
@@ -0,0 +1,80 @@
...
+                                    input.size);
+  std::string output_string;
+  absl::WebSafeBase64Escape(input_view, &output_string);
+  std::memcpy(output.data, output_string.c_str(), output_string.size());
+  if constexpr (null_terminated) {
+    output.data[output_size.size()] = 0;
```
eggrobin: Just memcpy one more byte, `std::string` has null-terminated storage.

base/base32768.hpp
```diff
@@ -15,19 +15,19 @@ namespace internal_base32768 {
 template<bool null_terminated>
 class Base32768Encoder : public Encoder<char16_t, null_terminated> {
  public:
-  void Encode(Array<std::uint8_t const> input,
-              Array<char16_t> output) override;
+  inline void Encode(Array<std::uint8_t const> input,
```
eggrobin: Redundant inline, this is templatized.

Same elsewhere.

base/base64_body.hpp
pleroy: Non lo so.  There is an `absl::string_view`.  We'll switch to that if Linux/macOS doesn't have the `std` version.

eggrobin: We can always summon the most boring kind of UB by
```C++
#if OS_MACOSX
namespace std {
using absl::string_view;
}  // namespace std
#endif
```

PR #2031, opened 2018-12-20
Reimplement the Bundle to not look like a thread pool

pleroy:
And to reduce contention.

### DIFF
--- a/base/bundle.hpp
+++ b/base/bundle.hpp
@@ -2,89 +2,58 @@
 #pragma once
 
 #include <functional>
-#include <queue>
-#include <memory>
+#include <list>
 #include <optional>
 #include <thread>
 #include <vector>
 
 #include "absl/base/thread_annotations.h"
+#include "absl/synchronization/notification.h"
 #include "absl/synchronization/mutex.h"
-#include "base/macros.hpp"
-#include "base/monostable.hpp"
-#include "base/not_null.hpp"
 #include "base/status.hpp"
 
 namespace principia {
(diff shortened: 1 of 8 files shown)


base/bundle.hpp
```diff
@@ -2,89 +2,58 @@
...
 
-// Functions that can be cooperatively aborted should check |AbortRequested()|
-// when appropriate.  The function is thread-safe.
-extern thread_local std::function<bool()> AbortRequested;
-
-// A thread pool with cooperative aborting (but no cooperative scheduling).
-// We refer to the thread on which the |Bundle| is created as its master thread.
-// The |Bundle| itself cooperatively aborts if |AbortRequested()| on its master
-// thread.
+// TODO(phl): comment
```
eggrobin: Indeed.

base/bundle.hpp
```diff
@@ -2,89 +2,58 @@
...
-  // Whether the workers should terminate when no tasks are available.  Set by
-  // |Join|.
-  Monostable wait_on_empty_ GUARDED_BY(lock_);
+  absl::Mutex lock_;
 
-  std::queue<Task> tasks_ GUARDED_BY(lock_);
-  std::vector<std::thread> workers_ GUARDED_BY(lock_);
+  // Whether |Join| has been called.  When set to true, |Add| should not be
+  // called.
+  bool joining_ GUARDED_BY(lock_) = false;
```
eggrobin: Could this be atomic as well?

PR #2033, opened 2018-12-21
A mechanism for asynchronous deletion of objects

pleroy:


base/graveyard.hpp
```diff
@@ -0,0 +1,26 @@
...
+#include <cstdint>
+#include <memory>
+
+#include "base/thread_pool.hpp"
+
+namespace principia {
+namespace base {
+
+template<typename T>
+class Graveyard {
```
eggrobin: Templatize `Bury`, not the class, that way a single `Graveyard` can be used for different types.

base/graveyard_body.hpp
```diff
@@ -0,0 +1,21 @@
...
+namespace principia {
+namespace base {
+
+template<typename T>
+Graveyard<T>::Graveyard(std::int64_t const number_of_threads)
+    : gravedigger_(number_of_threads) {}
+
+template<typename T>
+void Graveyard<T>::Bury(std::unique_ptr<T> t) {
+  gravedigger_.Add([coffin = t.release()]() {
```
eggrobin: Add a TODO so that we remember to investigate the possibility of a mutable λ with `std::packaged_task` in the `ThreadPool` instead of `std::function`.

base/graveyard.hpp
```diff
@@ -0,0 +1,26 @@
...
+ public:
+  explicit Graveyard(std::int64_t number_of_threads);
+
+  void Bury(std::unique_ptr<T> t);
```
eggrobin: Should `t` be called `hearse`?

base/graveyard.hpp
pleroy: Done.

base/graveyard.hpp
pleroy: Nope.

base/graveyard_body.hpp
pleroy: Done.

PR #2032, opened 2018-12-20
Selenopotential to degree and order 50

eggrobin:
Bump `max_geopotential_degree` and extend the numerical tables accordingly.

PR #2034, opened 2018-12-22
Remove the implicit constructor for AccuracyParameters

eggrobin:
Use `0x1p-24` in most places; this should make the tests (and benchmarks, etc.) a bit faster, and most importantly prevent them from becoming massively slower with #2032.

### DIFF
--- a/benchmarks/ephemeris.cpp
+++ b/benchmarks/ephemeris.cpp
@@ -123,7 +123,8 @@ void BM_EphemerisKSPSystem(benchmark::State& state) {
     astronomy::StabilizeKSP(*at_origin);
     Instant const final_time = at_origin->epoch() + 100 * JulianYear;
     auto const ephemeris = at_origin->MakeEphemeris(
-        FittingTolerance(state.range(0)),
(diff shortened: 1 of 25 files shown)


benchmarks/ephemeris.cpp
```diff
@@ -123,7 +123,8 @@ void BM_EphemerisKSPSystem(benchmark::State& state) {
     astronomy::StabilizeKSP(*at_origin);
     Instant const final_time = at_origin->epoch() + 100 * JulianYear;
     auto const ephemeris = at_origin->MakeEphemeris(
-        FittingTolerance(state.range(0)),
+        /*accuracy_parameters=*/{FittingTolerance(state.range(0)),
```
pleroy: Replace the function `FittingTolerance` by a function `MakeAccuracyParameters` that sets both parameters.

eggrobin: As discussed, this doesn't really work well because of the existing `MakeAccuracyParameters` in this file that would still need `FittingTolerance`.

eggrobin:
retest this please

PR #2042, opened 2018-12-28
An allocator based on arenas

pleroy:


base/arena_allocator_test.cpp
```diff
@@ -0,0 +1,28 @@
...
+  }
+
+  google::protobuf::Arena arena_;
+  ArenaAllocator<std::string> allocator_;
+};
+
+TEST_F(ArenaAllocatorTest, Container) {
+  CHECK_EQ(0, arena_.SpaceUsed());
+  std::vector<std::string, ArenaAllocator<std::string>> v(1000, allocator_);
+  CHECK_LT(0, arena_.SpaceUsed());
```
eggrobin: `EXPECT_THAT(arena_.SpaceUsed(), Gt(1000));`, this is not `std::vector<bool>`, at least one byte should have been allocated per object.

Consider also trying an `std::vector<std::basic_string<char, ArenaAllocator<char>>, ArenaAllocator<std::basic_string<char, ArenaAllocator<char>>>>` and checking that it allocates more.

base/arena_allocator_test.cpp
```diff
@@ -0,0 +1,28 @@
...
+  ArenaAllocatorTest() : allocator_(&arena_) {
+    arena_.Init(google::protobuf::ArenaOptions());
+  }
+
+  google::protobuf::Arena arena_;
+  ArenaAllocator<std::string> allocator_;
+};
+
+TEST_F(ArenaAllocatorTest, Container) {
+  CHECK_EQ(0, arena_.SpaceUsed());
```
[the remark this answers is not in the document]
pleroy: Done.

base/arena_allocator_test.cpp
pleroy: Done.

PR #2047, opened 2018-12-31
Trappist optimization results

pleroy:


astronomy/trappist_dynamics_test.cpp
```diff
@@ -1176,29 +1180,49 @@ TEST_F(TrappistDynamicsTest, MathematicaPeriods) {
 }
 
 TEST_F(TrappistDynamicsTest, MathematicaTransits) {
-  Instant const a_century_later = system_.epoch() + 100 * JulianYear;
-  ephemeris_->Prolong(a_century_later);
-
-  TransitsByPlanet computations;
+  // Run this test with different ephemeris to make sure that the system is
+  // converged.
```
eggrobin: Should this be a TEST_P instead?

astronomy/trappist_dynamics_test.cpp
```diff
@@ -1176,29 +1180,49 @@ TEST_F(TrappistDynamicsTest, MathematicaPeriods) {
 }
 
 TEST_F(TrappistDynamicsTest, MathematicaTransits) {
-  Instant const a_century_later = system_.epoch() + 100 * JulianYear;
-  ephemeris_->Prolong(a_century_later);
-
-  TransitsByPlanet computations;
+  // Run this test with different ephemeris to make sure that the system is
```
eggrobin: different ephemerides, or different ephemeris parameters (since the ephemerides share the same system, only the parameters change).

astronomy/trappist_dynamics_test.cpp
```diff
@@ -1176,29 +1180,49 @@ TEST_F(TrappistDynamicsTest, MathematicaPeriods) {
...
-  double const χ² = Transitsχ²(observations, computations, info);
-#if 0
-  CHECK_LT(χ², 482.0);
-#endif
-  CHECK_GT(χ², 470.0);
-  LOG(ERROR) << u8"χ²: " << χ² << " " << info;
+    std::string info;
+    double const χ² = Transitsχ²(observations, computations, info);
+    CHECK_LT(χ², 359.0);
+    CHECK_GT(χ², 358.0);
```
eggrobin: Why are these checks and not expectations?

tools/generate_configuration.cpp
```diff
@@ -105,8 +105,25 @@ void GenerateConfiguration(std::string const& game_epoch,
...
+      if (body.gravitational_parameter().find_first_of(u8"☉🜨") ==
+          std::string::npos) {
+        gravity_model_cfg << "    gravitational_parameter = "
+                          << body.gravitational_parameter() << "\n";
+      } else {
+        GravitationalParameter const gravitational_parameter =
+            ParseQuantity<GravitationalParameter>(
+                body.gravitational_parameter());
+        gravity_model_cfg << "    gravitational_parameter = "
+                          << std::scientific
```
eggrobin: `DebugString(gravitational_parameter)` (in SI units, rather than multiples of SI units) would avoid losing an ULP from the division by 1 000 000 000.

astronomy/trappist_dynamics_test.cpp
pleroy: There would be a lot of boilerplate, so, no.

astronomy/trappist_dynamics_test.cpp
pleroy: Done.

astronomy/trappist_dynamics_test.cpp
pleroy: Done.

tools/generate_configuration.cpp
pleroy: We use `km^3/s^2` everywhere, now is not the time to change this.

astronomy/trappist_dynamics_test.cpp
```diff
@@ -1218,8 +1218,8 @@ TEST_F(TrappistDynamicsTest, MathematicaTransits) {
 
     std::string info;
     double const χ² = Transitsχ²(observations, computations, info);
-    CHECK_LT(χ², 359.0);
-    CHECK_GT(χ², 358.0);
+    EXPECT_LT(χ², 359.0);
+    EXPECT_GT(χ², 358.0);
```
eggrobin: `EXPECT_THAT(χ², AllOf(Gt(358.0), Lt(359.0)));`

PR #2051, opened 2019-01-04
Limit the geopotential degree to 30 by default

eggrobin:
We should figure out a cleaner way to do this; as it stands the degree 50 path is never compiled (and this makes the RGT test a bit weird, since its goal is to compare truncations with the degree 50 replication of the original paper).


physics/solar_system_body.hpp
```diff
@@ -107,6 +107,8 @@ SolarSystem<Frame>::SolarSystem(
     CheckFrame(initial_state_);
   }
 
+  for (auto& body : *gravity_model_.mutable_body())
```
pleroy: Remove.

Issue #2056, opened 2019-01-05
Crash when craft breaks in atmosphere with explosions in KSP 1.3.1 with RP-1 (Erdos)


nikolain:
When craft starts breaking apart and explode on atmospheric reentry, there is a chance for Principia Erdos to crash. This happens with <10% of such reentries. 

Native call stack:
```
 	ucrtbase.dll!abort()	Unknown
 	libglog.dll!google::Symbolize() + 577 bytes	Unknown
 	libglog.dll!google::LogMessage::SendToLog() + 829 bytes	Unknown
 	libglog.dll!google::LogMessage::Flush() + 213 bytes	Unknown
 	libglog.dll!google::LogMessageFatal::~LogMessageFatal() + 135 bytes	Unknown
 	principia.dll!principia::ksp_plugin::internal_pile_up::PileUp::AdvanceTime(const principia::geometry::internal_point::Point<principia::quantities::internal_quantities::Quantity<principia::quantities::internal_dimensions::Dimensions<0,0,1,0,0,0,0,0> > > & t) Line 372	C++
 	principia.dll!principia::ksp_plugin::internal_pile_up::PileUp::DeformAndAdvanceTime(const principia::geometry::internal_point::Point<principia::quantities::internal_quantities::Quantity<principia::quantities::internal_dimensions::Dimensions<0,0,1,0,0,0,0,0> > > & t) Line 131	C++
 	[Inline Frame] principia.dll!principia::ksp_plugin::internal_plugin::Plugin::CatchUpLaggingVessels::__l6::<lambda_181fb935bfd596ed15d2291f87f03453>::operator()() Line 719	C++
 	[Inline Frame] principia.dll!std::_Invoker_functor::_Call(principia::ksp_plugin::internal_plugin::Plugin::CatchUpLaggingVessels::__l6::<lambda_181fb935bfd596ed15d2291f87f03453> &)	C++
 	[Inline Frame] principia.dll!std::invoke(principia::ksp_plugin::internal_plugin::Plugin::CatchUpLaggingVessels::__l6::<lambda_181fb935bfd596ed15d2291f87f03453> &)	C++
 	[Inline Frame] principia.dll!std::_Invoker_ret<principia::base::Status,0>::_Call(principia::ksp_plugin::internal_plugin::Plugin::CatchUpLaggingVessels::__l6::<lambda_181fb935bfd596ed15d2291f87f03453> &)	C++
[...]

PR #2064, opened 2019-01-13
Propagate errors found during integration to kill the vessel

pleroy:
This may help with #2056 but I don't have an easy way to test.  There will be a message in the WARNING log to see if we are properly handling the condition.

### DIFF
--- a/ksp_plugin/plugin.cpp
+++ b/ksp_plugin/plugin.cpp
@@ -769,14 +769,14 @@ void Plugin::WaitForVesselToCatchUp(PileUpFuture& pile_up_future,
   PileUp const* const pile_up = pile_up_future.pile_up;
   auto& future = pile_up_future.future;
   future.wait();
-  if (!future.get().ok()) {
+  Status const status = future.get();
+  if (!status.ok()) {
     for (not_null<Part*> const part : pile_up->parts()) {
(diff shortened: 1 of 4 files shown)


ksp_plugin/plugin.cpp
```diff
@@ -769,13 +769,14 @@ void Plugin::WaitForVesselToCatchUp(PileUpFuture& pile_up_future,
...
+  if (!future.get().ok()) {
     for (not_null<Part*> const part : pile_up->parts()) {
       not_null<Vessel*> const vessel =
           FindOrDie(part_id_to_vessel_, part->part_id());
       if (collided_vessels.insert(vessel).second) {
-        LOG(INFO) << "Vessel " << vessel->ShortDebugString()
-                  << " collided with a celestial";
+        LOG(WARNING) << "Vessel " << vessel->ShortDebugString()
+                     << " collided with a celestial: "
+                     << future.get().ToString();
```
eggrobin: We cannot call `future.get()` twice, see https://en.cppreference.com/w/cpp/thread/future/get:
> The behavior is undefined if `valid()` is `false` before the call to this function.
Any shared state is released. `valid()` is `false` after a call to this method.

pleroy:
retest this please

PR #2079, opened 2019-02-17
Geopotential API

### LINKED ISSUE #2074: An API to get geopotential coefficients for analytic perturbation theories
One day @lamont-granquist will implement primer vector guidance based to the Principia flight plan, integrated by Principia (see https://github.com/mockingbirdnest/Principia/issues/1659#issuecomment-430159816).

In the meantime however, there is much to do in low-Earth orbit (in particular for launch, where the Principia flight plan does not exist and third-body effects are irrelevant).

There, the usual analytic perturbation models are adequate; launching to given 𝐽<sub>2</sub>-mean elements would be vastly more useful than having the guidance pretend that the osculating elements are true.

In order to use these, the geopotential coefficients must be obtained from Principia.

An API along the lines of
```C#
Status ExternalGetGeopotentialCoefficient(int body_index,
                                          int degree,
                                          int order,
                                          out XY coefficient);
```
should do the job (where the `XY` contains `x` = 𝐶<sub>𝑛𝑚</sub>, `y` = 𝑆<sub>𝑛𝑚</sub>).

eggrobin:
Fix #2074.

### DIFF
--- a/serialization/journal.proto
+++ b/serialization/journal.proto
@@ -423,38 +423,12 @@ message ExternalFlowFreefall {
   optional Return return = 3;
 }
 
-// Returns the first point of the coast phase following the given manoeuvre
-// which is locally nearest to the reference position, or the nearest endpoint
-// if there is no local minimum.
-message ExternalGetNearestPlannedCoastDegreesOfFreedom {
-  extend Method {
-    optional ExternalGetNearestPlannedCoastDegreesOfFreedom extension = 5152;
-  }
-  message In {
-    required fixed64 plugin = 1 [(pointer_to) = "Plugin const",
-                                 (is_subject) = true];
-    required int32 central_body_index = 2;
-    required string vessel_guid = 3;
-    required int32 manoeuvre_index = 4;
-    required XYZ world_body_centred_reference_position = 5;
-  }
-  message Out {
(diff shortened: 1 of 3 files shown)


serialization/journal.proto
```diff
@@ -449,6 +449,55 @@ message ExternalGetNearestPlannedCoastDegreesOfFreedom {
...
+  }
+  message Return {
+    required Status result = 1;
+  }
+  optional In in = 1;
+  optional Out out = 2;
+  optional Return return = 3;
+}
+
+// Sets |reference_radius| to the reference radius of the geopotential model for
```
pleroy: In metres.

serialization/journal.proto
```diff
@@ -423,23 +423,22 @@ message ExternalFlowFreefall {
   optional Return return = 3;
 }
 
-// Returns the first point of the coast phase following the given manoeuvre
-// which is locally nearest to the reference position, or the nearest endpoint
-// if there is no local minimum.
-message ExternalGetNearestPlannedCoastDegreesOfFreedom {
+// Sets |coefficient| to the geopotential coefficient of the given |degree| and
```
pleroy: normalized coefficient

PR #2084, opened 2019-02-19
Make ContinuousTrajectory thread-safe

pleroy:
Preparatory work for asynchronous predictions (aka prognostications).


eggrobin:
Did you run this through clang thread safety analysis?

PR #2087, opened 2019-03-08
Better semantics for AttachFork

pleroy:


physics/discrete_trajectory.hpp
```diff
@@ -108,9 +108,9 @@ class DiscreteTrajectory : public Forkable<DiscreteTrajectory<Frame>,
   // trajectory.
   not_null<DiscreteTrajectory<Frame>*> NewForkAtLast();
 
-  // The first point of |fork| is removed from |fork| and appended (using
-  // Append) to this trajectory.  Then |fork| is made a fork of this trajectory
-  // at the newly-inserted point.  |fork| must be a non-empty root.
+  // Changes |fork| to become a fork of this trajectory at the end of this
+  // trajectory.  |fork| must be a non-empty root.  It may or may not have a
+  // point at the same time as the end of this trajectory.
```
eggrobin: Clarify the requirements on `fork` that were implicit in `using Append`, e.g., it must not have points before the end of the trajectory, etc.

Clarify the behaviour with respect to the possible common point: which one is kept, etc.

physics/discrete_trajectory_body.hpp
```diff
@@ -108,19 +108,41 @@ template<typename Frame>
 void DiscreteTrajectory<Frame>::AttachFork(
     not_null<std::unique_ptr<DiscreteTrajectory<Frame>>> fork) {
   CHECK(fork->is_root());
-  CHECK(!fork->timeline_.empty());
+  CHECK(!Empty());
```
eggrobin: This breaks the flight plan.
https://github.com/mockingbirdnest/Principia/blob/d682ca43aa569ef475b0a8e96c7c5315d4b9da81/ksp_plugin/flight_plan.cpp#L115-L121

physics/discrete_trajectory.hpp
pleroy: Done.

PR #2088, opened 2019-03-08
Make the Ephemeris fully thread-safe and improve error reporting

pleroy:
Also take advantage of the fact that ContinuousTrajectory is now thread-safe.

### DIFF
--- a/physics/ephemeris.hpp
+++ b/physics/ephemeris.hpp
@@ -29,6 +29,7 @@ namespace principia {
 namespace physics {
 namespace internal_ephemeris {
 
+using base::Error;
 using base::not_null;
 using base::Status;
 using geometry::Instant;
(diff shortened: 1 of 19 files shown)


physics/ephemeris.hpp
```diff
@@ -171,20 +172,20 @@ class Ephemeris {
...
 
   // Returns true if at least one of the trajectories is empty.
-  virtual bool empty() const EXCLUDES(lock_);
+  virtual bool empty() const;
 
   // The maximum of the |t_min|s of the trajectories.
   virtual Instant t_min() const EXCLUDES(lock_);
   // The mimimum of the |t_max|s of the trajectories.
-  virtual Instant t_max() const EXCLUDES(lock_);
+  virtual Instant t_max() const;
```
eggrobin: Comment: the values of `t_min()` and `t_max()` are only affected by calls to `Prolong` and `ForgetBefore`; while `lock_` is held, their value cannot be changed by another thread.

In the absence of calls to `ForgetBefore`, `t_max()` can only increase, and `t_min()` can only decrease.

This effectively makes `ForgetBefore` hostile to operating on a trajectory from another thread.

Perhaps we should get rid of it; it makes things very messy.

PR #2089, opened 2019-03-08
An interface that can actually be called

eggrobin:
The `ExternalInterface` abstracts away the fetching and passing of the plugin, and converts statuses into more idiomatic exceptions; it turns out parameters into return values, since the former are annoying to use by reflection.
By wrapping the `Interface.ExternalMeow` calls, it works around the calling convention issues described below.

See the draft documentation at https://github.com/mockingbirdnest/Principia/wiki/Interface-for-other-KSP-mods.

The API added in #2079 for #2074 is cumbersome to call directly (the caller has to look for a `PrincipiaPluginAdapter`, fetch its private `plugin_` field, and check that the private static field `Loader.loaded_principia_dll_` is true), and calls to functions with parameters fail hard anyway, because mono somehow messes up the calling convention when calling a P/Invoke function directly by reflection: an attempt to call

 ```C#
  internal static extern Status ExternalGeopotentialGetCoefficient(
      this IntPtr plugin,
      int body_index,
      int degree,
      int order,
      out XY coefficient);
```
```C++
extern "C" PRINCIPIA_DLL
Status CDECL principia__ExternalGeopotentialGetCoefficient(
    Plugin const* const plugin,
    int const body_index,
    int const degree,
    int const order,
    XY* const coefficient);
``` 
with
```
plugin = 0x5AFEA730,
body_index = 1,
degree = 2,
order = 0
```
led to, on the C++ side,
```
plugin = 1,
body_index = 2,
order = 0x3F3B58D0
coefficient = 18
```
and a subsequent access violation when dereferencing `1`.


ksp_plugin_adapter/external_interface.cs
```diff
@@ -0,0 +1,77 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Security.Authentication;
+
+namespace principia {
+namespace ksp_plugin_adapter {
+
+public class ExternalInterface {
+  public XY GeopotentialGetCoefficient(
+      int body_index,
```
pleroy: Would this fit on the previous line?

ksp_plugin_adapter/external_interface.cs
```diff
@@ -0,0 +1,77 @@
...
+      int degree,
+      int order) {
+    ThrowOnError(
+        adapter_.Plugin().ExternalGeopotentialGetCoefficient(
+            body_index, degree, order, out XY result));
+    return result;
+  }
+
+  public double GeopotentialGetReferenceRadius(
+      int body_index) {
```
pleroy: Same.

ksp_plugin_adapter/ksp_plugin_adapter.cs
```diff
@@ -88,7 +88,10 @@ private enum UnityLayers {
   private bool selecting_active_vessel_target_ = false;
   private bool selecting_target_celestial_ = false;
 
-  private IntPtr plugin_ = IntPtr.Zero;
+  private IntPtr plugin_;
```
pleroy: Make the default value explicit.

tools/journal_proto_processor.cpp
```diff
@@ -947,9 +947,15 @@ void JournalProtoProcessor::ProcessInterchangeMessage(
     cs_interface_type_declaration_[descriptor] =
         "internal partial class " + name + " {\n";
   } else {
-    cs_interface_type_declaration_[descriptor] =
-        "[StructLayout(LayoutKind.Sequential)]\ninternal partial struct " +
-        name + " {\n";
+  MessageOptions const& message_options = descriptor->options();
```
pleroy: Bad, bad indent!

tools/journal_proto_processor.cpp
```diff
@@ -947,9 +947,15 @@ void JournalProtoProcessor::ProcessInterchangeMessage(
...
+  std::string const visibility =
+      message_options.GetExtension(journal::serialization::is_public)
```
pleroy: This makes it possible to write `(is_public) = false`, which is something we try to avoid.  Look how things are done for `is_consumed` around line 365.

pleroy:
retest this please

PR #2091, opened 2019-03-10
Remove GetTargetVesselPrediction

pleroy:
Fix a bug in the computation of closest approaches, and do some cleanup in Renderer and Plugin.


ksp_plugin/renderer.cpp
```diff
@@ -315,7 +302,7 @@ Renderer::Target::Target(
 not_null<NavigationFrame const*> Renderer::GetPlottingFrame(
     Instant const& time) const {
   if (target_) {
-    GetTargetVesselPrediction(time);
+    target_->vessel->FlowPrediction(time);
```
eggrobin: Should we not keep the check?

pleroy: Done.

PR #2095, opened 2019-03-16
Move two dialogs to a separate class

pleroy:


ksp_plugin_adapter/ksp_plugin_adapter.cs
```diff
@@ -223,29 +221,43 @@ private enum UnityLayers {
   private Dictionary<uint, QP> part_id_to_degrees_of_freedom_ =
       new Dictionary<uint, QP>();
 
-  // The first apocalyptic error message.
+  // UI for the apocalypse notification.
   [KSPField(isPersistant = true)]
   private String revelation_ = "";
-  // Whether we have encountered an apocalypse already.
   [KSPField(isPersistant = true)]
   private bool is_post_apocalyptic_ = false;
```
eggrobin: Keep the comments that document the semantics of the string and bool.

ksp_plugin_adapter/ksp_plugin_adapter.cs
```diff
@@ -223,29 +221,43 @@ private enum UnityLayers {
...
+  // UI for the bad installation notification.
+  [KSPField(isPersistant = true)]
+  private String bad_installation_message_ = "";
+  [KSPField(isPersistant = true)]
+  private bool is_bad_installation_ = false;
+  [KSPField(isPersistant = true)]
+  private int bad_installation_dialog_x_ = Dialog.XCentre;
+  [KSPField(isPersistant = true)]
+  private int bad_installation_dialog_y_ = Dialog.YCentre;
+  private Dialog bad_installation_dialog_ = new Dialog();
```
eggrobin: We should have utilities in `Dialog` to save/load the coordinates from a ConfigNode, which we would call from OnLoad and OnSave here, I don't think the `KSPField` attribute is improving readability at this point.

OK to do this in another CL.

ksp_plugin_adapter/dialog.cs
```diff
@@ -0,0 +1,51 @@
...
+    WindowUtilities.EnsureOnScreen(ref dialog_window_rectangle);
+    x = (int)dialog_window_rectangle.xMin;
+    y = (int)dialog_window_rectangle.yMin;
+  }
+
+  public static int XCentre {
+    get {
+      return UnityEngine.Screen.width / 2;
+    }
+  }
```
eggrobin: 1. Why is this a property and not a field?
2. If it is a property, it could at least be [auto-implemented](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/auto-implemented-properties).
```C#
public static int XCentre { get; } = UnityEngine.Screen.width / 2;
```

ksp_plugin_adapter/ksp_plugin_adapter.cs
```diff
@@ -223,29 +221,43 @@ private enum UnityLayers {
...
-  private int apocalypse_window_y_ = UnityEngine.Screen.height / 3;
-  private UnityEngine.Rect apocalypse_window_rectangle_;
+  private int apocalypse_dialog_y_ = Dialog.YCentre;
+  private Dialog apocalypse_dialog_ = new Dialog();
+
+  // UI for the bad installation notification.
+  [KSPField(isPersistant = true)]
+  private String bad_installation_message_ = "";
+  [KSPField(isPersistant = true)]
+  private bool is_bad_installation_ = false;
```
eggrobin: Why would you want to persist this bit?
If the user is missing the redistributable framework, sees the message, installs it, restarts KSP (now with a good Principia installation), and reloads the save, the now-outdated error message will be shown; the save will forever be doomed, whereas there never was a problem with the save itself.

ksp_plugin_adapter/ksp_plugin_adapter.cs
pleroy: Added a TODO.

ksp_plugin_adapter/dialog.cs
pleroy: Done.

ksp_plugin_adapter/ksp_plugin_adapter.cs
pleroy: Changed to not persist the message and the bit.

ksp_plugin_adapter/ksp_plugin_adapter.cs

PR #2096, opened 2019-03-17
Change the Dialog class to persist its data using a ConfigNode

pleroy:
Also centre the dialog by default.

### DIFF
--- a/ksp_plugin_adapter/ksp_plugin_adapter.cs
+++ b/ksp_plugin_adapter/ksp_plugin_adapter.cs
@@ -229,18 +229,12 @@ public partial class PrincipiaPluginAdapter
   [KSPField(isPersistant = true)]
   private bool is_post_apocalyptic_ = false;
   [KSPField(isPersistant = true)]
-  private int apocalypse_dialog_x_ = Dialog.XCentre;
-  [KSPField(isPersistant = true)]
-  private int apocalypse_dialog_y_ = Dialog.YCentre;
   private Dialog apocalypse_dialog_ = new Dialog();
 
   // UI for the bad installation notification.
   private String bad_installation_message_ = "";
   private bool is_bad_installation_ = false;
   [KSPField(isPersistant = true)]
-  private int bad_installation_dialog_x_ = Dialog.XCentre;
-  [KSPField(isPersistant = true)]
-  private int bad_installation_dialog_y_ = Dialog.YCentre;
   private Dialog bad_installation_dialog_ = new Dialog();
 
   public event Action render_windows;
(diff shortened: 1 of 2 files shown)


pleroy:
retest this please

ksp_plugin_adapter/ksp_plugin_adapter.cs
```diff
@@ -222,25 +222,15 @@ private enum UnityLayers {
...
 
   // UI for the bad installation notification.
-  private String bad_installation_message_ = "";
-  private bool is_bad_installation_ = false;
+  private bool is_bad_installation_ = false;  // Don't persist.
   [KSPField(isPersistant = true)]
-  private int bad_installation_dialog_x_ = Dialog.XCentre;
-  [KSPField(isPersistant = true)]
-  private int bad_installation_dialog_y_ = Dialog.YCentre;
   private Dialog bad_installation_dialog_ = new Dialog();
```
eggrobin: As discussed, the following are unsightly:
- a non-`null` invalid `Dialog`;
- a `bool` field equivalent to the `null`ness of another field.

Make the `Dialog`s `null` by default, remove the `is_post_apocalyptic_` field, create the `Dialog`s when the `bool`s would be set.
Take a message at construction in `Dialog`, so that there is no need for the `"SHOW WITHOUT MESSAGE"` thing.

For abstraction and readability, replace the `is_post_apocalyptic_` field with a property:
```C#
private bool is_post_apocalyptic { get { return apocalypse_dialog_ != null; } }
```
`is_bad_installation_` remains inequivalent to `bad_installation_dialog != null`; it can either become an auto-implemented property or remain a field.

```C#
if (is_post_apocalyptic_) {
  apocalypse_dialog_.Show();
}
```
becomes
```C#
apocalypse_dialog_?.Show();
```

pleroy: Done, except that `is_post_apocalyptic_` is really not needed anymore.

ksp_plugin_adapter/ksp_plugin_adapter.cs
```diff
@@ -1384,9 +1363,11 @@ private enum UnityLayers {
       time_is_advancing_ = time_is_advancing(universal_time);
       if (time_is_advancing_) {
         plugin_.AdvanceTime(universal_time, Planetarium.InverseRotAngle);
-        if (!is_post_apocalyptic_) {
-          is_post_apocalyptic_ |=
-              plugin_.HasEncounteredApocalypse(out revelation_);
+        if (apocalypse_dialog_ == null) {
+          String revelation = "";
+          if (plugin_.HasEncounteredApocalypse(out revelation)) {
```
eggrobin:
```C#
out string revelation
```

PR #2100, opened 2019-03-18
Make Dialog a WindowRenderer

pleroy:
To do this, split "supervised" from "unsupervised" renderer.  Also move persistence to the renderer (not used yet to persist the flight plan and frame selector).


pleroy:
retest this please

pleroy:
retest this please

PR #2101, opened 2019-03-20
Get rid of the "disposable" pattern for windows

pleroy:
* Windows are now constructed with just their supervisor and later initialized with the plugin.  This makes it possible to reuse the same window when the plugin gets recreated (e.g., reloaded from a save).
* Windows are no longer using the "disposable" pattern and are no longer wrapped in a `Controlled` object.  Instead, they use the `clear_locks` and `dispose_windows` events to release the locks and unregister themselves at the right place.
* A bug is fixed whereby locks were not cleared when the main window was hidden.
* A bit of code reordering to bring related things closer together.

### DIFF
--- a/ksp_plugin_adapter/ksp_plugin_adapter.cs
+++ b/ksp_plugin_adapter/ksp_plugin_adapter.cs
@@ -236,6 +236,7 @@ public partial class PrincipiaPluginAdapter
   private FlightPlanner flight_planner_;
   private MapNodePool map_node_pool_;
 
+  public event Action dispose_windows;
   public event Action render_windows;
 
   PrincipiaPluginAdapter() {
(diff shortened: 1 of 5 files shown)


ksp_plugin_adapter/ksp_plugin_adapter.cs
```diff
@@ -264,10 +269,11 @@ private enum UnityLayers {
                 "; this build targets " + expected_version + ".");
     }
     map_node_pool_ = new MapNodePool();
-  }
-
-  ~PrincipiaPluginAdapter() {
-    Cleanup();
```
eggrobin: The `PrincipiaPluginAdapter` does have ownership of an unmanaged resource, namely the plugin. It would make sense to destroy it here if it is not IntPtr.Zero (perhaps there are cases where the normal lifecycle of the game is interrupted, and the object finalized without ever going through OnDisable).

PR #2102, opened 2019-03-21
Make the hide/show status of windows persistent

pleroy:
* Move the hide/show bit to `WindowRenderer` and save/load it.
* Add `Hide`, `Show`, `Shown` and `Toggle` to manage that bit.
* Make `RenderWindow` nonabstract to factor out logic that's common to all the windows (e.g., clearing the lock when hiding).
* Subclasses are now expected to override `Title` and a new `RenderWindow(int)` method.
* Pass GUI options at construction if needed with a reasonable default.
* Drudgery to centre a window the first time it is shown.

### DIFF
--- a/ksp_plugin_adapter/ksp_plugin_adapter.cs
+++ b/ksp_plugin_adapter/ksp_plugin_adapter.cs
@@ -218,9 +218,6 @@ public partial class PrincipiaPluginAdapter
       new Dictionary<uint, QP>();
 
   // UI for the apocalypse notification.
-  // Whether we have encountered an apocalypse already.
-  [KSPField(isPersistant = true)]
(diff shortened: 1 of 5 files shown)


ksp_plugin_adapter/ksp_plugin_adapter.cs
```diff
@@ -679,12 +675,11 @@ private enum UnityLayers {
   private void OnGUI() {
     if (is_bad_installation_) {
       bad_installation_dialog_.Show();
+      bad_installation_dialog_.RenderWindow();
```
eggrobin: Please make the `show_` bit equivalent to the window being shown.

PR #2103, opened 2019-03-21
Use constexpr std::array aggregates instead of std::initializer_list

eggrobin:
The overload meant
```C++
FixedMatrix<double, 1, 1>{{{0}}}
```
was not a constant expression, whereas
```C++
FixedMatrix<double, 2, 2>{{{1, 0
                            0, 1}}}
```
was.

With the overload gone, these are both constexpr, and can even be rewritten with fewer brackets as
```C++
FixedMatrix<double, 1, 1>{{0}}
FixedMatrix<double, 2, 2>{{1, 0
                           0, 1}}
```


pleroy:
retest this please

PR #2105, opened 2019-03-23
Implement the main window using a separate class

pleroy:


ksp_plugin_adapter/main_window.cs
```diff
@@ -0,0 +1,493 @@
...
+        AdaptiveStepParameters adaptive_step_parameters =
+            plugin_.VesselGetPredictionAdaptiveStepParameters(vessel_guid);
+        prediction_length_tolerance_index_ = Array.FindIndex(
+            prediction_length_tolerances_,
+            (double tolerance) =>
+                tolerance >=
+                    adaptive_step_parameters.length_integration_tolerance);
+        prediction_steps_index_ = Array.FindIndex(
+            prediction_steps_,
+            (Int64 step) => step >= adaptive_step_parameters.max_steps);
```
eggrobin: What happens if `Array.FindIndex` returns `-1`?

pleroy: Added handling of that case.

ksp_plugin_adapter/main_window.cs
```diff
@@ -0,0 +1,504 @@
...
+    node.SetValue("verbose_logging",
+                  verbose_logging_,
+                  createIfNotFound : true);
+
+    node.SetValue("must_record_journal",
+                  must_record_journal_,
+                  createIfNotFound : true);
+  }
+
+  protected override string Title { get; } = "Principia";
```
eggrobin: https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/properties#expression-body-definitions
```C#
protected override string Title => "Principia";
```

PR #2108, opened 2019-03-24
Map node pool improvements

eggrobin:
Since `MapNode`s are costly to `Create` (proportionally to the number of existing `MapNode`s),
- reuse `MapNode`s as much as possible:
  - make unused nodes invisible instead of terminating them,
  - do not clear the `MapNodePool` when there is no trajectory to render;
- limit the number of map nodes of one kind to 64.

Further, fix the execution order issue that caused the nodes to lag behind the trajectories when moving the camera.

### DIFF
--- a/ksp_plugin_adapter/ksp_plugin_adapter.cs
+++ b/ksp_plugin_adapter/ksp_plugin_adapter.cs
@@ -1489,8 +1489,7 @@ public partial class PrincipiaPluginAdapter
     // otherwise, the map nodes will lag behind when the camera is moved.
     // The only timing that satisfies these constraints is BetterLateThanNever
     // in LateUpdate.
-    string main_vessel_guid = (FlightGlobals.ActiveVessel ??
-                               space_tracking?.SelectedVessel)?.id.ToString();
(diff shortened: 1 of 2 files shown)


ksp_plugin_adapter/ksp_plugin_adapter.cs
```diff
@@ -1548,6 +1554,34 @@ private enum UnityLayers {
...
+  private void BetterLateThanNeverLateUpdate() {
+    // While we draw the trajectories directly (and thus do so after everything
+    // else has been rendered), we rely on the game to render its map nodes.
+    // Since the screen position is determined in |MapNode.NodeUpdate|, it must
+    // be called before rendering occurs, but after the cameras have moved;
+    // otherwise, the map nodes will lag behind when the camera is moved.
+    // The only timing that satisfies these constraints is BetterLateThanNever
+    // in LateUpdate.
+    string main_vessel_guid = (FlightGlobals.ActiveVessel ??
+                               space_tracking?.SelectedVessel)?.id.ToString();
```
pleroy: Pretty soon there will be a function for this.  Use it?

PR #2114, opened 2019-03-30
Hide the flight planner when there is no vessel

pleroy:


ksp_plugin_adapter/flight_planner.cs
```diff
@@ -34,6 +34,12 @@ class FlightPlanner : SupervisedWindowRenderer {
     if (UnityEngine.GUILayout.Button("Flight plan...")) {
       Toggle();
     }
+    // Override the state of the toggle if there is no active vessel.
+    string vessel_guid = vessel_ ?.id.ToString();
```
eggrobin: Extraneous space.

PR #2123, opened 2019-04-06
Finite differences

eggrobin:
Implement finite differences at any node on equally-spaced nodes.

Remove `backward_differences`, which had coefficients but no implementation, and only for backward differences.


numerics/finite_difference_body.hpp
```diff
@@ -0,0 +1,51 @@
...
+
+namespace principia {
+namespace numerics {
+namespace internal_finite_difference {
+
+template<typename Value, typename Argument, int n>
+Derivative<Value, Argument> FiniteDifference(
+    std::array<Value, n> const& values,
+    Argument const& step,
+    int offset) {
```
pleroy: `const`

numerics/finite_difference_body.hpp
```diff
@@ -0,0 +1,51 @@
...
+    // In the general case, we evaluate the sum Σᵢ aᵢ f(xᵢ), with Σᵢ aᵢ = 0,
+    // where the sums over i run from 0 to n - 1, as
+    //   Σⱼ (Σₖ aₖ) (f(xⱼ) - f(xⱼ₊₁)),
+    // where the sum over j runs from 0 to n - 2, and the sum over
+    // k runs from 0 to j.
+    double numerator = 0;
+    Difference<Value> sum{};
+    for (int j = 0; j <= n - 2; ++j) {
+      numerator += numerators[j];
+      Difference<Value> difference = values[j] - values[j + 1];
```
pleroy: `const`

numerics/finite_difference_body.hpp
```diff
@@ -0,0 +1,51 @@
...
+      sum += numerator * difference;
+    }
+    return sum / (denominator * step);
```
pleroy: You could move the declaration of `sum` before the `if` and the `return` after the `if`.  Small is beautiful when it comes to numerical code.

PR #2124, opened 2019-04-06
Dynamic scaling of the UI

pleroy:
The real solution would be to use the new canvas API, but that's quite a bit of work.  This solution is a bit hacky, but it works well enough and improves usability on modern displays.


ksp_plugin_adapter/window_renderer.cs
```diff
@@ -6,9 +6,97 @@
...
+  protected UnityEngine.GUILayoutOption GUILayoutHeight(int units) {
+    return UnityEngine.GUILayout.Height(unit_ * units);
+  }
+
+  protected UnityEngine.GUILayoutOption GUILayoutMinWidth(int units) {
+    return UnityEngine.GUILayout.MinWidth(unit_ * units);
+  }
+
+  protected UnityEngine.GUILayoutOption GUILayoutWidth(int units) {
+    return UnityEngine.GUILayout.Width(unit_ * units);
```
eggrobin: `UnityEngine.GUILayout.Width(Width(units))`

PR #2127, opened 2019-04-08
Add support for editing burn parameters using a text field

### LINKED ISSUE #2121: Add direct values input apart from sliders.
Please add/convert numerical value fields, so we can input values directly for all Flight Plan parameters. Something similar to MJ maneuver editor.

I've been playing with Principia for the past few years, and it's great, except that horrible sliders.
I mean, with game stuttering you will almost always overshoot. I could plot Moon trajectory in few seconds with numeric input, but with sliders, it sometimes takes like minutes. Not to mention longer plots, when you want to put MN on distant body periapsis, it's just ridiculous how much time is wasted when trying to fiddle with those sliders, when I know exact number I want already.
Of course, helpers would be nice like in Precise Manouver, when you can one-click snap MN to AP/PE/AN/DN etc., but that is just wishful thinking =)

So, for the love of Newton, PLEASE allow to input exact numerical values apart from sliders, converting current read-only fields to editable and getting values from them should be enough.

pleroy:
Fix #2121.


ksp_plugin_adapter/flight_planner.cs
```diff
@@ -129,13 +128,16 @@ class FlightPlanner : SupervisedWindowRenderer {
...
+      var style = new UnityEngine.GUIStyle(UnityEngine.GUI.skin.textField);
+      string message = "";
+      if (final_time_.value != actual_final_time) {
+        style.normal.textColor = XKCDColors.Orange;
+        message = "Timed out after " +
+                  FormatPositiveTimeSpan(TimeSpan.FromSeconds(
+                      actual_final_time -
+                      plugin_.FlightPlanGetInitialTime(vessel_guid)));
+      }
+      UnityEngine.GUILayout.TextField(message, style);
```
eggrobin: This is either empty or orange, we might as well make it unconditionally orange.

ksp_plugin_adapter/differential_slider.cs
```diff
@@ -6,31 +6,59 @@ namespace ksp_plugin_adapter {
...
 
-  public double value { get; set; }
+  public double value {
+    get {
+      return value_ ?? 0.0;
+    }
+    set {
+      if (!value_.HasValue || value_ != value) {
+        value_ = value;
+        formatted_value_ = formatter_(value_.Value);
```
eggrobin: ...

ksp_plugin_adapter/burn_editor.cs
```diff
@@ -255,13 +255,32 @@ class BurnEditor : ScalingRenderer {
...
+    if (!FlightPlanner.TryParseTimeSpan(str, out ts)) {
+      return false;
+    }
+    value = ts.TotalSeconds;
+    return true;
+  }
+
+  private double previous_burn_final_time => previous_burn_?.final_time ??
+                                             plugin_.FlightPlanGetInitialTime(
+                                                 vessel_.id.ToString());
```
eggrobin: This is a confusing name, since it may not be `previous_burn_.final_time`.
Call it `time_base`, this is what we call it in the UI anyway.

ksp_plugin_adapter/differential_slider.cs
```diff
@@ -72,15 +149,22 @@ internal class DifferentialSlider : ScalingRenderer {
...
           value_changed = true;
-          value = 0;
+          // Force a change of value so that any input is discarded.
+          value = zero_value_ + 1;
+          value = zero_value_;
         }
         if (slider_position_ != 0.0) {
           value_changed = true;
+          // Moving the slider doesn't cause a loss of focus so we terminate
+          // input if necessary.
```
eggrobin: If I recall correctly, the issue is not that it doesn't cause a loss of focus (it does terminate input if it is another slider), rather that we start changing the value before committing the text input (perhaps because of event order?).

ksp_plugin_adapter/burn_editor.cs
```diff
@@ -35,21 +36,18 @@ class BurnEditor : ScalingRenderer {
...
                                log10_upper_rate : Log10ΔvUpperRate,
                                text_colour      : XKCDColors.PurplePink);
-    initial_time_ =
+    previous_coast_duration_ =
         new DifferentialSlider(
                 label            : "t initial",
                 unit             : null,
                 log10_lower_rate : Log10TimeLowerRate,
                 log10_upper_rate : Log10TimeUpperRate,
+                zero_value       : 0.001,
```
eggrobin: This deserves a comment.

ksp_plugin_adapter/differential_slider.cs
pleroy: I am not proud of it.  (I am, a bit.)

ksp_plugin_adapter/burn_editor.cs
pleroy: Done.

ksp_plugin_adapter/differential_slider.cs
```diff
@@ -91,19 +175,23 @@ internal class DifferentialSlider : ScalingRenderer {
...
   private readonly double max_value_;
 
-  private readonly ValueFormatter format_;
+  private readonly ValueFormatter formatter_;
+  private readonly ValueParser parser_;
   private readonly UnityEngine.Color? text_colour_;
+
+  private float slider_position_ = 0.0f;
+  private DateTime last_time_;
+  private double? value_;
```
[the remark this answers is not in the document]
pleroy: Added comments.

ksp_plugin_adapter/differential_slider.cs
```diff
@@ -72,15 +149,22 @@ internal class DifferentialSlider : ScalingRenderer {
 
         if (UnityEngine.GUILayout.Button("0", GUILayoutWidth(1))) {
           value_changed = true;
-          value = 0;
+          // Force a change of value so that any input is discarded.
+          value = zero_value_ + 1;
```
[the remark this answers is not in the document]
pleroy: Done.

ksp_plugin_adapter/differential_slider.cs
pleroy: I *think* that clicking on the slider/button of the same differential slider doesn't cause a loss of focus at all, and clicking elsewhere does.  I don't feel like experimenting to nail down the exact semantics of Unity, though.

ksp_plugin_adapter/burn_editor.cs
pleroy: Done.

ksp_plugin_adapter/flight_planner.cs
pleroy: Done.

ksp_plugin_adapter/burn_editor.cs
```diff
@@ -155,7 +157,7 @@ class BurnEditor : ScalingRenderer {
         specific_impulse_in_seconds_g0 = specific_impulse_in_seconds_g0_,
         frame = reference_frame_selector_.FrameParameters(),
         initial_time = previous_coast_duration_.value +
-                       previous_burn_final_time,
+                       time_base,
```
eggrobin: Fits on the preceding line.

PR #2131, opened 2019-04-10
Fix null pointer exception in the flight plan and avoid UI flickering

pleroy:


ksp_plugin_adapter/flight_planner.cs
```diff
@@ -413,13 +440,12 @@ class FlightPlanner : SupervisedWindowRenderer {
     return true;
   }
 
-  // Not owned.
```
eggrobin: That comment should have been on `plugin_`, but it still applies.

ksp_plugin_adapter/flight_planner.cs
```diff
@@ -266,8 +280,7 @@ class FlightPlanner : SupervisedWindowRenderer {
     for (int i = 0; i < burn_editors_.Count; ++i) {
       NavigationManoeuvre manoeuvre =
           plugin_.FlightPlanGetManoeuvre(vessel_guid, i);
-      // TODO(phl): Evil changes of widgets between layout and repaint...
-      if (manoeuvre.final_time > current_time) {
+      if (first_future_manoeuvre_.HasValue && i >= first_future_manoeuvre_) {
```
eggrobin: `if (i >= first_future_manoeuvre_)`, `null` compares false with non-null values.

ksp_plugin_adapter/flight_planner.cs
```diff
@@ -266,8 +280,7 @@ class FlightPlanner : SupervisedWindowRenderer {
     for (int i = 0; i < burn_editors_.Count; ++i) {
```
eggrobin: Since we have computed `first_future_manoeuvre_` already, could we not get rid of the loop, and just do it for `i == first_future_manoeuvre_`?

ksp_plugin_adapter/flight_planner.cs
pleroy: Done.

ksp_plugin_adapter/flight_planner.cs
pleroy: OBE.

ksp_plugin_adapter/flight_planner.cs
pleroy: Done.

pleroy:
retest this please

PR #2132, opened 2019-04-11
More compact UI

pleroy:


ksp_plugin_adapter/burn_editor.cs
```diff
@@ -62,12 +62,20 @@ class BurnEditor : ScalingRenderer {
...
   // changed.
-  public bool Render(bool enabled) {
+  public bool Render(string header, bool enabled) {
     bool changed = false;
+    using (new UnityEngine.GUILayout.HorizontalScope()) {
+      UnityEngine.GUILayout.Label(header);
+      string frame_warning = "";
+      if (!reference_frame_selector_.FrameParameters().Equals(
+              adapter_.plotting_frame_selector_.FrameParameters())) {
+        frame_warning = "Manœuvre frame differs from plotting frame";
```
eggrobin: This really should not be a warning (it was, but it should not be); people get confused and think this is bad, whereas it is the right thing to do in many circumstances.

pleroy: Done.

PR #2133, opened 2019-04-12
Use the warning style if the initial time of a manoeuvre is too large

pleroy:


ksp_plugin_adapter/differential_slider.cs
```diff
@@ -77,10 +83,12 @@ internal class DifferentialSlider : ScalingRenderer {
...
       if (enabled) {
-        // If the text is not syntactically correct, inform the user by drawing
-        // it in the warning style.
+        // If the text is not syntactically correct, or it exceeds the upper
+        // bound, inform the user by drawing it in the warning style.  Note the
+        // fudge factor to account for uncertainty in text/double conversions.
         var style = Style.RightAligned(UnityEngine.GUI.skin.textField);
-        if (!parser_(formatted_value_, out double v1)) {
+        if (!parser_(formatted_value_, out double v1) ||
+            v1 > max_value_ * (1 + 1.0e-5)) {
```
eggrobin: Use a fixed tolerance, since the display is fixed point.

PR #2136, opened 2019-04-14
Render the guidance node in the adapter and not in the flight planner

### LINKED ISSUE #1565: Switching between vessels does not update maneuver node.
When I switch from one vessel to another, with an active maneuver plan, the node for this maneuver isn't the one that is planned next. I have to reactivate the show maneuver in navball checkbox to update the maneuver into the right direction.

pleroy:
Fix #1565.

### DIFF
--- a/ksp_plugin_adapter/ksp_plugin_adapter.cs
+++ b/ksp_plugin_adapter/ksp_plugin_adapter.cs
@@ -1438,7 +1438,8 @@ public partial class PrincipiaPluginAdapter
     if (plugin_.HasVessel(vessel_guid) &&
         plugin_.FlightPlanExists(vessel_guid)) {
       // Here the vessel is known to the plugin and has a flight plan.
-      //Duplicate.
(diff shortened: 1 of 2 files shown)


ksp_plugin_adapter/ksp_plugin_adapter.cs
```diff
@@ -1532,6 +1433,191 @@ private enum UnityLayers {
     }
   }
 
+  private void RenderGuidance(Vessel active_vessel) {
+    string vessel_guid = active_vessel.id.ToString();
+    if (plugin_.HasVessel(vessel_guid) &&
+        plugin_.FlightPlanExists(vessel_guid)) {
+      // Here the vessel is known to the plugin and has a flight plan.
+      //Duplicate.
```
eggrobin: What is meant by duplicate here?

Badly formatted comment.

pleroy: Clarified.

PR #2143, opened 2019-04-24
Take the thrust limiter into account

### LINKED ISSUE #2128: Take the engine thrust limiter into account in manœuvres
Would this be possible to implement inclusion of engine's thrust offset in maneuver planning? For ex.  it's hard to set capture burn correctly from interplanetary trajectory with ion engine when you can't supply enough energy.

eggrobin:
Fix #2128.


pleroy:
retest this please

Issue #2090, opened 2019-03-10
Remove Ephemeris::ForgetBefore


pleroy:
This function breaks a number of invariants, and it is dangerous to call in parallel with iteration over the Ephemeris.  It may be a remnant of a time when we were concerned about memory usage.  One option would be to not serialize the deep past, so when reloading a save memory would be reclaimed, but without ill-effects on the actual computations.

pleroy:
Better idea: have a lock system that ensures that the interval [*t*, +∞[ is not affected by `ForgetBefore` when computing a prediction that starts at *t*.  Acquire at the beginning of the prognostication loop, release at the end (probably using an RAII object).  `ForgetBefore` can only discard the unlocked part of the timeline.  Locking can fail if *t* has already been forgotten.

PR #2147, opened 2019-04-27
Remove checkpointing of Ephemeris and ContinuousTrajectory

pleroy:
**Do not** merge this prior to Fáry!  This is preparatory work for #2090, it's not urgent.

This has been tested with saves created in mid-December and mid-April.


pleroy:
Superseded by #2149.

PR #2149, opened 2019-05-01
Checkpointing done right

pleroy:
This is intended to replace #2147.


physics/checkpointer.hpp
```diff
@@ -0,0 +1,95 @@
...
+// ForgetBefore: this ensures that there is always a sufficient old checkpoint
+// available the next time serialization is performed.
+// Logically checkpoints would serialize to/deserialize from a specific message,
+// but for historical reasons they just fill some fields of a message that may
+// contain other information.
+// This class is thread-safe.
+template<typename Message>
+class Checkpointer {
+ public:
+  // A function that reconstructs an object from a checkpoint.  It must return
```
eggrobin: A function that reconstructs an object from a checkpoint **as a side effect**, I got confused looking for the object in the signature.

Alternatively, templatize on the object and pass it by non-const reference.

physics/checkpointer_body.hpp
```diff
@@ -0,0 +1,75 @@
...
+template<typename Message>
+Instant Checkpointer<Message>::WriteToMessage(
+    not_null<Message*> const message) const {
+  absl::ReaderMutexLock l(&lock_);
+  if (checkpoints_.empty()) {
+    // TODO(phl): declare this next to Instant.
+    static Instant infinite_future = Instant() + quantities::Infinity<Time>();
+    return infinite_future;
+  } else {
+    message->MergeFrom(checkpoints_.cbegin()->second);
```
eggrobin: Will this have the correct behaviour for repeated fields, since those are appended instead of overwritten? If it does not, there should be a comment warning of that in the header file.

physics/checkpointer.hpp
pleroy: I started with templatizing on the object, but it looked silly because it would take the pointer at construction and store it just to pass it back to the callbacks.  Incidentally, this schema would technically make it possible to checkpoint *several* objects together.

Beefed up the comment.

physics/checkpointer_body.hpp
pleroy: Added a comment.

pleroy:
retest this please

pleroy:
Verified that both recent and old saves could be loaded.

PR #2150, opened 2019-05-04
Quis custodiet ipsos custodes?

pleroy:
A mechanism for protecting the `Ephemeris` from `ForgetBefore`, not used yet.  Will help with #2090.


pleroy:
retest this please

physics/ephemeris.hpp
```diff
@@ -286,6 +284,18 @@ class Ephemeris {
   static not_null<std::unique_ptr<Ephemeris>> ReadFromMessage(
       serialization::Ephemeris const& message) EXCLUDES(lock_);
 
+  // A |Guard| is an RAII object that protects a critical section against
+  // changes to |t_min| due to calls to |TryToForgetBefore|.
+  class Guard final {
```
eggrobin: Make this move-only.

physics/ephemeris.hpp
```diff
@@ -185,13 +185,11 @@ class Ephemeris {
...
-  // increase, so if it is called is parallel with code that iterates over the
-  // trajectories of the ephemeris, it can cause trouble.
-  // TODO(phl): Consider eliminating this function and truncating on
-  // serialization/deserialization.
-  virtual void ForgetBefore(Instant const& t) EXCLUDES(lock_);
+  // If the time |t| is not protected by a |Guard|, calls |ForgetBefore| on all
+  // trajectories and returns true, after which |t_min() == t|.  If the time |t|
+  // is protected by a |Guard|, returns false; the actual action is delayed
+  // until the destruction of the |Guard|.
+  virtual bool TryToForgetBefore(Instant const& t) EXCLUDES(lock_);
```
eggrobin: This doesn't just try, it *will* eventually forget before `t`; `MarkForForgetting`?

In what circumstances would we use the result? `ForgetBefore` only exists for performance purposes, I can't think of a case where we would do something differently if we cannot forget immediately.

physics/ephemeris.hpp
pleroy: Good point, done.

physics/ephemeris.hpp
pleroy: 1. I'd rather have an identifier that contains `ForgetBefore` for clarity.
1. In tests, it's convenient to check that the right thing happens.  Could also be useful in `CHECK`s.

PR #2151, opened 2019-05-04
Use Ephemeris::Guard to protect the computation of the prognostication

### LINKED ISSUE #2090: Remove Ephemeris::ForgetBefore

pleroy:
Fix #2090, but not by removing `ForgetBefore`.  Tested in the game.


pleroy:
retest this please

pleroy:
retest this please

pleroy:
retest this please

pleroy:
retest this please

PR #2153, opened 2019-05-05
A better reference plane for the the ECI etc. frames.

### LINKED ISSUE #1841: Show ☊ and ☋ with respect to the equator in body-centred inertial frames
This came up during the design discussion in #1840.

This is not quite as simple as changing the conditions for [prediction nodes](https://github.com/mockingbirdnest/Principia/blob/2018051512-Darboux/ksp_plugin_adapter/ksp_plugin_adapter.cs#L1916-L1919) and [flight plan nodes](https://github.com/mockingbirdnest/Principia/blob/2018051512-Darboux/ksp_plugin_adapter/ksp_plugin_adapter.cs#L2000-L2003), since the xy plane for the body-centred inertial frames is currently the global reference plane (Earth's equator in RSS, Kerbin's in stock) rather than the plane of the equator of the relevant body.

Perhaps it would be easiest to change the definition of [`BodyCentredNonRotatingDynamicFrame`](https://github.com/mockingbirdnest/Principia/blob/2018051512-Darboux/ksp_plugin_adapter/ksp_plugin_adapter.cs#L2000-L2003) so that it uses the same z axis as [`BodySurfaceDynamicFrame`](https://github.com/mockingbirdnest/Principia/blob/2018051512-Darboux/physics/body_surface_dynamic_frame.hpp#L33-L36), making the y axis the node Q rather than making the x axis the point B from figure 1 of [the 2009 report of the IAU WGCCRE](https://astropedia.astrogeology.usgs.gov/download/Docs/WGCCRE/WGCCRE2009reprint.pdf).

Note that Q≠`♈︎` for Earth, since α₀=0; this motivates making the y axis the node Q, rather than the x axis, so that the x axis points towards `♈︎` for Earth (in other words, `BodyCentredNonRotatingDynamicFrame` for Earth would be the same as `Barycentric` in RSS).

Addendum: it seems GitHub ignores VS-15 on U+2648 ARIES unless it is within a code block.

eggrobin:
This will make it possible to address #1841; additional logic would be needed for to properly address the irrelevance of the solar equator, other than that this is a simple matter of changing the following condition in the adapter.
https://github.com/mockingbirdnest/Principia/blob/80be4b3fdf0c0fe0b6f9e038815171ffc1ba82f7/ksp_plugin_adapter/ksp_plugin_adapter.cs#L1922-L1925

Note: this breaks tests (some values that were very small have become 0 because of the changes in (bi)equatorial), and it probably needs some new tests to check the orientation.

### DIFF
--- a/physics/body_centred_non_rotating_dynamic_frame.hpp
+++ b/physics/body_centred_non_rotating_dynamic_frame.hpp
@@ -44,7 +44,7 @@ using quantities::Acceleration;
 // - the Z axis is the pole Z from the figure.
 // Note that, when |InertialFrame| is the ICRS, these axes make |ThisFrame| the
 // usual celestial reference frame (the GCRS) if |centre| is the Earth, as the X
-// axis will point towards ♈; however, if |centre| is another planet,
+// axis will point towards ♈︎; however, if |centre| is another planet,
 // |ThisFrame| will not have the axes of its natural celestial reference frame,
 // as the X axis will not point towards its equinox.
 // REMOVE BEFORE FLIGHT: Do we really want that? This seems like a very
--- a/physics/body_centred_non_rotating_dynamic_frame.hpp
+++ b/physics/body_centred_non_rotating_dynamic_frame.hpp
@@ -47,10 +47,6 @@ using quantities::Acceleration;
 // axis will point towards ♈︎; however, if |centre| is another planet,
 // |ThisFrame| will not have the axes of its natural celestial reference frame,
 // as the X axis will not point towards its equinox.
-// REMOVE BEFORE FLIGHT: Do we really want that? This seems like a very
-// contrived way get something that only makes sense for the Earth with respect
-// to the ICRS; in particular, if the XY plane of |InertialFrame| is the
-// invariable plane, Q should naturally be the X axis, not the Y axis.
 template<typename InertialFrame, typename ThisFrame>
 class BodyCentredNonRotatingDynamicFrame
     : public DynamicFrame<InertialFrame, ThisFrame> {
--- a/physics/body_centred_non_rotating_dynamic_frame.hpp
+++ b/physics/body_centred_non_rotating_dynamic_frame.hpp
@@ -26,12 +26,31 @@ namespace internal_body_centred_non_rotating_dynamic_frame {
 
 using base::not_null;
 using geometry::Instant;
+using geometry::OrthogonalMap;
 using geometry::Position;
 using geometry::Vector;
 using quantities::Acceleration;
 
-// The origin of the frame is the centre of mass of the body.  The axes are
-// those of |InertialFrame|.
+// The origin of the frame is the centre of mass of the body.  The Y axis is at
+// the intersection of the equator and the XY plane of |InertialFrame|, in the
+// direction 90° + α from the X axis of |InertialFrame|, where α is the right
+// ascension of the |polar_axis|.  The Z axis is the |polar_axis|.  The X axis
+// in on the equator so that the frame has the same orientation as
+// |InertialFrame|.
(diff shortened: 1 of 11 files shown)


physics/body_centred_non_rotating_dynamic_frame.hpp
```diff
@@ -26,12 +26,31 @@ namespace internal_body_centred_non_rotating_dynamic_frame {
...
+// With respect to figure 1 of the 2015 report of the IAU WGCCRE if |polar_axis|
+// is the north pole, or figure 2 if |polar_axis| is the positive pole,
+// - the Y axis is the node Q from the figure;
+// - the Z axis is the pole Z from the figure.
+// Note that, when |InertialFrame| is the ICRS, these axes make |ThisFrame| the
+// usual celestial reference frame (the GCRS) if |centre| is the Earth, as the X
+// axis will point towards ♈︎; however, if |centre| is another planet,
+// |ThisFrame| will not have the axes of its natural celestial reference frame,
+// as the X axis will not point towards its equinox.
+// REMOVE BEFORE FLIGHT: Do we really want that? This seems like a very
```
pleroy: Remove?

physics/body_centred_non_rotating_dynamic_frame_body.hpp
```diff
@@ -4,22 +4,45 @@
...
+using geometry::Rotation;
 
 template<typename InertialFrame, typename ThisFrame>
 BodyCentredNonRotatingDynamicFrame<InertialFrame, ThisFrame>::
-BodyCentredNonRotatingDynamicFrame(
-    not_null<Ephemeris<InertialFrame> const*> const ephemeris,
-    not_null<MassiveBody const*> const centre)
+    BodyCentredNonRotatingDynamicFrame(
+        not_null<Ephemeris<InertialFrame> const*> const ephemeris,
+        not_null<MassiveBody const*> const centre)
```
pleroy: Revert this change, we don't like the way that `clang-format` does it.

physics/body_centred_non_rotating_dynamic_frame_body.hpp
```diff
@@ -4,22 +4,45 @@
...
+            // class is not templatized on |InertialFrame|.
+            auto const rotating_body =
+                dynamic_cast<RotatingBody<InertialFrame> const*>(&*centre_);
+            if (rotating_body == nullptr) {
+              return Identity<InertialFrame, ThisFrame>().Forget();
+            }
+            // In coordinates, the third parameter is |polar_axis|, but we seem
+            // to be a bit confused as to which of these things should be
+            // vectors or bivectors here.
+            // TODO(egg): Figure that out.
```
pleroy: It feels reasonable to use a bivector for the axis.  I think we have a similar "confused" comment somewhere else.

physics/body_centred_non_rotating_dynamic_frame_body.hpp
```diff
@@ -45,10 +68,9 @@ BodyCentredNonRotatingDynamicFrame<InertialFrame, ThisFrame>::ToThisFrameAtTime(
     Instant const& t) const {
   DegreesOfFreedom<InertialFrame> const centre_degrees_of_freedom =
       centre_trajectory_->EvaluateDegreesOfFreedom(t);
-  RigidTransformation<InertialFrame, ThisFrame> const
-      rigid_transformation(centre_degrees_of_freedom.position(),
-                           ThisFrame::origin,
-                           Identity<InertialFrame, ThisFrame>().Forget());
+
+  RigidTransformation<InertialFrame, ThisFrame> const rigid_transformation(
+      centre_degrees_of_freedom.position(), ThisFrame::origin, orthogonal_map_);
```
pleroy: Preserve the previous formatting.

PR #2171, opened 2019-05-18
Change the API of Manœuvre to unify it with Burn

pleroy:
* Make the specification of intensity and timing happen at construction and ensure that a manœuvre is usable immediately after construction.
* Make it easier to clone manœuvres by using `shared_ptr`s.
* This PR is large because of API changes but it is supposed to be semantics-preserving.


ksp_plugin/manœuvre_body.hpp
```diff
@@ -14,30 +14,49 @@ namespace ksp_plugin {
...
+      // would yield NaN.
+      intensity.duration = Time();
+    } else {
+      intensity.duration =
+          initial_mass_ * specific_impulse() *
+          (1 - std::exp(-speed / specific_impulse())) / thrust();
+    }
+  } else {
+    CHECK(intensity.direction && intensity.duration);
+    // Циолко́вский's equation.
```
eggrobin: Циолковский without the acute accent, this is not a dictionary entry.

Issue #1936, opened 2018-09-12
[Request] Multiple nodes editing


EthanWang706:
Now, in Principia, if you want to edit the earlier node after planning many nodes, you have to delete the node in front of it, and adding it back after finish editing that node, and the effect to the after node would be a problem, because there's not a way unless you add it back on. But when there's too many nodes, it's kind of impossible to delete them and add them on again. The request is I can edit nodes earlier anytime, so multiple node orbit planning could get a solution. 
Another request is choose to show how will next node be after completing this node, another way to say it is show the after node dynamically, so if there is a little error, caused by RCS or other things, Principia could still show wether the node you planned before is working or need fix, instead of delete the whole flight plan then add nodes back on again.

EthanWang706:
Thanks

kc9jud:
If I wanted to take a crack at this and come up with a PR, where should I start? It looks like `ksp_plugin/interface_flight_plan.cpp`, `ksp_plugin/flight_plan.hpp`, and `ksp_plugin/flight_plan.cpp` are relevant? Are there any particular caveats I should be aware of?

eggrobin:
@kc9jud 
> where should I start

The first thing to do before starting to write code to resolve this issue is to figure out what the exact behaviour should be. We have been unsure about that, which is one of the reasons why we have not made progress here.

Right now, the manoeuvres are specified with an absolute time, conceptually, something like “starting at 2019-02-17T00:38:52 UTC, impart a Δv of 300 m/s in the tangent direction, 5 m/s in the normal direction, 10 m/s in the binormal direction, with the Frenet frame defined by the trajectory in the ECI frame”.

The problem is that with such a specification, changing an earlier burn is likely to turn later ones into nonsense.
For instance, if the flight plan consists of a translunar injection (TLI) from a circular low earth orbit, followed by a capture into lunar orbit, changing the timing of the TLI will have the effect of putting the capture burn in the middle of nowhere, either on the way to the Moon or on the way back; it will cease to be a capture burn, and will have to be moved back to the right place to become one again.

We think a possible approach here would be to somehow define the manoeuvre timings with respect to trajectory “events”, e.g., to define a manoeuvre as “starting at 30 s before perigee, impart a Δv of 300 m/s in the tangent direction, 5 m/s in the normal direction, 10 m/s in the binormal direction, with the Frenet frame defined by the trajectory in the ECI frame”.

[...]

PR #2174, opened 2019-05-19
Changes to the FlightPlan API

pleroy:
* Makes it possible to keep "broken" manœuvres in the flight plan.
  * Tracks the number of anomalous manœuvres in case of problems during integration.
  * The trajectories after the first anomalous one have length 0.
* Reports errors to the caller using `Status`.
* The functions that change manœuvres generally have an effect (which can include making parts of the plan anomalous), except if the input parameters are absurd.
* This is only the C++ side of the changes, the interface tries to preserve compatibility.
* This is the first step in addressing #1936.


ksp_plugin/flight_plan.hpp
```diff
@@ -113,55 +140,58 @@ class FlightPlan {
       serialization::FlightPlan const& message,
       not_null<Ephemeris<Barycentric>*> ephemeris);
 
-  static std::int64_t constexpr max_ephemeris_steps_per_frame = 1000;
+  static constexpr std::int64_t max_ephemeris_steps_per_frame = 1000;
+
+  static constexpr Error bad_desired_final_time = Error::INVALID_ARGUMENT;
+  static constexpr Error does_not_fit = Error::INVALID_ARGUMENT;
+  static constexpr Error singular = Error::INVALID_ARGUMENT;
```
eggrobin: The fact that these three are equal is going to lead to a lot of confusion, especially since we test for specific statuses in the interface.

`does_not_fit` and `bad_desired_final_time` are both `FAILED_PRECONDITION`-like (they are about the compatibility of the argument with the flight plan, rather than the argument being intrinsically bad like `manœuvre.IsSingular()`).
Make one of them `OUT_OF_RANGE` and we will have distinct statuses.

ksp_plugin/flight_plan.hpp
```diff
@@ -48,19 +52,34 @@ class FlightPlan {
...
+  // The number of manœuvres in the flight plan.
   virtual int number_of_manœuvres() const;
-  // |index| must be in [0, number_of_manœuvres()[.
+
+  // The number of manœuvres at the end of flight plan that are anomalous, i.e.,
+  // lead to their burn or a subsequent trajectory failing to integrate.  Note
+  // that this returns 1 for a flight plan without manœuvres if the first coast
+  // fails to integrate.  The functions that change manœuvres may change the
+  // number of anomalous manœuvres.
+  virtual int number_of_anomalous_manœuvres() const;
```
eggrobin: If`number_of_anomalous_manœuvres` can be less than `number_of_manœuvres`, there is a naming problem.

Let us define an anomalous manœuvre as one whose burn integration cannot be started (because it starts after an anomalous segment). This becomes
`(anomalous_segments_ - 1) / 2`, rather than ~~`(anomalous_segments_ + 1) / 2`~~ (0 for 1 or 2 anomalous segments, 1 for 3 because the segment preceding the last burn is anomalous).

This is only used outside of tests as `number_of_anomalous_manœuvres() <= 1`, which should be changed to `number_of_anomalous_manœuvres() == 1`: the existing semantics are that there are no anomalous manœuvres, that is, the integration of their burn can at least start, and they are editable in the UI.

ksp_plugin/flight_plan.hpp
pleroy: Changed the first two to be `OUT_OF_RANGE`.

ksp_plugin/flight_plan.hpp
pleroy: Done.

PR #2177, opened 2019-05-25
Multiline label in the frame selector

### LINKED ISSUE #2169: The description of the target LVLH frame is garbled
For some reason the labels do not appear to expand vertically to accommodate the several lines of text.

screenshot63

pleroy:
Fix #2169.


pleroy:
The problem was introduced when UI resizing was implemented, in Fáry.

PR #2192, opened 2019-06-02
Correctly handle the case where a single Visual Studio version is installed

eggrobin:
PowerShell works in mysterious ways.


pleroy:
retest this please

PR #2197, opened 2019-06-09
Support edition of all manœuvres in a flight plan

### LINKED ISSUE #1936: [Request] Multiple nodes editing

pleroy:
* Remove `ReplaceLast` everywhere and use `Replace` instead.
* Remove the `enabled` boolean from the UI: all burn editors are always enabled.
* Change the flight planner to handle all burn editors identically.

Fix #1936.

### DIFF
--- a/ksp_plugin_adapter/burn_editor.cs
+++ b/ksp_plugin_adapter/burn_editor.cs
@@ -59,11 +59,10 @@ class BurnEditor : ScalingRenderer {
 
   // Renders the |BurnEditor|.  Returns true if and only if the settings were
   // changed.
-  public bool Render(string header,
-                     double? actual_final_time = null) {
+  public bool Render(string header, double actual_final_time) {
     bool changed = false;
-    previous_coast_duration_.max_value =
-        (actual_final_time ?? double.PositiveInfinity) - time_base;
+    //TODO(phl):Is this right?
+    previous_coast_duration_.max_value = actual_final_time - time_base;
     using (new UnityEngine.GUILayout.HorizontalScope()) {
       UnityEngine.GUILayout.Label(header);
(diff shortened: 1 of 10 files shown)


ksp_plugin_adapter/burn_editor.cs
```diff
@@ -59,12 +59,9 @@ class BurnEditor : ScalingRenderer {
 
   // Renders the |BurnEditor|.  Returns true if and only if the settings were
   // changed.
-  public bool Render(string header,
-                     bool enabled,
-                     double? actual_final_time = null) {
+  public bool Render(string header, double final_time) {
```
eggrobin: I think we should retain some sort of `enabled=false` on anomalous manoeuvres: editing them would happen without seeing any resulting effects.
A manoeuvre cannot be anomalous because of its own parameters, except for its timing; could it make sense to have the timing enabled on all manoeuvres, but the other parameters only enabled on non-anomalous ones?

pleroy: Yes, that makes sense.  It's also a good visual feedback.  Will do.

PR #2199, opened 2019-06-10
Apply some ReSharper fixes to our code

pleroy:
Two types of changes:
1. Systematic changes of things that we always want (e.g., unused `using` clauses).
1. Proposed changes that we may or may not want applied by hand.

No contentious things like adjusting naming conventions or reformatting.

### DIFF
--- a/ksp_plugin_adapter/main_window.cs
+++ b/ksp_plugin_adapter/main_window.cs
@@ -117,7 +117,7 @@ internal class MainWindow : SupervisedWindowRenderer {
     Log.SetVerboseLogging(verbose_logging_);
 
     if (must_record_journal_) {
-      journaling_ = true;
+      journalling_ = true;
       Log.ActivateRecorder(true);
     }
   }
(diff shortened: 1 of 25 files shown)


ksp_plugin_adapter/main_window.cs
```diff
@@ -367,13 +364,13 @@ internal class MainWindow : SupervisedWindowRenderer {
           value   : must_record_journal_,
           text    : "Record journal (starts on load)");
       UnityEngine.GUILayout.Label(
-          "Journalling is " + (journaling_ ? "ON" : "OFF"),
+          "Journalling is " + (journalling_ ? "ON" : "OFF"),
```
eggrobin: It seems the spelling of the variable was good and that the string should be changed instead: https://www.lexico.com/en/definition/journal.

ksp_plugin_adapter/optional_marshaler.cs
```diff
@@ -35,16 +35,14 @@ internal class OptionalMarshaler<T> : ICustomMarshaler where T : struct {
       throw Log.Fatal("The runtime returns null for null objects");
     }
     T value;
-    var value_if_boxed = managed_object as T?;
-    if (value_if_boxed != null) {
-      value = value_if_boxed.Value;
+    if (managed_object is T value_if_boxed) {
```
eggrobin: Fancy!

ksp_plugin_adapter/main_window.cs
pleroy: Done.

ksp_plugin_adapter/optional_marshaler.cs
pleroy: Indeed, it does that with a single click.  I didn't even know that this existed.

PR #2201, opened 2019-06-13
A straightforward translation of Fukushima's code for elliptic functions

pleroy:
Complete with a unit test and a benchmark.

```
Run on (4 X 3310 MHz CPU s)
CPU Caches:
  L1 Data 32K (x4)
  L1 Instruction 32K (x4)
  L2 Unified 262K (x4)
  L3 Unified 6291K (x1)
---------------------------------------------------------
Benchmark                  Time           CPU Iterations
---------------------------------------------------------
BM_Gscd                  166 ns        166 ns   16780000
BM_Gscd                  166 ns        165 ns   16780000
BM_Gscd                  166 ns        165 ns   16780000
BM_Gscd                  166 ns        166 ns   16780000
BM_Gscd                  166 ns        165 ns   16780000
BM_Gscd                  166 ns        166 ns   16780000
BM_Gscd                  167 ns        166 ns   16780000
BM_Gscd                  166 ns        165 ns   16780000
BM_Gscd                  166 ns        165 ns   16780000
BM_Gscd                  167 ns        166 ns   16780000
BM_Gscd_mean             166 ns        166 ns         10
BM_Gscd_median           166 ns        166 ns         10
BM_Gscd_stddev             0 ns          1 ns         10
```


serialization/testing_utilities.proto
```diff
@@ -0,0 +1,14 @@
+syntax = "proto2";
+
+package principia.serialization;
+
+// An auxiliary message used to represent tabulated data.  Useful for tests that
+// verify precomputed results.  The arguments and values are to be interpreted
+// by the test.
```
eggrobin: .proto files are cheap, and we are using this for only one table; just give it fields m, u, s, c, d (or more verbose names as appropriate).

numerics/elliptic_functions.hpp
```diff
@@ -0,0 +1,16 @@
...
+// Fukushima, Toshio. (2012). xgscd.txt (Fortran program package to compute the
+// Jacobian elliptic functions, sn(u|m), cn(u|m), dn(u|m)).
+// Downloaded from:
+// https://www.researchgate.net/publication/233903220_xgscdtxt_Fortran_program_package_to_compute_the_Jacobian_elliptic_functions_snum_cnum_dnum
+namespace principia {
+namespace numerics {
+
+void Gscd(double const u, double const mc, double& s, double& c, double& d);
+
+double Elk(double const mc);
```
eggrobin: No const in the declaration for parameters passed by value:
```C++
void Gscd(double u, double mc, double& s, double& c, double& d);

double 🦌(double mc);
```

serialization/testing_utilities.proto
pleroy: Proto are not so cheap when it comes to compilation time.  Furthermore, this proto will be used in the next PR to test the integrals.  I'm keeping it as is for now, we can revisit when we have more use cases.  (It's not like it's really error-prone: if you swap `m` and `u` for instance, the test is going to fail all over the place.)

numerics/elliptic_functions.hpp
pleroy: Done.

PR #2202, opened 2019-06-15
Move the elliptic integrals test to a proto, add a new test, fix bugs

pleroy:


numerics/elliptic_integrals_test.cc
```diff
@@ -1,316 +1,31 @@
-
+﻿
+#include <filesystem>
 #include <vector>
```
eggrobin: This file should have the cpp extension.

pleroy: Amusing.

PR #2203, opened 2019-06-15
A small performance improvement and more sensible benchmarks

pleroy:
Let's exercise only arguments that are in range...
```
Run on (4 X 3310 MHz CPU s)
CPU Caches:
  L1 Data 32K (x4)
  L1 Instruction 32K (x4)
  L2 Unified 262K (x4)
  L3 Unified 6291K (x1)
---------------------------------------------------------
Benchmark                  Time           CPU Iterations
---------------------------------------------------------
BM_Elbdj                 274 ns        274 ns    9976000
BM_Elbdj                 275 ns        274 ns    9976000
BM_Elbdj                 273 ns        271 ns    9976000
BM_Elbdj                 272 ns        274 ns    9976000
BM_Elbdj                 273 ns        271 ns    9976000
BM_Elbdj                 273 ns        272 ns    9976000
BM_Elbdj                 272 ns        272 ns    9976000
BM_Elbdj                 272 ns        272 ns    9976000
BM_Elbdj                 274 ns        272 ns    9976000
BM_Elbdj                 272 ns        272 ns    9976000
BM_Elbdj_mean            273 ns        272 ns         10
BM_Elbdj_median          273 ns        272 ns         10
BM_Elbdj_stddev            1 ns          1 ns         10
```


numerics/elliptic_integrals.cpp
```diff
@@ -60,7 +60,7 @@ void Elbdj(double const phi,
   // TODO(phl): CHECK_EQ(π / 2, phi + phic);
   double m, nc, h, c, x, d2, z, bc, dc, jc, sz, t, v, t2;
 
-  if (phi < 1.345) {
+  if (phi < 1.249) {
```
eggrobin: Can we document this?
Previously the lack of documentation was explained by "translated from undocumented FORTRAN", but if we start making changes we should probably comment our magic constants.

pleroy: I added a big bad block of comments, but mostly this is "let's make the code do what the paper says" (the paper has like 2 pages to explain this constant).

PR #2206, opened 2019-06-16
Rename all the elliptic functions

pleroy:
I am trying to give them names that mimic the _Mathematica_ conventions.  I am also adding a bit of bibliography and some comments to each function.

Note to the reviewer: this is _not_ a diffbase.


numerics/elliptic_integrals.cpp
```diff
@@ -1378,7 +1425,7 @@ void Elbdj(double const phi,
 //
 //     Output: elk
```
eggrobin: I am not sure this comment is very helpful anymore.

pleroy: Maybe it produces an elk...

PR #2211, opened 2019-06-17
 Rewrite BulirschCel and FukushimaEllipticBDJ

pleroy:
Note to the reviewer: this is a diffbase, only c2ca6f2 needs to be reviewed.

### DIFF
--- a/numerics/elliptic_integrals.cpp
+++ b/numerics/elliptic_integrals.cpp
@@ -14,6 +14,8 @@
 // 3. Figure something for the uninitialized variables.
 
 // Bibliography:
+// [Buli69] Bulirsch (1969), Numerical Calculation of Elliptic Integrals and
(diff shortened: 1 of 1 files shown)


numerics/elliptic_integrals.cpp
```diff
@@ -94,66 +96,66 @@ double FukushimaT(double t, double h);
...
   if (kc == 0.0) {
-    if (b == 0.0) {
-      kc = 1.e-16;
-    } else {
-      err = 1;
-      return 1.e99;
-    }
+    // "In this case ... then cel is undefined."
+    DCHECK_EQ(0.0, b) << "kc = " << kc << " nc = " << nc << " a = " << a
+                      << " b = " << b;
```
eggrobin: We should return a NaN (logging in debug mode sounds like a good idea though).

PR #2212, opened 2019-06-18
 Rewrite FukushimaT

pleroy:


numerics/elliptic_integrals.cpp
```diff
@@ -88,6 +92,43 @@ double FukushimaEllipticJsMaclaurinSeries(double y, double n, double m);
 // Fukushima's T function [Fuku11c].
 double FukushimaT(double t, double h);
 
+template<int n>
+struct FukushimaTMaclaurin {
+  static auto constexpr series =
+      std::tuple_cat(FukushimaTMaclaurin<n - 1>::series,
+                     std::tuple<double>{1.0 / (2.0 * n + 1.0)});
```
eggrobin: Use an `integer_sequence` and `1.0 / (2.0 * k + 1.0)` instead of a recursive construction.

pleroy: Done.

PR #2221, opened 2019-06-24
 Rewrite JacobiSNCNDN

pleroy:


numerics/elliptic_functions.cpp
```diff
@@ -138,65 +137,63 @@ void JacobiSNCNDN(double const u,
...
-      sx = c / d;
-      c = kc * s / d;
+    double const k = EllipticK(mc);
+    double const k_over_2 = 0.5 * k;
+    double const three_k_over_2 = 1.5 * k;
+    double const two_k = 2.0 * k;
+    double const five_k_over_2 = 2.5 * k;
+    double const three_k = 3.0 * k;
+    double const seven_k_over_2 = 3.5 * k;
+    double const four_k = 4.0 * k;
```
eggrobin: k/2, 3k/2, 5k/2, and 7k/2 are all used only once; inline them.

Consider trusting common subexpression elimination and inlining `2 * k`, `3 * k`, and `4 * k` as well.

pleroy: First part done.

PR #2241, opened 2019-07-10
A simple solver for Euler's rotation equations

pleroy:
At this point the test only checks that the initial angular momentum can be recomputed, which is not a trivial test.


pleroy:
retest this please

physics/euler_solver.cpp
```diff
@@ -0,0 +1,146 @@
...
+          {B₁₃_ * sech, G_ * Tanh(angle), B₃₁_ * sech});
+    }
+    case Formula::Sphere : {
+      // NOTE(phl): It's unclear how the formulæ degenerate in this case, but
+      // surely λ₃_ becomes 0, so the dependency in time disappears, so this is
+      // my best guess.
+      return initial_angular_momentum_;
+    }
+    default:
+      LOG(FATAL) << "Unexpected formula";
```
eggrobin: Log `formula_`?

PR #2254, opened 2019-07-28
Save the prediction parameters when they are modified in the UI

### LINKED ISSUE #2239: The tolerance and steps settings are not preserved across scene changes
The tolerance and steps settings on the main Principia window reset to the default values (1.0e-2 m and 1.02e3) on every scene change.

pleroy:
Fix #2239.

### DIFF
--- a/ksp_plugin_adapter/main_window.cs
+++ b/ksp_plugin_adapter/main_window.cs
@@ -377,11 +377,9 @@ internal class MainWindow : SupervisedWindowRenderer {
     if (vessel_ != predicted_vessel_()) {
       vessel_ = predicted_vessel_();
     }
-    bool enabled = false;
     AdaptiveStepParameters? adaptive_step_parameters = null;
     string vessel_guid = vessel_?.id.ToString();
     if (vessel_guid != null && plugin.HasVessel(vessel_guid)) {
-      enabled = true;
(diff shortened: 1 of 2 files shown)


ksp_plugin_adapter/main_window.cs
```diff
@@ -377,51 +374,72 @@ internal class MainWindow : SupervisedWindowRenderer {
...
+            default_prediction_length_tolerance_index_;
       }
+      prediction_steps_index_ = Array.FindIndex(
+          prediction_steps_,
+          (long step) => step >= adaptive_step_parameters.Value.max_steps);
+      if (prediction_steps_index_ < 0) {
+        prediction_steps_index_ = default_prediction_steps_index_;
+      }
+    } else {
+      enabled = false;
```
eggrobin: This else clause has no effect.
Let us just get rid of `enabled`, and use `adaptive_step_parameters.HasValue` which is now equivalent below (we display them only if they are there).

Name it at the call site, i.e., `enabled : adaptive_step_parameters.HasValue`.
Consider reordering the arguments so as not to have to name all the other parameters (booleans benefit from naming more often than other parameters).

pleroy: Done.

PR #2265, opened 2019-08-06
Do not show our UI in contexts where it's not useful

### LINKED ISSUE #2262: Principia UI visibility at KSC
If you have the Principia UI open and switch to any building except one of the editors (VAB/SPH), the UI stays open with no way to close it. Since it is of no use, I suggest hiding the window in
* R&D
* Mission Control
* Astronaut Complex
* Administration

pleroy:
Fix #2262.


ksp_plugin_adapter/ksp_plugin_adapter.cs
```diff
@@ -636,7 +663,7 @@ private enum UnityLayers {
 
     if (hide_all_gui_) {
```
eggrobin: `hide_all_gui_ || !in_principia_scene_`, so that it's clearer that we treat those equivalently.

PR #2268, opened 2019-08-07
A change detector test for the dynamics of the solar system

pleroy:


astronomy/solar_system_dynamics_test.cpp
```diff
@@ -272,75 +275,240 @@ TEST_F(SolarSystemDynamicsTest, DISABLED_TenYearsFromJ2000) {
...
+  // For Eris the perihelion drift is surprisingly large.  It's not clear what
+  // ephemeris HORIZONS uses either.  The inclination drift looks like a non-
+  // integrated ephemeris.
+  // For Pluto, WTF is wrong?
+  std::map<int, OrbitError> const expected_planet_orbit_errors{
+      {SolarSystemFactory::Jupiter,
+       {/*separation_per_orbit=*/0.033516 * ArcSecond,
+        /*inclination_drift_per_orbit=*/0.000014 * ArcSecond,
+        /*longitude_of_ascending_node_drift_per_orbit=*/0.010286 * ArcSecond,
+        /*argument_of_periapsis_drift_per_orbit=*/std::nullopt}},
```
eggrobin: Make `argument_of_periapsis_drift_per_orbit` unconditional (but you might want to use `longitude_of_periapsis_drift_per_orbit` instead).
Same for `longitude_of_ascending_node_drift_per_orbit`.

They can be a bit ill-conditioned, but since the point is to have a change detector, let's look at them.

pleroy: Done.  As discussed, sticking to `argument_of_periapsis_drift_per_orbit`.

PR #2271, opened 2019-08-08
Add support for specifying the geopotential using J

### LINKED ISSUE #2270: The definition of geopotentials should allow specifying J
That would avoid having to normalize by hand for Jupiter and Saturn.

pleroy:
Fix #2270.


pleroy:
retest this please

PR #2285, opened 2019-08-15
A class to represent an approximate quantity

pleroy:
First part of addressing #2269.


testing_utilities/approximate_quantity_body.hpp
```diff
@@ -0,0 +1,161 @@
...
+    if (c >= '1' && c <= '9') {
+      error_representation[i] = '0';
+      last_digit_index = i;
+    } else if (is_hexadecimal &&
+               ((c >= 'a' && c <= 'f') || (c >= 'A' && c <= 'F'))) {
+      error_representation[i] = '0';
+      last_digit_index = i;
+    } else if ((!is_hexadecimal && (c == 'e' || c == 'E')) ||
+               (is_hexadecimal && (c == 'p' || c == 'P'))) {
+      CHECK(last_digit_index);
```
eggrobin: `last_digit_index.has_value()`.

testing_utilities/approximate_quantity_body.hpp
```diff
@@ -0,0 +1,161 @@
...
+      break;
+    }
+  }
+  error_representation[*last_digit_index] = '0' + ulp;
+  double const value = std::strtod(representation, nullptr);
+  double const error = std::strtod(error_representation.c_str(), nullptr);
+  LOG(ERROR)<<error_representation;
+  LOG(ERROR)<<error;
```
eggrobin: Remove the traces.

testing_utilities/approximate_quantity_body.hpp
```diff
@@ -0,0 +1,161 @@
...
+    double const min_multiplier,
+    double const max_multiplier,
+    Quantity<Dimensions> const& unit)
+    : representation_(representation),
+      min_multiplier_(min_multiplier),
+      max_multiplier_(max_multiplier),
+      unit_(unit) {}
+
+ApproximateQuantity<double> ApproximateQuantity<double>::Parse(
+    char const* const representation,
```
eggrobin: `std::string_view`

testing_utilities/approximate_quantity_body.hpp
```diff
@@ -0,0 +1,161 @@
...
+    int const ulp) {
+  std::string error_representation(representation);
+  std::optional<int> last_digit_index;
+  bool const is_hexadecimal =
+      error_representation.size() >= 2 &&
+      error_representation[0] == '0' &&
+      (error_representation[1] == 'x' || error_representation[1] == 'X');
+  for (int i = 0; i < error_representation.size(); ++i) {
```
eggrobin: I think that this would be more readable if we were building `error_representation` (by looping on `representation` and appending) than rewriting it with 0s; either way, add a comment describing what we are doing.

testing_utilities/approximate_quantity_body.hpp
```diff
@@ -0,0 +1,161 @@
...
+
+ApproximateQuantity<double> operator""_⑺(char const* const representation) {
+  return ApproximateQuantity<double>::Parse(representation, /*ulp=*/7);
+}
+
+ApproximateQuantity<double> operator""_⑻(char const* const representation) {
+  return ApproximateQuantity<double>::Parse(representation, /*ulp=*/8);
+}
+
+ApproximateQuantity<double> operator""_⑼(char const* const representation) {
```
eggrobin: If we are going to do this beyond ⑴, add 🄐 .. 🄕 as well (morally they should `CHECK` for a hexadecimal `representation` but I am OK with trusting the sanity of the author here).

Maybe invest in a macro to factor the definitions?

Historical note: the parenthesized letters were encoded for compatibility with an ARIB standard https://www.unicode.org/wg2/docs/n3469.pdf (they are ARIB 9433..9458); some of the parenthesized digits are also present in that standard (ARIB 9417..9428), though they stop at 12 in ARIB (the parenthesized digits are from Unicode 1.1, so the rationale for their encoding may be lost to time).

testing_utilities/approximate_quantity_test.cpp
```diff
@@ -0,0 +1,105 @@
...
+  EXPECT_THAT(l3.max(), AlmostEquals(123.46 / Second, 1));
+
+  ApproximateQuantity<Speed> const l4 = 123.45_⑴ * Metre / Second;
+  EXPECT_THAT(l4.min(), AlmostEquals(123.44 * Metre / Second, 0));
+  EXPECT_THAT(l4.max(), AlmostEquals(123.46 * Metre / Second, 1));
+}
+
+TEST(ApproximateQuantityTest, DebugString) {
+  EXPECT_EQ("[+1.23439999999999998e+02, +1.23460000000000008e+02]",
+            (123.45_⑴).DebugString());
```
eggrobin: As discussed, keep the `representation` string around, and output 123.45(1) or 123.45_⑴; probably the former, terminal font fallback is still dodgy.

testing_utilities/approximate_quantity_body.hpp
pleroy: It would not be more readable because I would need more state variables to copy the exponent without patching it.  Added a comment.

testing_utilities/approximate_quantity_body.hpp
pleroy: Done, with check and macro.

testing_utilities/approximate_quantity_body.hpp
```diff
@@ -120,41 +132,29 @@ std::ostream& operator<<(std::ostream& out,
...
+PRINCIPIA_APPROXIMATE_QUANTITY_OPERATOR_DEFINITION(⑴, 1)
+PRINCIPIA_APPROXIMATE_QUANTITY_OPERATOR_DEFINITION(⑵, 2)
+PRINCIPIA_APPROXIMATE_QUANTITY_OPERATOR_DEFINITION(⑶, 3)
+PRINCIPIA_APPROXIMATE_QUANTITY_OPERATOR_DEFINITION(⑷, 4)
+PRINCIPIA_APPROXIMATE_QUANTITY_OPERATOR_DEFINITION(⑸, 5)
+PRINCIPIA_APPROXIMATE_QUANTITY_OPERATOR_DEFINITION(⑹, 6)
+PRINCIPIA_APPROXIMATE_QUANTITY_OPERATOR_DEFINITION(⑺, 7)
+PRINCIPIA_APPROXIMATE_QUANTITY_OPERATOR_DEFINITION(⑻, 8)
+PRINCIPIA_APPROXIMATE_QUANTITY_OPERATOR_DEFINITION(⑼, 9)
+PRINCIPIA_APPROXIMATE_QUANTITY_OPERATOR_DEFINITION(🄐, 10)
```
eggrobin: `PRINCIPIA_APPROXIMATE_QUANTITY_OPERATOR_DEFINITION(🄐, 0xA)`?

Issue #2237, opened 2019-07-07
Trajectory above surface on Phobos ends in Collision


fat-lobyte:
Some trajectories around Phobos end up by Principia detecting collisions.

In this case, when the vessel goes below 2685m "sea level" and 738 m above ground while time warping, it freezes in place. If the timewarp is stopped, the vessel continues moving away but the camera stays in place. The Warning log file reads:

```
W0707 14:14:08.615089  9604 plugin.cpp:778] Vessel Mars 4 - Phobos Lander (4f6db824-027f-47e7-9413-d05abba21700) collided with a celestial: OUT_OF_RANGE: Collision detected
```

Here is a screenshot of the trajectory: 
![screenshot218](https://user-images.githubusercontent.com/173609/60768432-2bc6e500-a0c4-11e9-9589-2a09aa894739.png)

Here is a quicksave file to reacreate the situation: [phobos-plunge-logs.zip](https://github.com/mockingbirdnest/Principia/files/3365624/phobos-plunge-logs.zip)
Logs: [phobos-plunge-logs.zip](https://github.com/mockingbirdnest/Principia/files/3365632/phobos-plunge-logs.zip)

pleroy:
I think that you linked to the logs twice but _didn't_ link to the save.  It would be useful to get it because from the logs everything is as expected, ie the vessel seems to be killed (which is what we are trying to do after the collision).

Now it's a bit embarrassing that we incorrectly detect the collision in this case, but it's not easy to fix: on the one hand we cannot trust KSP to do the right thing, on the other hand we cannot afford to take the body shape into account in the middle of the force computation.

fat-lobyte:
> I think that you linked to the logs twice but didn't link to the save

Oops, my bad. Here's the save: [phobos-plunge.zip](https://github.com/mockingbirdnest/Principia/files/3388983/phobos-plunge.zip)

> but it's not easy to fix: on the one hand we cannot trust KSP to do the right thing, on the other hand we cannot afford to take the body shape into account in the middle of the force computation.

Fair enough. This is not really much of an issue for me, I just wanted to visit Phobos and Deimos once. I'll keep away from space potatoes for now.

As a workaround , are there any rules where I can and can't safely fly?

pleroy:
We detect a collision at 90% of the mean radius (this is working fine for just about everything, except the Mars satellites).  The mean radius of Deimos is 6.2 km and that of Phobos is 11.08 km.  So around 10 km you should be fine.

[...]

PR #2298, opened 2019-08-24
Propagate the min and max radius from the game

pleroy:
And use the min radius for collision detection.  This addresses the collision issues in #2237.


physics/ephemeris_body.hpp
```diff
@@ -68,9 +68,6 @@ using ::std::placeholders::_3;
 constexpr Length pre_ἐρατοσθένης_default_ephemeris_fitting_tolerance =
     1 * Milli(Metre);
 constexpr Time max_time_between_checkpoints = 180 * Day;
-// Below this threshold detect a collision to prevent the integrator and the
-// downsampling from going postal.
-constexpr double mean_radius_tolerance = 0.9;
```
eggrobin: Let's keep a tolerance so that falling a little too fast doesn't cause Princpia to vaporize the spacecraft; 0.99 should be good enough (if you are going fast enough that your centre of mass would be 1 % of the way to the centre of the body after one frame, survival is unlikely).

Issue #2297, opened 2019-08-24
Strange missing redistributable VC++ runtime error


kc9jud:
Possibly related: #1790 and #1066

I'm getting odd errors about the Visual C++ Redistributable Packages not being found -- I can clearly see in Apps & features that it's installed:

![image](https://user-images.githubusercontent.com/4133644/63633465-ee3bfe00-c616-11e9-80a2-ab11ddd8400a.png)

However, I'm getting the following errors in KSP.log:
```
[LOG 02:21:38.443] [UIMasterController]: HideUI
[LOG 02:21:38.857] Loading Depletion Nodes
[LOG 02:21:38.857] DepNodeCount:  0
[LOG 02:21:38.857] Loading Biome Nodes
[LOG 02:21:38.857] BiomeNodeCount:  0
[LOG 02:21:38.857] Loading Planet Nodes
[LOG 02:21:38.857] PlanetNodeCount:  0
[LOG 02:21:38.868] [ScenarioDestructibles]: Loading... 0 objects registered
[EXC 02:21:38.902] DllNotFoundException: GameData/Principia/x64/principia.dll
	principia.ksp_plugin_adapter.Log.InitGoogleLogging ()
	principia.ksp_plugin_adapter.Loader.LoadPrincipiaDllAndInitGoogleLogging ()
	UnityEngine.Debug:LogException(Exception)
	principia.ksp_plugin_adapter.Loader:LoadPrincipiaDllAndInitGoogleLogging()
	principia.ksp_plugin_adapter.PrincipiaPluginAdapter:.ctor()
	UnityEngine.GameObject:AddComponent(Type)
	ScenarioRunner:AddModule(String)
	ScenarioRunner:AddModule(ConfigNode)
	ProtoScenarioModule:Load(ScenarioRunner)
	ScenarioRunner:LoadModules(List`1)
	ScenarioRunner:SetProtoModules(List`1)
	Game:Load()
	<Start>c__Iterator0:MoveNext()
	UnityEngine.SetupCoroutine:InvokeMoveNext(IEnumerator, IntPtr)
[ERR 02:21:38.902] The Principia DLL failed to load.
Dependencies, namely the Visual C++ Redistributable Packages for Visual Studio 2017 on x64, were not found.

[ERR 02:21:39.016] Exception loading ScenarioModule PrincipiaPluginAdapter: System.DllNotFoundException: GameData/Principia/x64/principia.dll
[...]

PR #2299, opened 2019-08-24
Improve error reporting for bad installations

pleroy:
* We lost the "bad installation" dialog because its state was incorrectly restored from the save.
* The VC++ redistributable version should mention the unified 2015-2019 runtime.

This was prompted by #2297.

### DIFF
--- a/ksp_plugin_adapter/dialog.cs
+++ b/ksp_plugin_adapter/dialog.cs
@@ -26,7 +26,7 @@ internal class Dialog : UnsupervisedWindowRenderer, IConfigNode {
     UnityEngine.GUI.DragWindow();
   }
 
-  public new void Load(ConfigNode node) {
+  public override void Load(ConfigNode node) {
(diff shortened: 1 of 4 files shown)


ksp_plugin_adapter/dialog.cs
```diff
@@ -12,22 +12,23 @@ internal class Dialog : UnsupervisedWindowRenderer, IConfigNode {
...
+                                         UnityEngine.GUI.skin.textArea));
     }
     UnityEngine.GUI.DragWindow();
   }
 
   public new void Load(ConfigNode node) {
+    // Whether a dialog is shown or not should not be a persisted property.
+    // It's convenient to save and restore show_ in the base class, but we
+    // override this behaviour here.  For the same reason, we don't persist the
+    // message.
```
eggrobin: Don't we need to persist the message for the apocalypse?

PR #2301, opened 2019-08-24
Fix the mac build

eggrobin:


ksp_plugin/interface.cpp
```diff
@@ -164,12 +164,12 @@ serialization::OblateBody::Geopotential MakeGeopotential(
     int const order = std::stoi(element.order);
     column.set_order(order);
     if (element.j != nullptr) {
-      CHECK_EQ(element.cos, nullptr);
+      CHECK(element.cos == nullptr);
```
pleroy: Why do we need this change?

PR #2312, opened 2019-08-28
Use C# 8 and the nullable reference type specification in the adapter

pleroy:


ksp_plugin_adapter/class_marshalers.cs
```diff
@@ -14,18 +13,18 @@ internal class InBodyParametersMarshaler : ICustomMarshaler {
...
-      public string mean_radius;
-      public string max_radius;
-      public string axis_right_ascension;
-      public string axis_declination;
-      public string reference_angle;
-      public string angular_frequency;
-      public string reference_radius;
-      public string j2;
+      public string? name;
+      public string? gravitational_parameter;
```
eggrobin: `name` and `gravitational_parameter` are `required` in the proto, why are they optional here?

pleroy: Because of the default construction at line 37, which is hard to avoid.

ksp_plugin_adapter/class_marshalers.cs
```diff
@@ -48,8 +47,7 @@ internal class BodyParametersRepresentation {
     }
 
     public IntPtr MarshalManagedToNative(object managed_object) {
-      var parameters = managed_object as BodyParameters;
-      Debug.Assert(parameters != null, nameof(parameters) + " != null");
+      var parameters = (managed_object as BodyParameters)!;
```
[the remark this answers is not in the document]
pleroy: Done.

ksp_plugin_adapter/class_marshalers.cs
```diff
@@ -70,7 +68,7 @@ internal class BodyParametersRepresentation {
       } else {
         int sizeof_element = Marshal.SizeOf(typeof(BodyGeopotentialElement));
         representation.geopotential = Marshal.AllocHGlobal(
-            sizeof_element * parameters.geopotential.Length);
+            sizeof_element * parameters.geopotential!.Length);
```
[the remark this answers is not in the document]
pleroy: Done.

ksp_plugin_adapter/flight_planner.cs
```diff
@@ -67,7 +67,7 @@ class FlightPlanner : SupervisedWindowRenderer {
     // The UI code proper, executed identically for Layout and Repaint.  We
     // can freely change the state in events like clicks (e.g., in if statements
     // for buttons) as these don't happen between Layout and Repaint.
-    string vessel_guid = vessel_?.id.ToString();
+    string? vessel_guid = vessel_?.id.ToString();
     if (vessel_guid == null || !plugin.HasVessel(vessel_guid)) {
```
eggrobin: put `vessel_ == null` in this conditional, so that we don't need the `!` in `FlightPlanCreate` below.

If we want to avoid repeating null checks or `.id.ToString()`, one option is
```C#
string vessel_guid;
if (vessel_ == null || !plugin.HasVessel(vessel_guid = vessel.id.ToString())) {
  return;
}
```
which convinces the compiler that `vessel` is non-null and `vessel_guid` is initialized after the `if` statement.

ksp_plugin_adapter/flight_planner.cs
```diff
@@ -77,7 +77,7 @@ class FlightPlanner : SupervisedWindowRenderer {
     } else if (UnityEngine.GUILayout.Button("Create flight plan")) {
       plugin.FlightPlanCreate(vessel_guid,
                               plugin.CurrentTime() + 1000,
-                              vessel_.GetTotalMass());
+                              vessel_!.GetTotalMass());
```
eggrobin: Remove, see above.

ksp_plugin_adapter/flight_planner.cs
```diff
@@ -105,7 +105,7 @@ class FlightPlanner : SupervisedWindowRenderer {
     }
 
     if (burn_editors_ == null) {
-      string vessel_guid = vessel_?.id.ToString();
+      string? vessel_guid = vessel_?.id.ToString();
       if (vessel_guid != null &&
           plugin.HasVessel(vessel_guid) &&
           plugin.FlightPlanExists(vessel_guid)) {
```
eggrobin: Same dance, put `vessel_ != null` in that condition.

ksp_plugin_adapter/flight_planner.cs
```diff
@@ -117,7 +117,7 @@ class FlightPlanner : SupervisedWindowRenderer {
           // Dummy initial time, we call |Reset| immediately afterwards.
           burn_editors_.Add(
               new BurnEditor(adapter_,
-                             vessel_,
+                             vessel_!,
```
eggrobin: Remove, see above.

ksp_plugin_adapter/flight_planner.cs
```diff
@@ -128,7 +128,7 @@ class FlightPlanner : SupervisedWindowRenderer {
     }
 
     if (burn_editors_ != null) {
-      string vessel_guid = vessel_?.id.ToString();
+      string vessel_guid = vessel_!.id.ToString();
```
eggrobin: Can we have a comment explaining why `vessel_` cannot be null here?

I find this function very confusing in general, perhaps it should be split in a future pull request.

ksp_plugin_adapter/flight_planner.cs
```diff
@@ -285,7 +285,7 @@ class FlightPlanner : SupervisedWindowRenderer {
           }
           var editor =
               new BurnEditor(adapter_,
-                             vessel_,
+                             vessel_!,
```
eggrobin: If I understand correctly, `burn_editors_!` and `vessels_!` come from preconditions of this function; add a comment explaining that.

ksp_plugin_adapter/flight_planner.cs
```diff
@@ -312,7 +312,7 @@ class FlightPlanner : SupervisedWindowRenderer {
   }
 
   private void RenderUpcomingEvents() {
-    string vessel_guid = vessel_.id.ToString();
+    string vessel_guid = vessel_!.id.ToString();
```
eggrobin: Also a precondition?

ksp_plugin_adapter/flight_planner.cs
```diff
@@ -412,7 +412,7 @@ class FlightPlanner : SupervisedWindowRenderer {
   internal string FormatPlanLength(double value) {
     return FormatPositiveTimeSpan(TimeSpan.FromSeconds(
                value -
-               plugin.FlightPlanGetInitialTime(vessel_.id.ToString())));
+               plugin.FlightPlanGetInitialTime(vessel_!.id.ToString())));
```
eggrobin: Another precondition?

ksp_plugin_adapter/flight_planner.cs
```diff
@@ -421,7 +421,7 @@ class FlightPlanner : SupervisedWindowRenderer {
       return false;
     }
     value = ts.TotalSeconds +
-            plugin.FlightPlanGetInitialTime(vessel_.id.ToString());
+            plugin.FlightPlanGetInitialTime(vessel_!.id.ToString());
```
eggrobin: Precondition?

ksp_plugin_adapter/flight_planner.cs
```diff
@@ -470,7 +470,7 @@ class FlightPlanner : SupervisedWindowRenderer {
                          "centre of a celestial)" + time_out_message;
         remedy_message = "avoiding collisions with a celestial";
       } else if (status_.is_invalid_argument()) {
-        status_message = "manœuvre #" + (first_error_manœuvre_.Value + 1) +
+        status_message = "manœuvre #" + (first_error_manœuvre_!.Value + 1) +
```
eggrobin: Comment why this is not null.

ksp_plugin_adapter/flight_planner.cs
pleroy: Added first order logic annotations.

ksp_plugin_adapter/ksp_plugin_adapter.cs
```diff
@@ -1603,7 +1606,7 @@ private enum UnityLayers {
       // Texture the ball.
       navball_changed_ = false;
       if (plotting_frame_selector_.target_override) {
-        set_navball_texture(target_navball_texture_);
+        set_navball_texture(target_navball_texture_!);
```
eggrobin: Comment that the textures are set in `OnAwake` (we could write lazy-load properties for them so that things are clearer).

ksp_plugin_adapter/ksp_plugin_adapter.cs
```diff
@@ -2058,12 +2061,12 @@ private enum UnityLayers {
...
       IntPtr plugin,
-      ConfigNode numerics_blueprint) {
+      ConfigNode? numerics_blueprint) {
     if (numerics_blueprint == null) {
       return;
     }
-    ConfigNode ephemeris_parameters =
-        numerics_blueprint.GetAtMostOneNode("ephemeris");
+    ConfigNode? ephemeris_parameters =
+        numerics_blueprint!.GetAtMostOneNode("ephemeris");
```
eggrobin: Is this `!` necessary with the null check above?

ksp_plugin_adapter/ksp_plugin_adapter.cs
```diff
@@ -2121,7 +2124,7 @@ private enum UnityLayers {
                                       node => node.GetUniqueValue("name"));
       BodyProcessor insert_body = body => {
         Log.Info("Inserting " + body.name + "...");
-        if (!name_to_gravity_model.TryGetValue(
+        if (!name_to_gravity_model!.TryGetValue(
```
eggrobin: Change lines 2111 .. 2113 from
```C#
if (name_to_gravity_model == null) {
  Log.Fatal("Cartesian config without gravity models");
}
```
to
```C#
if (name_to_gravity_model == null) {
  throw Log.Fatal("Cartesian config without gravity models");
}
```
so that control flow analysis understands that `name_to_gravity_model` is not null here.

ksp_plugin_adapter/ksp_plugin_adapter.cs
```diff
@@ -2162,13 +2165,13 @@ private enum UnityLayers {
...
         Log.Info("Inserting " + body.name + "...");
-        ConfigNode body_gravity_model = null;
+        ConfigNode? body_gravity_model = null;
         if (name_to_gravity_model?.TryGetValue(
                 body.name,
                 out body_gravity_model) == true) {
           Log.Info("using custom gravity model");
         }
-        Orbit orbit = unmodified_orbits_.GetValueOrNull(body);
+        Orbit? orbit = unmodified_orbits_!.GetValueOrNull(body);
```
eggrobin: Comment that `unmodified_orbits_` has been set in `OnAwake` (we should probably create the dictionary at construction so that `unmodified_orbits_` is not null, and just fill it in `OnAwake`, but we can clean that up later).

ksp_plugin_adapter/window_renderer.cs
```diff
@@ -114,7 +114,7 @@ internal abstract class BaseWindowRenderer : ScalingRenderer, IConfigNode {
     var old_skin = UnityEngine.GUI.skin;
     if (skin_ == null) {
       UnityEngine.GUI.skin = null;
-      skin_ = MakeSkin(UnityEngine.GUI.skin);
+      skin_ = MakeSkin(UnityEngine.GUI.skin!);
```
eggrobin: This is a case where we would really need to annotate the engine declarations:
the property `UnityEngine.GUI.skin` is `UnityEngine.GUISkin`, not `UnityEngine.GUISkin?`, but it is `[AllowNull]` (`set` accepts null, `get` never returns null).

PR #2320, opened 2019-09-01
Modernize the code of our scripts

pleroy:


parallel_test_runner/parallel_test_runner.cs
```diff
@@ -21,12 +21,11 @@ enum Granularity {
     return (T)Enum.Parse(typeof(T), value, true);
   }
 
-  const string vsinstr =
+  const string vsinstr_ =
```
eggrobin: `const` in C# is more like `static const` or `constexpr` in C++, do we really want a trailing underscore?

PR #2322, opened 2019-09-02
Use structured bindings

eggrobin:
### Iteration on maps
This changes most instances of
```C++
for (auto const& pair : ⟨map⟩)
```
to
```C++
for (auto const& [key, value] : ⟨map⟩)
```
The exceptions are
https://github.com/mockingbirdnest/Principia/blob/a8cfa72ed7356caee03945da2f68e9c253bcefa2/physics/solar_system_body.hpp#L636-L638
where the name `degrees_of_freedom` is taken,
and the public functions of the journal proto processor:
https://github.com/mockingbirdnest/Principia/blob/a8cfa72ed7356caee03945da2f68e9c253bcefa2/tools/journal_proto_processor.cpp#L112-L188
In both cases, switching to structured bindings felt clumsier than the existing code. It is noteworthy that both are examples of the same pattern, turning the values of a map into a vector; perhaps we could write an utility function for that.

#### `_` for unused keys

Where the key is unused, this pull request uses
```C++
for (auto const& [_, value] : ⟨map⟩)
```
In the future, we will probably want to introduce an iterator adapter `ValuesIn` and change these usages to
```C++
for (auto const& value : ValuesIn(⟨map⟩))
```
Iterator adapters would also make it possible to turn usages of `Vessel::ForAllParts` into actual loops.

### `insert` and `emplace`

use `bool const inserted = ⟨container⟩.insert(...).second` if only the boolean is used, reusing an existing `bool inserted` if there are scoping issues; use structured bindings if both are used (where previously either `std::tie` or references into the pair would be used).

One usage of `std::tie` remains, which is a proper tie of pre-existing variables:
https://github.com/mockingbirdnest/Principia/blob/c5a4878f17ddbb49bd1741dd7ca298a0ddbe6594/ksp_plugin/plugin.cpp#L365-L386

Note that in the future, we should probably change `LOG_IF` in our glog fork to use `if` rather than `?:`, so that it is possible to narrowly scope `inserted`, thus
```C++
CHECK(bool const inserted = container.insert(...).second; inserted) << ...;
```

### DIFF
--- a/ksp_plugin/interface.cpp
+++ b/ksp_plugin/interface.cpp
@@ -180,7 +180,7 @@ serialization::OblateBody::Geopotential MakeGeopotential(
   }
 
   serialization::OblateBody::Geopotential result;
-  for (auto const& [degree, row] : rows) {
+  for (auto const& [_, row] : rows) {
     *result.add_row() = row;
   }
   return result;
--- a/ksp_plugin/interface.cpp
+++ b/ksp_plugin/interface.cpp
@@ -180,8 +180,8 @@ serialization::OblateBody::Geopotential MakeGeopotential(
   }
 
   serialization::OblateBody::Geopotential result;
-  for (auto const& pair : rows) {
-    *result.add_row() = pair.second;
+  for (auto const& [degree, row] : rows) {
+    *result.add_row() = row;
   }
   return result;
 }
--- a/ksp_plugin/plugin.cpp
+++ b/ksp_plugin/plugin.cpp
@@ -1323,9 +1323,9 @@ not_null<std::unique_ptr<Plugin>> Plugin::ReadFromMessage(
     if (vessel_message.kept()) {
       plugin->kept_vessels_.insert(vessel.get());
     }
-    auto const inserted =
+    auto const [it, inserted] =
         plugin->vessels_.emplace(vessel_message.guid(), std::move(vessel));
-    CHECK(inserted.second);
+    CHECK(inserted);
   }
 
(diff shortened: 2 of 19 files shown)


ksp_plugin/interface.cpp
```diff
@@ -180,8 +180,8 @@ serialization::OblateBody::Geopotential MakeGeopotential(
   }
 
   serialization::OblateBody::Geopotential result;
-  for (auto const& pair : rows) {
-    *result.add_row() = pair.second;
+  for (auto const& [degree, row] : rows) {
```
pleroy: I am wondering if we should write:
```
for (auto const& [_, row] : rows) {
```
in the case where one of the elements is not used.  That would maybe make the code more readable (otherwise I am left wondering where `degree` is used).

ksp_plugin/plugin.cpp
```diff
@@ -482,12 +480,8 @@ void Plugin::IncrementPartIntrinsicForce(PartId const part_id,
 }
 
 void Plugin::PrepareToReportCollisions() {
-  for (auto const& pair : vessels_) {
-    Vessel& vessel = *pair.second;
-    // TODO(egg): we're taking the address of a parameter passed by reference
-    // here; but then I don't think I want to pass this by pointer, it's quite
-    // convenient everywhere else...
```
pleroy: I would keep the comment.

eggrobin: Changed it to a note and documented the lifetime requirements in various places.

PR #2327, opened 2019-09-09
Support for diagonalizing a symmetric bilinear form

pleroy:


geometry/symmetric_bilinear_form_body.hpp
```diff
@@ -42,6 +55,64 @@ SymmetricBilinearForm<Scalar, Frame>::operator()(
...
+  // The vectors (v₀, v₁, v₂) forms an orthonormal basis.  Make sure that it is
+  // direct.
+  auto v₂ = PickEigenvector(m₂);
+  if (R3x3Matrix(v₀, v₁, v₂).Determinant() < 0) {
+    v₂ = -v₂;
+  }
+  Rotation<Frame, Eigenframe> const rotation{
+      Vector<double, Frame>(v₀),
+      Vector<double, Frame>(v₁),
+      Bivector<double, Frame>(v₂)};  // This bivector is a bit dodgy.
```
eggrobin: The bivector is not dodgy, the if statement above where we manually mess with the orientation is; the third argument here should just be `Wedge(v₀, v₁)`, we do not need `v₂`.

pleroy: Done.

PR #2329, opened 2019-09-14
Separate the nominal recurrence from OrbitGroundTrack

eggrobin:
The slow part of OrbitGroundTrack is the computation of the nodes; analysing how well pre-computed nodes follow a particular nominal recurrence grid is fast enough that it can be done in the UI thread, and this makes the analyser more convenient to use.

### DIFF
--- a/astronomy/orbit_ground_track.hpp
+++ b/astronomy/orbit_ground_track.hpp
@@ -35,30 +35,40 @@ class OrbitGroundTrack {
     Time year;
   };
 
+  class EquatorCrossingLongitudes {
+   public:
+    // The longitudes of the equator crossings (ascending or descending,
+    // depending on the parity of pass_index), reduced to a grid interval around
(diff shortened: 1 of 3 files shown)


pleroy:
retest this please

astronomy/orbit_ground_track.hpp
```diff
@@ -35,30 +36,40 @@ class OrbitGroundTrack {
...
   };
 
+  class EquatorCrossingLongitudes {
+   public:
+    // The longitudes of the equator crossings (ascending or descending,
+    // depending on the parity of pass_index), reduced to a grid interval around
+    // the pass with the given index.  This provides both an indication of how
+    // well the orbit follows the nominal recurrence grid, and where that grid
+    // is located.
+    Interval<Angle> longitudes_reduced_to_pass(int const pass_index) const;
```
pleroy: Remove `const`.

PR #2332, opened 2019-09-18
Better error handling in the stack decoder, again

pleroy:


stacktrace_decoder/stacktrace_decoder.cs
```diff
@@ -167,10 +167,13 @@ class StackTraceDecoder {
     IntPtr handle = new IntPtr(1729);
     SymSetOptions(SYMOPT_LOAD_LINES);
     Win32Check(SymInitializeW(handle, null, fInvadeProcess: false));
+    string principia_dll_path =
+        Path.Combine(principia_directory, "principia.dll");
+    Win32Check(File.Exists(principia_dll_path));
```
eggrobin: `File.Exists` is a .NET function, not a direct call to the Windows API (that would be `PathFileExistsW` or something along those lines); I don't expect `GetLastError` to tell us anything useful when it returns.

This should be `Check`.

pleroy: You may not expect it, but I tested this and it returns 2 which is [`ERROR_FILE_NOT_FOUND`](https://docs.microsoft.com/en-us/windows/win32/debug/system-error-codes--0-499-).

PR #2336, opened 2019-09-23
Minor cleanups to the lunar eclipse test

eggrobin:
We have an equatorial radius now.

There is a problem with the IMCCE link: the relevant page moved to https://promenade.imcce.fr/fr/pages3/333.html, but the relevant section *Correction de la parallaxe lunaire dans les calculs d'éclipse de Lune* was removed in the process. Should we link to http://web.archive.org/web/20170915031958/http://www.imcce.fr:80/langues/fr/ephemerides/phenomenes/eclipses/lune/?

### DIFF
--- a/astronomy/lunar_eclipse_test.cpp
+++ b/astronomy/lunar_eclipse_test.cpp
@@ -34,6 +34,7 @@ using quantities::ArcSin;
 using quantities::Length;
 using quantities::Sin;
 using quantities::Time;
+using quantities::astronomy::TerrestrialEquatorialRadius;
 using quantities::si::Day;
 using quantities::si::Kilo;
 using quantities::si::Metre;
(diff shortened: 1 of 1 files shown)


astronomy/lunar_eclipse_test.cpp
```diff
@@ -78,7 +79,8 @@ class LunarEclipseTest : public ::testing::Test {
     // and http://www.imcce.fr/langues/fr/ephemerides/phenomenes/eclipses/lune/.
     // In particular, this means that we must use the equatorial radius for the
     // Earth, not the mean radius.
-    r_earth_ = 6378.1363 * Kilo(Metre);
+    // REMOVE BEFORE FLIGHT: what do we do with the dead link?
```
pleroy: Let's use the `archive.org` link.

PR #2337, opened 2019-09-24
A window that shows mean orbital elements for the active vessel

eggrobin:
Example with 100 days on a Молния orbit:
screenshot28

PR #2338, opened 2019-09-25
Expose the orbit analyser through Vessel and the interface

eggrobin:
This is the C++ backing for #2337.


ksp_plugin/interface_body.hpp
```diff
@@ -167,6 +172,53 @@ inline bool operator==(NavigationManoeuvreFrenetTrihedron const& left,
...
+                       EquatorialCrossings const& right) {
+  return left.longitudes_reduced_to_ascending_pass ==
+             right.longitudes_reduced_to_ascending_pass &&
+         left.longitudes_reduced_to_descending_pass ==
+             right.longitudes_reduced_to_descending_pass;
+}
+
+inline bool operator==(OrbitGroundTrack const& left,
+                       OrbitGroundTrack const& right) {
+  return left.equatorial_crossings ==
```
pleroy: Would fit on one line.

serialization/journal.proto
```diff
@@ -155,8 +155,65 @@ message XY {
   required double y = 2;
 }
 
+message Interval {
```
pleroy: Document the ordering like is done at line 13-14.

serialization/journal.proto
```diff
@@ -155,8 +155,65 @@ message XY {
...
+  required double nodal_precession = 4;
+  required Interval mean_semimajor_axis = 5;
+  required Interval mean_eccentricity = 6;
+  required Interval mean_inclination = 7;
+  required Interval mean_longitude_of_ascending_nodes = 8;
+  required Interval mean_argument_of_periapsis = 9;
+}
+
+message OrbitRecurrence {
+  // νₒ.
```
pleroy: Comment at end of line?

serialization/journal.proto
```diff
@@ -155,8 +155,65 @@ message XY {
...
+  required EquatorialCrossings equatorial_crossings = 1;
+  // TODO(egg): mean solar times of the nodes.
+}
+
+message OrbitAnalysis {
+  required double progress_of_next_analysis = 1;
+  required int32 primary_index = 2;
+  required double mission_duration = 3;
+  // NOTE(egg): we currently have no support in the interface generator for
+  // optional non-string fields in interchange structs.  We use separate
```
pleroy: Why don't we fix the generator instead of doing this horrible hack?

ksp_plugin/vessel.hpp
```diff
@@ -173,6 +176,11 @@ class Vessel {
       PileUp::PileUpForSerializationIndex const&
           pile_up_for_serialization_index);
 
+  void RefreshOrbitAnalysis(not_null<RotatingBody<Barycentric> const*> primary,
+                            Time const& mission_duration);
+  double progress_of_orbit_analysis() const;
```
pleroy: Blank line before.

serialization/journal.proto
eggrobin: Fixing the generator is exceedingly complicated; we cannot have an `std::optional<T>` on the C++ side as it does not have standard layout, so we would need to define our own optional type. On the C# side, we would need to marshal the `T?` completely by hand.

PR #2341, opened 2019-09-28
Fix problems found on Linux and enable strict conformance for MSVC

pleroy:


numerics/polynomial.hpp
```diff
@@ -23,6 +23,14 @@ using quantities::Derivatives;
 using quantities::Product;
 using quantities::Quotient;
 
+// In Visual Studio 2019 16.2.3 using std::max directly below causes strange
+// ambiguity.  Possibly related to
+// https://developercommunity.visualstudio.com/content/problem/388596/compilation-fails-with-permissive.html.  // NOLINT
```
eggrobin: No `NOLINT`, the linter recognizes URLs.

PR #2359, opened 2019-10-19
Bring the notation in elliptic_integrals.{hpp,cpp} in line with the papers

eggrobin:
They were an odd mix between a casefolding of the papers and remnants of the FORTRAN, leading to such things as `bc` for “𝐵 complete” next to `Bc` for 𝐵<sub>𝑐</sub>, j for 𝐽 where the paper calls an index 𝑗, etc.

Also align the Biber/XeLaTeX bibliography with the one in the header file.


numerics/elliptic_integrals.hpp
```diff
@@ -2,52 +2,87 @@
...
-Angle EllipticF(Angle const& φ,
-                double mc);
+// Returns the incomplete elliptic integral of the second kind E(φ|m), where
+// m = 1 - mc.
+Angle EllipticE(Angle const& φ, double mc);
 
-Angle EllipticΠ(Angle const& φ,
-                double n,
-                double mc);
+// Returns the incomplete elliptic integral of the third kind Π(φ, n|m), where
```
pleroy: Often, *n* appears first in the list of parameters of Π, followed by a semicolon.  But the notation varies...

numerics/elliptic_integrals.cpp
```diff
@@ -44,54 +34,64 @@ using quantities::si::Radian;
...
-                             Angle& d,
-                             Angle& j);
+                             Angle& Bs_sǀm,
+                             Angle& Ds_sǀm,
+                             Angle& Js_s_nǀm);
 
-// Maclaurin series expansion of Bs and Ds [Fuku11a].
+// Computes the Maclaurin series expansion ∑ Bₗ(m) yˡ and ∑ Dₗ(m) yˡ used in the
+// computation of of Bs and Ds, see [Fuk11b], equation (15).
 // NOTE(phl): I believe that this is a Maclaurin series but it's not completely
```
pleroy: Remove that note, maybe.

numerics/elliptic_integrals.cpp
```diff
@@ -1214,109 +1198,106 @@ double EllipticNomeQ(double const mc) {
...
-    d = ((2.0 * log_2 - 1.0) - 0.5 * std::log(mc)) * Radian;
+    B_m = 1.0 * Radian;
+    D_m = ((2.0 * log_2 - 1.0) - 0.5 * std::log(mc)) * Radian;
   } else if (mc < 0.1) {
-    // This algorithm differs from the one in [Fuku11a] because it divides
-    // log(q(mc)), not just log(mc / 16).  It tries to retain the same notation,
-    // though.
+    // This algorithm (from [Fuk18]) differs from the one in [Fuk11a] because it
+    // divides log(q(mc)), not just log(mc / 16).  It tries to retain the same
+    // notation, though.
```
pleroy: Point at our documentation.

numerics/elliptic_integrals.cpp
```diff
@@ -1435,158 +1429,163 @@ void FukushimaEllipticBsDsMaclaurinSeries(double const y,
...
-  double const js4 = fukushima_elliptic_js_maclaurin_m_4.Evaluate(m);
+      fukushima_elliptic_Js_maclaurin_m_4(
+          std::make_tuple(fukushima_elliptic_Js_maclaurin_n_4_0.Evaluate(n),
+                          fukushima_elliptic_Js_maclaurin_n_4_1.Evaluate(n),
+                          fukushima_elliptic_Js_maclaurin_n_4_2.Evaluate(n),
+                          fukushima_elliptic_Js_maclaurin_n_4_3.Evaluate(n),
+                          fukushima_elliptic_Js_maclaurin_n_4_4.Evaluate(n)));
+  // The first five coefficients Jₗ(n|m); the others are computed as needed.
+  // TODO(egg): J₁(n|m) = 1/3; could we make it constexpr?
+  double const J₁ = fukushima_elliptic_Js_maclaurin_m_0.Evaluate(m);
```
pleroy: The change of index is confusing because J1 is for m^0 now.  Could we go back to 0-based?

numerics/elliptic_integrals.cpp
```diff
@@ -1714,137 +1699,144 @@ void FukushimaEllipticBDJ(Angle const& φ,
...
+// those in the function declaration.
+// The notation here follows [Fuk09], whereas the notation in the function
+// declaration uses |mc| for consistency with the other functions.
+Angle EllipticK(double const mʹ) {
+  DCHECK_LE(0, mʹ);
+  DCHECK_GE(1, mʹ);
   // TODO(phl): Use a binary split of [0, 1] to reduce the number of
   // comparisons.
-  double const m = 1.0 - mc;
+  double const m = 1.0 - mʹ;
```
pleroy: I don't think it's useful to swtch from `mc` to `mʹ` for this function.  As I recall this is the only place where Fukushima uses this notation, and the notation is in general messy and inconsistent enough that we shouldn't add to the mess.

documentation/bibliography.bib
```diff
@@ -780,6 +780,48 @@ @TechReport{Westra2017
 month = sep
 }
 
+@Article{Bulirsch1965,
+author="Bulirsch, Roland",
```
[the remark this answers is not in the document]
eggrobin: In the generated PDF, we do; in the .bib file, it is best to have as much information as possible, in case we want to switch to an unabbreviated style.

numerics/elliptic_integrals.cpp
eggrobin: J<sub>1</sub> is for y<sup>1</sup> in the result; this is 0-based, there just is no constant term.
This is the convention used in the paper.

Your indexing led to -1-based indices in the coefficient of the series.

numerics/elliptic_integrals.cpp
eggrobin: We mention K′ below (in the existing comment as well), so we use the notation from [Fuk09] here. If we start mixing notations between papers we are going to get a mess: calling K′ Kc would give its c a different semantic than in Bc.

numerics/elliptic_integrals.hpp
eggrobin: Fukushima uses Π(φ, n|m).

numerics/elliptic_integrals.cpp
eggrobin: Done.

numerics/elliptic_integrals.cpp
eggrobin: Done.

PR #2362, opened 2019-10-23
Templatize on the need for the integral of the third kind

eggrobin:
This makes it possible to compute F and E without computing J when Π is not needed.

Before:
```
Run on (4 X 2808 MHz CPU s)
CPU Caches:
  L1 Data 32K (x2)
  L1 Instruction 32K (x2)
  L2 Unified 262K (x2)
  L3 Unified 4194K (x1)
------------------------------------------------------
Benchmark               Time           CPU Iterations
------------------------------------------------------
BM_EllipticF          233 ns        228 ns    2951600
BM_EllipticFEΠ        235 ns        230 ns    2992000
```
After:
```
Run on (4 X 2808 MHz CPU s)
CPU Caches:
  L1 Data 32K (x2)
  L1 Instruction 32K (x2)
  L2 Unified 262K (x2)
  L3 Unified 4194K (x1)
------------------------------------------------------
Benchmark               Time           CPU Iterations
------------------------------------------------------
BM_EllipticF          133 ns        123 ns    4072800
BM_EllipticFEΠ        239 ns        230 ns    2992000
```

### DIFF
--- a/benchmarks/elliptic_integrals_benchmark.cpp
+++ b/benchmarks/elliptic_integrals_benchmark.cpp
@@ -31,7 +31,7 @@ void BM_EllipticF(benchmark::State& state) {
     mcs.push_back(distribution_mc(random));
   }
 
-  while (state.KeepRunningBatch(size * size * size)) {
+  while (state.KeepRunningBatch(size * size)) {
     Angle f;
     for (Angle const φ : φs) {
       for (double const mc : mcs) {
--- a/benchmarks/elliptic_integrals_benchmark.cpp
+++ b/benchmarks/elliptic_integrals_benchmark.cpp
@@ -23,24 +23,19 @@ void BM_EllipticF(benchmark::State& state) {
 
   std::mt19937_64 random(42);
   std::uniform_real_distribution<> distribution_φ(0.0, π / 2);
-  std::uniform_real_distribution<> distribution_n(0.0, 1.0);
   std::uniform_real_distribution<> distribution_mc(0.0, 1.0);
   std::vector<Angle> φs;
-  std::vector<double> ns;
   std::vector<double> mcs;
   for (int i = 0; i < size; ++i) {
     φs.push_back(distribution_φ(random) * Radian);
-    ns.push_back(distribution_n(random));
     mcs.push_back(distribution_mc(random));
   }
 
   while (state.KeepRunningBatch(size * size * size)) {
     Angle f;
     for (Angle const φ : φs) {
-      for (double const n : ns) {
-        for (double const mc : mcs) {
-          f += EllipticF(φ, mc);
-        }
+      for (double const mc : mcs) {
+        f += EllipticF(φ, mc);
       }
     }
     benchmark::DoNotOptimize(f);
--- a/benchmarks/elliptic_integrals_benchmark.cpp
+++ b/benchmarks/elliptic_integrals_benchmark.cpp
@@ -18,6 +18,35 @@ using quantities::si::Radian;
 
 namespace numerics {
 
+void BM_EllipticF(benchmark::State& state) {
+  constexpr int size = 20;
+
+  std::mt19937_64 random(42);
+  std::uniform_real_distribution<> distribution_φ(0.0, π / 2);
+  std::uniform_real_distribution<> distribution_n(0.0, 1.0);
+  std::uniform_real_distribution<> distribution_mc(0.0, 1.0);
+  std::vector<Angle> φs;
+  std::vector<double> ns;
+  std::vector<double> mcs;
+  for (int i = 0; i < size; ++i) {
+    φs.push_back(distribution_φ(random) * Radian);
+    ns.push_back(distribution_n(random));
+    mcs.push_back(distribution_mc(random));
+  }
+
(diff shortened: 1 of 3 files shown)


benchmarks/elliptic_integrals_benchmark.cpp
```diff
@@ -18,6 +18,35 @@ using quantities::si::Radian;
 
 namespace numerics {
 
+void BM_EllipticF(benchmark::State& state) {
+  constexpr int size = 20;
+
+  std::mt19937_64 random(42);
+  std::uniform_real_distribution<> distribution_φ(0.0, π / 2);
+  std::uniform_real_distribution<> distribution_n(0.0, 1.0);
```
pleroy: We don't need the n's and the loop at line 40 is probably leading to too efficient caching.

numerics/elliptic_integrals.cpp
```diff
@@ -45,39 +45,61 @@ double EllipticNomeQ(double mc);
...
+// without computing the integral of the third kind.  Passing |unused| bypasses
+// the corresponding code path at compile time, passing a reference to an
+// |Angle| performs the computation.
+
+struct UnusedResult {
+  constexpr UnusedResult(base::uninitialized_t) {}
+};
+
+template<typename T, typename = EnableIfAngleResult<T>>
+inline constexpr bool should_compute = !std::is_same_v<T, UnusedResult const>;
```
pleroy: Do we believe that template variables work on Mac?

numerics/elliptic_integrals.cpp
```diff
@@ -45,39 +45,61 @@ double EllipticNomeQ(double mc);
...
+
+inline constexpr UnusedResult unused{uninitialized};
```
pleroy: Same question for inline variables.

numerics/elliptic_integrals.cpp
```diff
@@ -45,39 +45,61 @@ double EllipticNomeQ(double mc);
...
+
+template<typename T>
+using EnableIfAngleResult =
+    std::enable_if_t<std::is_same_v<T, UnusedResult const> ||
+                     std::is_same_v<T, Angle>>;
```
pleroy: I'd reorder things a bit for ease of readability:
1. `UnusedResult`,
1. `unused`,
1. `EnableIfAngleResult`,
1. `should_compute`.

numerics/elliptic_integrals.cpp
```diff
@@ -1744,27 +1802,63 @@ void FukushimaEllipticBDJ(Angle const& φ,
...
 
+}  // namespace
+
+void FukushimaEllipticBDJ(Angle const& φ,
+                          double const n,
+                          double const mc,
+                          Angle& B_φǀm,
+                          Angle& D_φǀm,
+                          Angle& J_φ_nǀm) {
+  return FukushimaEllipticBDJ<Angle>(φ, n, mc, B_φǀm, D_φǀm, J_φ_nǀm);
```
pleroy: Is the specification of `Angle` necessary?  I would have expected that it would deduce it.

numerics/elliptic_integrals.cpp
```diff
@@ -1776,23 +1870,17 @@ Angle EllipticΠ(Angle const& φ, double const n, double const mc) {
...
-  DCHECK_LE(0, mc);
-  DCHECK_GE(1, mc);
-  Angle B{uninitialized};
-  Angle D{uninitialized};
-  Angle J{uninitialized};
-  FukushimaEllipticBDJ(φ, n, mc, B, D, J);
-  F_φǀm = B + D;
-  E_φǀm = B + mc * D;
-  Π_φ_nǀm = F_φǀm + n * J;
+  EllipticFEΠ<Angle>(φ, n, mc, F_φǀm, E_φǀm, Π_φ_nǀm);
```
pleroy: Same question.

numerics/elliptic_integrals.cpp
eggrobin: It would, but the non-templatized `FukushimaEllipticBDJ` (which we are defining) has precedence here.

numerics/elliptic_integrals.cpp
eggrobin: Same answer.

numerics/elliptic_integrals.cpp
eggrobin: An alternative would be to expose `UnusedResult` in `base`, and to only have the templatized version (declared in the header, with both explicit instantiation declarations).

numerics/elliptic_integrals.cpp
eggrobin: Seems to work.

numerics/elliptic_integrals.cpp
eggrobin: Same answer.

numerics/elliptic_integrals.cpp
eggrobin: Done. That also makes it legal, MSVC is overly lenient as usual.

benchmarks/elliptic_integrals_benchmark.cpp
eggrobin: Done.

PR #2372, opened 2019-11-14
Much ado about nothing’s sign bit

eggrobin:


astronomy/trappist_dynamics_test.cpp
```diff
@@ -1029,8 +1029,8 @@ class TrappistDynamicsTest : public ::testing::Test {
       Sign const xy_displacement_derivative_sign(xy_displacement_derivative(t));
       if (relative_dof.displacement().coordinates().z > 0.0 * Metre &&
           last_t &&
-          xy_displacement_derivative_sign == Sign(1) &&
-          last_xy_displacement_derivative_sign == Sign(-1)) {
+          xy_displacement_derivative_sign.is_positive() &&
+          last_xy_displacement_derivative_sign == Sign::Negative()) {
```
pleroy: `is_negative`?

geometry/sign.hpp
```diff
@@ -4,41 +4,65 @@
...
 using base::not_null;
+using quantities::Quantity;
 
 // An element of the multiplicative group ({+1, -1}, *). Useful for instance to
 // represent the determinant of an orthogonal map.
 class Sign final {
  public:
-  template<typename Scalar> explicit Sign(Scalar const& s);
+  explicit Sign(double x);
+  template<typename Dimensions>
```
pleroy: Blank line before.

geometry/sign.hpp
```diff
@@ -4,41 +4,65 @@
...
+  explicit Sign(Quantity<Dimensions> const& x);
+  // The deleted constructor forbids construction from an integer via
```
pleroy: Same.

geometry/sign.hpp
```diff
@@ -4,41 +4,65 @@
...
+  // integer-to-double conversion.  Integers have no signed 0, so this could
+  // lead to confusing behaviour.
+  template<typename T>
+  explicit Sign(T x) = delete;
+
+  static constexpr Sign Positive();
+  static constexpr Sign Negative();
+
+  template<typename T, typename = std::enable_if_t<std::is_arithmetic_v<T>>>
```
pleroy: I'd prefer `is_integral` here.  It's quite weird to call this function with a floating-point type when you have perfectly good constructors to do that.

geometry/sign.hpp
```diff
@@ -4,41 +4,65 @@
...
-  bool Positive() const;
+  constexpr Sign operator+() const;
+  constexpr Sign operator-() const;
 
-  bool operator==(Sign const& other) const;
-  bool operator!=(Sign const& other) const;
+  // Returns ±1.
+  constexpr operator int() const;
+
+  constexpr bool operator==(Sign const& other) const;
```
pleroy: `Sign` should probably be passed by copy.  I mean, it's one bool, what could it cost?  Ten bytes?

geometry/sign.hpp
```diff
@@ -4,41 +4,65 @@
...
-  friend Sign operator*(Sign const& left, Sign const& right);
+
+  friend constexpr Sign operator*(Sign const& left, Sign const& right);
   template<typename T>
-  friend T operator*(Sign const& left, T const& right);
+  friend constexpr T operator*(Sign const& left, T const& right);
 };
 
-Sign operator*(Sign const& left, Sign const& right);
+constexpr Sign operator*(Sign const& left, Sign const& right);
```
pleroy: Why do we have `*` here but `==` and `!=` in the class?  We should probably decide once and for all what we do with binary operators.

geometry/sign_body.hpp
```diff
@@ -11,22 +12,54 @@ namespace principia {
...
 namespace internal_sign {
 
-template<typename Scalar>
-Sign::Sign(Scalar const& scalar) : negative_(scalar < Scalar{}) {}
+using quantities::SIUnit;
 
-inline bool Sign::Negative() const {
-  return negative_;
+// TODO(egg): Consider intrinsics.
+inline Sign::Sign(double x) : negative_(std::signbit(x)) {}
```
pleroy: `const`

geometry/sign_body.hpp
```diff
@@ -36,20 +69,22 @@ inline void Sign::WriteToMessage(
 }
 
 inline Sign Sign::ReadFromMessage(serialization::Sign const& message) {
-  return Sign(message.negative() ? -1 : 1);
+  return Sign(/*negative=*/message.negative());
 }
 
-inline Sign operator*(Sign const& left, Sign const& right) {
-  return Sign(left.negative_ == right.negative_ ? 1 : -1);
+constexpr Sign::Sign(bool negative) : negative_(negative) {}
```
pleroy: `const`

geometry/sign_body.hpp
```diff
@@ -36,20 +69,22 @@ inline void Sign::WriteToMessage(
...
+
+constexpr Sign operator*(Sign const& left, Sign const& right) {
+  return Sign(/*negative=*/left.negative_ != right.negative_);
```
pleroy: Use `^` instead of `!=`.

astronomy/trappist_dynamics_test.cpp
eggrobin: No, that's an optional sign on the left hand side.

geometry/sign.hpp
eggrobin: In general we should definitely revisit our `const&` usage; we pass quantities by `const&`, and that is probably silly as well.

Done.

geometry/sign.hpp
eggrobin: A job for another pull request, we should harmonize `Quantity` with `Multivector` as well.

geometry/sign_body.hpp
eggrobin: `^` is an arithmetic operator, not a logical one, so this requires a cast; from the point of view of readability, do we want that?

In terms of generated code, it does appear to make a difference with MSVC, but only if `^` is applied to signed operands: https://gcc.godbolt.org/z/dGvY55; it makes no difference with clang.

PR #2379, opened 2019-11-24
Get the inertia tensor of the body and push it to the plugin

pleroy:


ksp_plugin/interface.cpp
```diff
@@ -800,30 +817,72 @@ void principia__InsertOrKeepLoadedPart(
...
        delta_t});
   CHECK_NOTNULL(plugin);
+
+  // We build the inertia tensor in the principal axes and then transform it to
+  // World.
+  using PartPrincipalAxes = Frame<serialization::Frame::PhysicsTag,
+                                  serialization::Frame::PRINCIPAL_AXES, false>;
+
+  auto const moments_of_inertia =
+      FromXYZ<R3Element<MomentOfInertia>>(moments_of_inertia_in_tonnes);
```
eggrobin: I'm not really comfortable with a non-SI-unit FromXYZ; use `R3Element<double>` and multiply by the appropriate unit.

pleroy: Done, in a slightly different manner.

PR #2382, opened 2019-11-29
Use a RigidMotion instead of multiple parameters when passing a part to the plugin

pleroy:


ksp_plugin/plugin.cpp
```diff
@@ -461,6 +459,9 @@ void Plugin::InsertOrKeepLoadedPart(
         main_body_frame.FromThisFrameAtTime(previous_time) *
         world_to_main_body_centred;
 
+    DegreesOfFreedom<World> const part_degrees_of_freedom(
+        part_rigid_motion.rigid_transformation()(RigidPart::origin),
+        part_rigid_motion.Inverse().velocity_of_to_frame_origin());
```
eggrobin: Don't handle the velocity and position separately; that's how you get bugs, see https://github.com/mockingbirdnest/Principia/commit/8aa5e56ba7ed384fc92b43d05849c0116e59be72.

`part_rigid_motion(DegreesOfFreedom<RigidPart>{RigidPart::origin, Velocity<RigidPart>{}})`.

We should probably have an `UnmovingOrigin` to construct those specific degrees of freedom more easily, but that's a job for a separate pull request.

pleroy: Klug.

PR #2386, opened 2019-11-30
Propagate the inertia tensor to the Part

pleroy:


ksp_plugin/part.cpp
```diff
@@ -174,13 +179,36 @@ not_null<std::unique_ptr<Part>> Part::ReadFromMessage(
...
+
+  std::unique_ptr<Part> part;
+  if (is_pre_fréchet) {
+    static constexpr MomentOfInertia zero;
+    static constexpr MomentOfInertia one = SIUnit<MomentOfInertia>();
+    InertiaTensor<Barycentric> inertia_tensor(
+        Mass::ReadFromMessage(message.mass()),
+        R3x3Matrix<MomentOfInertia>({one, zero, zero},
+                                    {zero, one, zero},
+                                    {zero, zero, one}),
```
eggrobin: As discussed, construct a reasonable moment of inertia; I think here you have a ball of radius sqrt(5 kg m<sup>2</sup> / (2m)), i.e., a ball that shrinks as it gets heavier!

pleroy: Done.

physics/inertia_tensor.hpp
```diff
@@ -53,11 +56,19 @@ class InertiaTensor {
     Rotation<Frame, PrincipalAxesFrame> rotation;
   };
 
+  // A factory that creates an inertia tensor for a solid sphere of water having
+  // the given mass.  Useful e.g. for save compatibility.
+  static InertiaTensor<Frame> MakeSphericalInertiaTensor(Mass const& mass);
```
eggrobin: Either pass ρ as a parameter, put "Water" in the name, or move this to ksp_plugin.

ksp_plugin/part.cpp
```diff
@@ -228,7 +248,7 @@ std::string Part::ShortDebugString() const {
 std::ostream& operator<<(std::ostream& out, Part const& part) {
   return out << "{"
              << part.part_id() << ", "
-             << part.mass() << "}";
+             << part.inertia_tensor().mass() << "}";
```
eggrobin: Log the tensor too.

pleroy: As discussed, let's not do that, this is not even used.

physics/inertia_tensor_body.hpp
```diff
@@ -84,14 +90,48 @@ InertiaTensor<Frame>::Diagonalize() const {
...
+    Mass const& mass) {
+  static constexpr Density ρ_of_water = 1000 * Kilogram / Pow<3>(Metre);
+  static constexpr MomentOfInertia zero;
+  MomentOfInertia const I =
+      Cbrt(9 * Pow<5>(mass) / (250 * Pow<2>(π * ρ_of_water)));
+  return InertiaTensor<Frame>(mass,
+                              R3x3Matrix<MomentOfInertia>({I, zero, zero},
+                                                          {zero, I, zero},
+                                                          {zero, zero, I}),
+                              Frame::origin);
```
eggrobin: Wait, we are using `InertiaTensor<Barycentric>`, and putting these water spheres at the origin? Shouldn't they be at the origin of the part instead?

Alternatively, shouldn't we use `InertiaTensor<RigidPart>`, so that we don't store the position twice?

pleroy: Done (the latter).

PR #2388, opened 2019-12-01
Propagate the rigid motion to the part

pleroy:


physics/rigid_motion.hpp
```diff
@@ -65,6 +68,14 @@ class RigidMotion final {
            typename = std::enable_if_t<std::is_same_v<F, T>>>
   static RigidMotion Identity();
 
+  // A factory that construct a non-rotating motion using the given degrees of
+  // freedom.  Useful e.g. for save compatibility.
+  static RigidMotion MakeNonRotatingMotion(
+      DegreesOfFreedom<ToFrame> const& degrees_of_freedom);
```
eggrobin: `degrees_of_freedom_of_from_frame_origin`? or just `from_frame_origin`?

pleroy: The former.

PR #2390, opened 2019-12-03
1.8.1

eggrobin:
Note: in order to build this pull request, the following 1.8.1 DLLs need to be in `KSP Assemblies\1.8.1`:
```
Assembly-CSharp.dll
UnityEngine.CoreModule.dll
UnityEngine.dll
UnityEngine.ImageConversionModule.dll
UnityEngine.IMGUIModule.dll
UnityEngine.InputLegacyModule.dll
UnityEngine.PhysicsModule.dll
UnityEngine.TextRenderingModule.dll
UnityEngine.UI.dll
```


ksp_plugin_adapter/ksp_plugin_adapter.csproj
```diff
@@ -18,22 +17,38 @@
...
     <DebugType>full</DebugType>
     <Optimize>true</Optimize>
     <OutputPath>..\Release\GameData\Principia\</OutputPath>
+    <DefineConstants>TRACE;KSP_VERSION_1_8_1</DefineConstants>
+    <ErrorReport>prompt</ErrorReport>
+    <WarningLevel>4</WarningLevel>
+    <DebugSymbols>true</DebugSymbols>
+    <AllowUnsafeBlocks>false</AllowUnsafeBlocks>
+    <Prefer32Bit>false</Prefer32Bit>
+    <TargetFrameworkVersion>v4.7.2</TargetFrameworkVersion>
```
pleroy: Why not 4.8?  It is [recommended](https://dotnet.microsoft.com/download/dotnet-framework).

ksp_plugin_adapter/source_location.cs
```diff
@@ -4,10 +4,14 @@
 
 namespace System.Runtime.CompilerServices {
 
+#if KSP_VERSION_1_7_3
+
```
pleroy: What happens to those in 1.8.1?

eggrobin: See the comments above:
```C#
// Pretend that we have the caller file path and line number attributes, even
// though they do not exist in .NET 3.5; since this is actually substituted
// at compile time it will work as intended.
```

They are in the .NET Framework 4.5 and later, so they are there; we don't need to pretend that they are.

ksp_plugin_adapter/ksp_plugin_adapter.csproj
eggrobin: Mono doesn't mention 4.8 support as of this writing, so let's stay on the safe side https://www.mono-project.com/docs/about-mono/compatibility/.

If we find a 4.8 feature that we want we can try to change this and test whether it works.

PR #2397, opened 2019-12-08
Propagate the dynamical elements from the parts to the pile-up

pleroy:
Not used much at this point.  Tested (a bit) in game.


ksp_plugin/frames.hpp
```diff
@@ -84,6 +84,18 @@ using Camera = Frame<serialization::Frame::PluginTag,
 using RigidPart = Frame<serialization::Frame::PluginTag,
                         serialization::Frame::RIGID_PART, false>;
 
+// The |PileUp| is seen as a (currently non-rotating) rigid body; the degrees of
+// freedom of the parts in the frame of that body can be set, however their
+// motion is not integrated; this is simply applied as an offset from the rigid
+// body motion of the |PileUp|.
+// The origin of |RigidPileUp| is the centre of mass of the pile up.  Its axes
+// are those of Barycentric for now; eventually we will probably want to use the
+// inertia ellipsoid.
```
eggrobin: I don't think we will want to use the principal axes of the pile up here, since we still want things to work out of timewarp, where these axes could vary discontinuously (but as you say, this comment will need to change anyway).

ksp_plugin/pile_up.cpp
```diff
@@ -416,6 +404,55 @@ void PileUp::AppendToPart(DiscreteTrajectory<Barycentric>::Iterator it) const {
...
+
+    InertiaTensor<RigidPart> const& part_inertia_tensor =
+        part->inertia_tensor();
+    pile_up_inertia_tensor += part->inertia_tensor().Transform(part_to_pile_up);
+
+    Bivector<AngularMomentum, RigidPart> const part_angular_momentum =
+        Anticommutator(part_inertia_tensor,
+                       part_rigid_motion.angular_velocity_of_to_frame());
+    pile_up_angular_momentum +=
+        part_to_pile_up.linear_map()(part_angular_momentum);
```
eggrobin: This is incorrect; you are computing the angular momentum of the parts with respect to their centres of mass (in `RigidPart`) and then changing coordinates to `RigidPileUp`; this means that you are missing the contribution of the linear motion of the part to the angular momentum in `RigidPileUp`.

ksp_plugin/frames.hpp
pleroy: I think we'll want both, but we'll see.

ksp_plugin/pile_up.cpp
pleroy: Done (I think).

ksp_plugin/part_subsets.cpp
```diff
@@ -71,8 +67,7 @@ void Subset<Part>::Properties::Collect(
   collected_ = true;
   if (EqualsExistingPileUp()) {
     PileUp& pile_up = *parts_.front()->containing_pile_up();
-    pile_up.set_mass(total_mass_);
-    pile_up.set_intrinsic_force(total_intrinsic_force_);
+    pile_up.RecomputeFromParts();
```
eggrobin: Aren't you making things quadratic in the number of parts instead of linear here? Part counts can get pretty big, and this runs at every frame; we should be careful.

ksp_plugin/pile_up.cpp
```diff
@@ -424,27 +426,40 @@ DegreesOfFreedom<Barycentric> PileUp::RecomputeFromParts(
...
-    pile_up_inertia_tensor += part->inertia_tensor().Transform(part_to_pile_up);
+    // NOTE(phl): The following call reads like the tensor "transforms" the
+    // rigid motion.  Improve this API.
+    pile_up_inertia_tensor += part->inertia_tensor().Transform(
+        part_to_pile_up.rigid_transformation());
 
     Bivector<AngularMomentum, RigidPart> const part_angular_momentum =
         Anticommutator(part_inertia_tensor,
-                       part_rigid_motion.angular_velocity_of_to_frame());
+                       part_to_barycentric.angular_velocity_of_to_frame());
```
eggrobin: `part_to_barycentric.angular_velocity_of_to_frame` is the angular velocity *of* `Barycentric`; that feels wrong somehow.

ksp_plugin/pile_up.cpp
```diff
@@ -424,27 +426,40 @@ DegreesOfFreedom<Barycentric> PileUp::RecomputeFromParts(
   // Then compute the inertia tensor and the angular momentum of the pile-up.
   InertiaTensor<RigidPileUp> pile_up_inertia_tensor;
   Bivector<AngularMomentum, RigidPileUp> pile_up_angular_momentum;
```
eggrobin: Using this type to compute the angular momentum only makes sense if `RigidPileUp` remains non-rotating, as it is in non-rotating frames that the quantity ∫ r ∧ v dm is conserved (and is called the angular momentum).

ksp_plugin/pile_up.cpp
```diff
@@ -416,6 +406,68 @@ void PileUp::AppendToPart(DiscreteTrajectory<Barycentric>::Iterator it) const {
   }
 }
 
+DegreesOfFreedom<Barycentric> PileUp::RecomputeFromParts(
+    std::list<not_null<Part*>> const& parts) {
+  // First compute the overall mass and centre of mass of the pile-up.  Also
+  // compute the overall force applied to it.
+  // TODO(phl): We assume that forces are applied at the centre of mass of the
+  // pile-up, but they are really applied at the centre of mass of the parts, so
+  // this introduces a torque.
```
eggrobin: They are not even applied at the centre of mass of the parts; they are applied at some position, which we do not currently transmit to the C++.

pleroy: Fixed the comment.

ksp_plugin/pile_up.cpp
```diff
@@ -424,27 +426,40 @@ DegreesOfFreedom<Barycentric> PileUp::RecomputeFromParts(
...
-    RigidTransformation<RigidPart, RigidPileUp> const part_to_pile_up =
-        pile_up_to_barycentric.Inverse() * part_to_barycentric;
+    RigidMotion<RigidPart, RigidPileUp> const part_to_pile_up =
+        barycentric_to_pile_up * part_to_barycentric;
 
     InertiaTensor<RigidPart> const& part_inertia_tensor =
         part->inertia_tensor();
-    pile_up_inertia_tensor += part->inertia_tensor().Transform(part_to_pile_up);
+    // NOTE(phl): The following call reads like the tensor "transforms" the
+    // rigid motion.  Improve this API.
```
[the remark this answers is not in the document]
pleroy: Agreed, but not today.

ksp_plugin/pile_up.cpp
pleroy: But `RigidPileUp` *is* non-rotating.  I have fixed the comment in `frames.hpp`.

ksp_plugin/pile_up.cpp
pleroy: Change to `part_to_pile_up` which is more logical but not different.

eggrobin: But still, this is the angular velocity of the pile up axes (i.e., of non rotating axes) in the axes of the part, rather than the angular velocity of the part as seen in non-rotating axes; I think that is a sign error.

ksp_plugin/part_subsets.cpp
pleroy: As discussed, this happens once when the pile-up is constructed/updated, so O(N).

ksp_plugin/pile_up.cpp
pleroy: Fixed by switching everything to `RigidPileUp`.

ksp_plugin/pile_up.cpp
```diff
@@ -416,6 +406,68 @@ void PileUp::AppendToPart(DiscreteTrajectory<Barycentric>::Iterator it) const {
...
+            part_inertia_tensor,
+            part_to_pile_up.Inverse().angular_velocity_of_to_frame());
+    DegreesOfFreedom<RigidPileUp> const part_degrees_of_freedom =
+        part_to_pile_up({RigidPart::origin, Velocity<RigidPart>{}});
+    pile_up_angular_momentum +=
+        part_angular_momentum +
+        Wedge(part_degrees_of_freedom.position() - RigidPileUp::origin,
+              part->inertia_tensor().mass() *
+                  part_degrees_of_freedom.velocity()) *
+            Radian;
```
eggrobin: Please open an issue to track moving the computation of the momenta from the motion and the inertia to some place that can be unit-tested.

PR #2399, opened 2019-12-11
First step of simplifying and improving frames

pleroy:
* Make it possible to create a frame using a local `enum class`.
* Replace the boolean `is_inertial` by an enum.
* Add handedness (not used for now).

### DIFF
--- a/geometry/frame.hpp
+++ b/geometry/frame.hpp
@@ -16,16 +16,49 @@ namespace internal_frame {
 using base::not_constructible;
 using base::not_null;
 
-template<typename FrameTag, FrameTag frame_tag, bool frame_is_inertial>
-struct Frame : not_constructible {
-  using Tag = FrameTag;
-  static Position<Frame> const origin;
-  static Tag const tag = frame_tag;
-  static bool const is_inertial = frame_is_inertial;
+enum Inertia {
+  Inertial,
+  NonInertial,
+};
 
+enum class Handedness {
+  Left,
+  Right,
+};
+
+// The frame is serializable if and only if FrameTag is a protocol buffer enum.
+// To declare a local frame that does not need serialization, use the following
+// pattern:
+//   using MyFrame = Frame<enum class MyFrameTag>;
+//
+// or:
+//   enum class MyFrameTag;
+//   using MyFrame = Frame<MyFrameTag, MyFrameTag{}, Inertial>;
+//
+// By default, the frame is non-inertial and right-handed.
+template<typename FrameTag,
+         FrameTag tag_ = FrameTag{},
+         Inertia inertia_ = NonInertial,
+         Handedness handedness_ = Handedness::Right>
+struct Frame : not_constructible {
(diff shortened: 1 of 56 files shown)


geometry/frame.hpp
```diff
@@ -16,12 +16,41 @@ namespace internal_frame {
...
+};
+
+// The frame is serializable if and only if FrameTag is a protocol buffer enum.
+// To declare a local frame that does not need serialization, use the following
+// pattern:
+//   using MyFrame = Frame<enum class MyFrameTag>;
+//
+// or:
+//   enum class MyFrameTag;
+//   using MyFrame = Frame<MyFrameTag, MyFrameTag{}, Inertial>;
```
eggrobin: 1. `Frame<enum class MyFrameTag, MyFrameTag{}, Inertial>;` would work (do it that way in the other files);
2. We should put the `tag_` parameter last, so that it can be omitted even if inertia and handedness are provided: when we define a formal namespace-scope serializable frame, having to explicitly specify inertia and handedness is a feature, and for a local throwaway frame with a local enum class, the `tag_` is the most useless parameter (and thus should be the most defaulted one).

Item 2. should probably be done in a separate pull request, it will be a little bit noisy.

geometry/frame.hpp
```diff
@@ -16,12 +16,41 @@ namespace internal_frame {
 using base::not_constructible;
 using base::not_null;
 
-template<typename FrameTag, FrameTag frame_tag, bool frame_is_inertial>
+enum Inertia {
```
eggrobin: `ReferenceFrameInertia`; we don't see that name much because it's an `enum` rather than an `enum class`, so it doesn't need to  be catchy, and using up `Inertia` saddens me (see my comment about the possibility of renaming `InertiaTensor` in a recent pull request).

Even better, `ReferenceFrameMotion`: I was thinking of distinguishing rotating and non-rotating inertial frames at some point for `KeplerOrbit`.

geometry/frame_test.cpp
```diff
@@ -28,7 +44,7 @@ TEST_F(FrameDeathTest, SerializationError) {
     serialization::Frame message;
     World1::WriteToMessage(&message);
     World2::ReadFromMessage(message);
-  }, "tag ==");
+  }, "\\(tag\\) ==");
```
eggrobin: `R"(\(tag\) ==)"`

geometry/frame.hpp
pleroy: 1. Done.
2. Let's wait until the dust settles.

geometry/frame.hpp
pleroy: The class is called `Frame`, not `ReferenceFrame`, so the word "reference" is weird.  Going for `FrameMotion`.

geometry/frame_test.cpp
pleroy: Done, but I'm not sure if the extra parentheses are better than the extra `\`.

PR #2393, opened 2019-12-07
Camera rotation

eggrobin:
With this change, the horizontal in map view is the reference plane of the selected plotting frame.
Further, the camera rotates with the plotting frame (in particular, if the camera is looking at a body which is fixed in the plotting frame, the camera is fixed in the plotting frame). This mean that the plotted trajectories do not rotate as time passes, a behaviour which was somewhat confusing and annoying.

Examples below.

legends: ⟨*subject*⟩, ⟨*reference frame*⟩ (⟨*reference plane*⟩)

|  |  |
|---|---|
| screenshot25 Earth, ECEF (Earth equator) | screenshot24 Earth, ECSA (ecliptic) |
| screenshot26 Inner planets, ECI (Earth equator) | screenshot27 Inner planets, Sun-centred inertial (Sun equator) |
| screenshot31 Saturnian system, Saturn-centred sun-aligned (Saturn orbit) | screenshot32 Saturnian system, Saturn-centred inertial (Saturn equator) |
| screenshot30 Uranian system, Uranus-centred sun-aligned (Uranus orbit) | screenshot29 Uranian system, Uranus-centred inertial (Uranus equator) |

PR #2402, opened 2019-12-12
Use a custom timing to adjust the planetarium camera.

### LINKED ISSUE #2401: Calling ScaledSpace LateUpdate in LateUpdate BetterLateThanNever induces timing issues
We call the ScaledSpace LateUpdate after reorienting the camera in BetterLateThanNever since #2398.

The things in https://kerbalspaceprogram.com/api/script_order.html that run after ScaledSpace and have a LateUpdate that depends on the position of scaled space are affected.

Notably, this includes `Sun`, leading to the sun being separated from its lens flare:

screenshot0

It also includes `PatchedConicRenderer`. While the use case of patched conics in Principia is making fun of stock, I would like not to break things unless we have to (who knows what weird things might depend on patched conics being drawn in the right place).
screenshot1

The easiest way to solve this may be to call the relevant `LateUpdate`s in order after the `ScaledSpace` one.

eggrobin:
Fix #2401.

The `LateUpdate` of the new file is essentially the old `OrientPlanetariumCamera`. Some of the remainder of the diff is removing workarounds for timing-induced issues; it may be more to compare the adapter with its state prior to #2393: https://github.com/mockingbirdnest/Principia/compare/14ee57622a1064f4518f6f6e1c8c4e3f89cb3fcc...eggrobin:camera-rotation?expand=1#diff-c8666bccb01bd25d11f7ae5e0b5f2d1f.

### DIFF
--- a/ksp_plugin_adapter/ksp_plugin_adapter.cs
+++ b/ksp_plugin_adapter/ksp_plugin_adapter.cs
@@ -83,17 +83,7 @@ public partial class PrincipiaPluginAdapter
 
   private bool time_is_advancing_;
 
-  // True if a discontinuous change to the camera reference rotation occurred
-  // (due to a change in plotting frame), in which case we need special handling
-  // to keep the motion of the camera continuous.
-  bool should_transfer_camera_coordinates_ = false;
-  // The camera reference rotation applied during the previous frame; this is
-  // used when transfering camera coordinates to preserve continuity.
-  UnityEngine.QuaternionD previous_camera_reference_rotation_;
-  // The roll of the camera; this is set when transferring camera coordinates to
-  // preserve continuity of orientation, and gradually brought down to 0 so that
-  // the camera is horizontal in the new reference frame.
-  double camera_roll_ = 0;
+  private PlanetariumCameraAdjuster planetarium_camera_adjuster_;
 
   private RenderingActions map_renderer_;
   private RenderingActions galaxy_cube_rotator_;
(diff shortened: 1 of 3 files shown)


ksp_plugin_adapter/ksp_plugin_adapter.cs
```diff
@@ -590,6 +580,10 @@ private enum UnityLayers {
                                  JaiFailliAttendre);
     // Timing4, 19.
     TimingManager.FixedUpdateAdd(TimingManager.TimingStage.Late, Late);
+    // Custom timing, 301.
+    planetarium_camera_adjuster_ =
+        gameObject.AddComponent<PlanetariumCameraAdjuster>();
+    planetarium_camera_adjuster_.adapter = this;
```
pleroy: Could this be set at construction?  (Thinking of perfect forwarding, but maybe KSP/Unity/C# are not smart enough.)

eggrobin: I think it is discouraged to have constructors on `MonoBehaviours` (`AddComponent` constructs by reflection, so a parameterized constructor that takes the adapter is definitely out of the question).

eggrobin: We could possibly do some Unity trickery to get it from the parent in `Awake` or something, to avoid the public mutability, but at that point I am not sure it would improve readability.

PR #2404, opened 2019-12-14
Change the ReadFromMessage method of the integrators to be static

pleroy:
This will be necessary to be able to SFINAE out the ReadFromMessage for types that depend on non-serializable frames: no SFINAE on virtual methods.


integrators/integrators_body.hpp
```diff
@@ -14,80 +14,254 @@
...
-                                                 typename ODE::Position>()
+#define PRINCIPIA_CASE_SPRK(kind, method, action)      \
+  case serialization::FixedStepSizeIntegrator::kind: { \
+    action(method);                                    \
+  }
+
+#define PRINCIPIA_CASE_SRKN(kind, method, action)      \
+  case serialization::FixedStepSizeIntegrator::kind: { \
+    action(method);                                    \
+  }
```
eggrobin: *You are in a maze of twisty little macros, all alike.*

Perhaps these should be one macro.

integrators/integrators_body.hpp
```diff
@@ -14,80 +14,254 @@
...
+
+#define PRINCIPIA_CASES(slms_action, sprk_action, srkn_action)                 \
```
eggrobin: This needs some comments regarding the syntax (it has to be inside a switch statement) and the semantics (switches on some enumeration to all integrators, performing the appropriate action for the appropriate integrators).

The name of the macro could also be clearer (`PRINCIPIA_INTEGRATOR_CASES` maybe?).

integrators/integrators_body.hpp
```diff
@@ -14,80 +14,254 @@
...
+  PRINCIPIA_CASE_SPRK(BLANES_MOAN_2002_S6,                                     \
+                      BlanesMoan2002S6,                                        \
+                      sprk_action);                                            \
```
eggrobin: I think this semicolon is flapping in the breeze after a `}` once this is all expanded.
Either remove the unneeeded semicolon, or change the macro so that it is needed (`do {...} while (false)`, `[&]{...}()`, or some other horrible standard trick).

integrators/integrators_body.hpp
```diff
@@ -14,80 +14,254 @@
...
 #include "integrators/symplectic_runge_kutta_nyström_integrator.hpp"
 
-#define PRINCIPIA_CASE_SLMS(kind, method)                      \
-  case serialization::FixedStepSizeIntegrator::kind:           \
-    return SymmetricLinearMultistepIntegrator<methods::method, \
-                                              typename ODE::Position>()
+#define PRINCIPIA_CASE_SLMS(kind, method, action)      \
+  case serialization::FixedStepSizeIntegrator::kind: { \
+    action(method);                                    \
+  }
```
eggrobin: if `action` does not return or break, this is a fallthrough. Maybe `break`? Perhaps even without a semicolon, so as to require a semicolon from the caller?

pleroy: Added a `break;`.  I hope that it won't be detected as dead code by some smart compiler.

Let's not play games with semicolons, it seems odd to force the macro to include a semicolon.

PR #2405, opened 2019-12-14
Compile-time detection of non-serializable frames

pleroy:
The litmus test is `apsides_test.cpp`, which uses a non-serializable frame and construct complex objects with it, and survived to tell the story.


base/traits.hpp
```diff
@@ -0,0 +1,35 @@
...
+namespace base {
+namespace internal_traits {
+
+template<typename, typename = void, typename = void>
+struct has_templated_read_message : std::false_type {};
+
+template<typename T>
+struct has_templated_read_message<
+    T, std::void_t<decltype(&T::template ReadFromMessage<>)>>
+    : std::true_type {};
```
eggrobin: `has_templatized_read_from_message`; but really we need the template parameters to all be defaulted, i.e., used for SFINAE; so perhaps `has_enabled_read_from_message`?

base/traits.hpp
```diff
@@ -0,0 +1,35 @@
...
+
+template<typename, typename = void, typename = void>
+struct has_vanilla_read_message : std::false_type {};
```
eggrobin: `has_unconditional_read_from_message`.

base/traits.hpp
pleroy: Calling it `has_sfinae_read_from_message` because that's what it is.

base/traits.hpp
pleroy: Done.

PR #2411, opened 2019-12-23
Signature

eggrobin:
This pull request introduces `Signature`; it does introduce usages of it.
We should use it in `OrthogonalMap` and `EulerSolver` later on.

`Signature` is handedness-correct.

### DIFF
--- a/geometry/signature.hpp
+++ b/geometry/signature.hpp
@@ -7,8 +7,13 @@ namespace principia {
 namespace geometry {
 namespace internal_signature {
 
-struct deduce_sign_t {};
-constexpr deduce_sign_t deduce_sign_from_handedness{};
+using base::not_constructible;
+
+struct deduce_sign_direct_t final {};
+struct deduce_sign_indirect_t final {};
+
+constexpr deduce_sign_direct_t deduce_sign_for_direct_isometry{};
+constexpr deduce_sign_indirect_t deduce_sign_for_indirect_isometry{};
 
 template<typename FromFrame, typename ToFrame>
 class Signature {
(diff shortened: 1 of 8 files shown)


geometry/signature.hpp
```diff
@@ -0,0 +1,117 @@
...
+// matrix with ±1 on the diagonal.  There are 8 possible signatures: the
+// identity 𝟙, the central inversion -𝟙, the 180° rotations around all three
+// axes, and the reflections across the planes orthogonal to all three axes.
+template<typename FromFrame, typename ToFrame>
+class Signature : public LinearMap<FromFrame, ToFrame> {
+ public:
+  static constexpr Sign determinant =
+      FromFrame::handedness == ToFrame::handedness ? Sign::Positive()
+                                                   : Sign::Negative();
+  using DeduceSign = std::conditional_t<determinant.is_positive(),
```
pleroy: Preference for comparing the handednesses instead of using the determinant.

geometry/signature.hpp
```diff
@@ -0,0 +1,117 @@
...
+struct DeduceSignReversingOrientation final {};
+
+// A coordinate change whose matrix is a signature matrix, i.e., a diagonal
+// matrix with ±1 on the diagonal.  There are 8 possible signatures: the
+// identity 𝟙, the central inversion -𝟙, the 180° rotations around all three
+// axes, and the reflections across the planes orthogonal to all three axes.
+template<typename FromFrame, typename ToFrame>
+class Signature : public LinearMap<FromFrame, ToFrame> {
+ public:
+  static constexpr Sign determinant =
```
pleroy: Do we need this?  Could it be private?

PR #2416, opened 2019-12-28
Use Signature in the Euler-Arnold solver

eggrobin:


physics/euler_solver.hpp
```diff
@@ -105,7 +107,7 @@ class EulerSolver {
 
   // A rotation that describes which axes are flipped to adjust the signs of the
```
pleroy: "A rotation"?  Technically correct.

physics/euler_solver_body.hpp
```diff
@@ -17,10 +17,12 @@ namespace physics {
 namespace internal_euler_solver {
 
 using geometry::Commutator;
+using geometry::DeduceSignPreservingOrientation;
 using geometry::DefinesFrame;
 using geometry::Normalize;
 using geometry::Quaternion;
 using geometry::Vector;
+using geometry::Sign;
```
pleroy: Albaphetize.

PR #2421, opened 2019-12-30
Propagate the apparent rigid motion of parts from the C# code to the pile-up

pleroy:


ksp_plugin/interface.cpp
```diff
@@ -1095,17 +1102,24 @@ void __cdecl principia__SetMainBody(Plugin* const plugin, int const index) {
   return m.Return();
 }
 
-void __cdecl principia__SetPartApparentDegreesOfFreedom(
+void __cdecl principia__SetPartApparentRigidMotion(
     Plugin* const plugin,
-    PartId const part_id,
+    uint32_t const part_id,
```
eggrobin: `PartId` (or at least `std::uint32_t`.

PR #2424, opened 2020-01-01
Use Frame::unmoving in most places to denote null velocity

pleroy:
Also add `nonrotating` for null angular velocity.

### DIFF
--- a/physics/body_centred_body_direction_dynamic_frame_test.cpp
+++ b/physics/body_centred_body_direction_dynamic_frame_test.cpp
@@ -202,14 +202,11 @@ TEST_F(BodyCentredBodyDirectionDynamicFrameTest, CoriolisAcceleration) {
   Instant const t = t0_ + 0 * Second;
   // The velocity is opposed to the motion and away from the centre.
   DegreesOfFreedom<MockFrame> const point_dof =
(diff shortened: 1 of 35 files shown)


physics/body_centred_body_direction_dynamic_frame_test.cpp
```diff
@@ -208,9 +208,8 @@ TEST_F(BodyCentredBodyDirectionDynamicFrameTest, CoriolisAcceleration) {
                             20 * Metre / Second,
                             30 * Metre / Second})};
   DegreesOfFreedom<ICRS> const big_dof =
-      {Displacement<ICRS>({0 * Metre, 0 * Metre, 0 * Metre}) +
-           ICRS::origin,
-       Velocity<ICRS>()};
+      {Displacement<ICRS>({0 * Metre, 0 * Metre, 0 * Metre}) + ICRS::origin,
+       ICRS::unmoving};
```
eggrobin: Let's simplify that to `{ICRS::origin, ICRS::unmoving}`, I don't think adding 0 adds much to clarity here.

pleroy: Done everywhere in this file.

PR #2430, opened 2020-01-02
reference_radius defaults to radius in the configuration of an oblate body

### LINKED ISSUE #2429: reference_radius does not appear to take a default value C#-side
Reported by @damien212.
While https://github.com/mockingbirdnest/Principia/wiki/Principia-configuration-files says
> - `reference_radius`: an optional `quantity(length)`.
>  The reference radius *a*<sub>*e*</sub> used to make the spherical harmonics dimensionless.
>  Defaults to the stock [`radius`](https://github.com/Kopernicus/kittopia-dumps/blob/e09154a/Configs/Bop.cfg#L11).

the code only does https://github.com/mockingbirdnest/Principia/blob/389f1fe480948759917f23618c0c085cc1969a23/ksp_plugin_adapter/config_node_parsers.cs#L84 whereas it does https://github.com/mockingbirdnest/Principia/blob/389f1fe480948759917f23618c0c085cc1969a23/ksp_plugin_adapter/config_node_parsers.cs#L79-L80 when a stock-derived default is actually used.

A cursory look at blame seems to show that there never was such a default.

Note also that the documentation is somewhat out of date, in that it mentions `mean_radius` which is no longer configurable.

pleroy:
Also regenerate the config files since apparently we forgot to do that when we removed `mean_radius`.

Also edited to wiki to remove references to `mean_radius`.

Fix #2429.


ksp_plugin_adapter/config_node_parsers.cs
```diff
@@ -22,9 +24,11 @@ internal static class ConfigNodeParsers {
         axis_declination        = node.GetAtMostOneValue("axis_declination"),
         reference_angle         = node.GetAtMostOneValue("reference_angle"),
         angular_frequency       = node.GetAtMostOneValue("angular_frequency"),
-        reference_radius        = node.GetAtMostOneValue("reference_radius"),
-        j2                      = node.GetAtMostOneValue("j2"),
-        geopotential            = node.GetBodyGeopotentialElements().ToArray()
+        reference_radius =
+            node?.GetAtMostOneValue("reference_radius") ??
+            (j2 != null || geopotential != null ? body.Radius + " m" : null),
```
eggrobin: Keep it required for `NewCartesianBodyParameters`, i.e., revert the changes in this function.
See the comment about *nominal* vs. *sufficient* in the documentation: `NewCartesianBodyParameters` is the *sufficient* path.

PR #2445, opened 2020-01-13
Fix a number of Clang warnings

pleroy:


ksp_plugin/interface_renderer.cpp
```diff
@@ -17,11 +17,11 @@ using physics::DiscreteTrajectory;
 
 namespace {
 
-Renderer& GetRenderer(Plugin* const plugin) {
+[[maybe_unused]] Renderer& GetRenderer(Plugin* const plugin) {
   return CHECK_NOTNULL(plugin)->renderer();
 }
 
-Renderer const& GetRenderer(Plugin const* const plugin) {
+[[maybe_unused]] Renderer const& GetRenderer(Plugin const* const plugin) {
```
eggrobin: Either one of those is used, or both should be removed.

PR #2451, opened 2020-01-17
An inverse for `Anticommutator`

eggrobin:


geometry/symmetric_bilinear_form_body.hpp
```diff
@@ -111,6 +111,16 @@ SymmetricBilinearForm<Scalar, Frame, Multivector>::Anticommutator() const {
       matrix_.Trace() * R3x3Matrix<double>::Identity() - matrix_);
 }
 
+template<typename Scalar,
+         typename Frame,
+         template<typename, typename> typename Multivector>
```
pleroy: There are places in this file and others where you give names to the first two typenames thus: `template<typename S, typename F> typename Multivector`.  Please change everything to be consistent.

I would personally favour dropping the `S` and `F` everywhere, but it's your call.

geometry/symmetric_bilinear_form.hpp
```diff
@@ -57,6 +57,17 @@ class SymmetricBilinearForm {
            typename = std::enable_if_t<base::is_same_template_v<M, Vector>>>
   SymmetricBilinearForm<Scalar, Frame, Bivector> Anticommutator() const;
 
+  // Considering ⋀²V as a subspace of V⊗V, and using the inner product to see a
+  // symmetric bilinear form on bivectors as an endomorphism of bivectors,
+  // |*this| is an element of V⊗V⊗V*⊗V*, i.e., a twice contravariant, twice
+  // covariant tensor Sᵏˡₘₙ. |Contraction()| performs the contraction Sᵏʲⱼₙ, or
+  // equivalently Sʲˡₘⱼ.  It is not the partial trace, which would be Sᵏʲₘⱼ (or
+  // Sʲˡⱼₙ).  The following property holds up to roundoff:
+  //   F.Anticommutator().Contraction() / 2 == F.
```
pleroy: In dimension `d` this would give `F + Tr(F) * Identity * (d - 3) / 2`, so dimension 3 is a bit magic.  Maybe a comment to that effect?

eggrobin: No, but it didn't make sense in other ways anyway.

PR #2460, opened 2020-01-25
Remove InertiaTensor and its usages, and use MechanicalSystem instead

pleroy:


serialization/ksp_plugin.proto
```diff
@@ -49,13 +49,20 @@ message Manoeuvre {
 }
 
 message Part {
+  // Pre-Frenet.
+	message InertiaTensor {
```
eggrobin: `PreFrenetInertiaTensor`, so that it is clear that this is a compatibility message.

PR #2463, opened 2020-01-28
Propagates the torques and the force positions to the parts

pleroy:


serialization/journal.proto
```diff
@@ -1539,6 +1510,75 @@ message NewPlugin {
   optional Return return = 3;
 }
 
+message PartIncrementIntrinsicForce {
+	extend Method {
```
eggrobin: The spurious renaming was not great, but changing spaces to tabs is actively harmful.

PR #2471, opened 2020-02-15
Try to conserve angular momentum

eggrobin:


ksp_plugin/pile_up.cpp
```diff
@@ -308,30 +310,59 @@ void PileUp::DeformPileUpIfNeeded() {
   }
 
   // Compute the apparent centre of mass of the parts.
```
pleroy: Should talk about the apparent mechanical system.

PR #2474, opened 2020-02-19
Serialization of EulerSolver

pleroy:


physics/euler_solver.hpp
```diff
@@ -104,9 +111,11 @@ class EulerSolver {
 
   // Construction parameters.
   R3Element<MomentOfInertia> const moments_of_inertia_;
+  Bivector<AngularMomentum, InertialFrame> const initial_angular_momentum_;
```
eggrobin: This should never be used for any purpose other than serialization (see `  // Do not use initial_angular_momentum after this point.` in the body).
Rename it to serializable_initial_angular_momentum_ (or perhaps serialized), and change preferred_initial_angular_momentum_ back to initial_angular_momentum_.

PR #2483, opened 2020-02-26
Fix the handling of unloaded parts to prevent vessels exploding on "set orbit"

pleroy:


ksp_plugin/part.cpp
```diff
@@ -63,6 +72,10 @@ PartId Part::part_id() const {
   return part_id_;
 }
 
+bool Part::loaded() const {
+  return loaded_;
+}
+
 void Part::set_mass(Mass const& mass) {
```
eggrobin: `loaded_` should be set once we get actual data from `SetApparent...`.
Perhaps for symmetry `InsertOrKeepLoadedPart` should set it.

Issue #2400, opened 2019-12-12
History persistence lets .sfs files get very large


lpgagnon:
from discord discussion. Not a big problem so far, and too early to be proposing "fixes"; but worth tracking and gathering more data.

rp-1 career, ~15 active vessels including a ~5-year-old first orbit vessel; .sfs size was up to 280MB. Deleting just the first-orbit vessel reduced it by 60MB. Cleaning up all vessels from completed contracts brings it down to 24MB.

Main impact is on save/load speeds; saving was up to ~5s on SSD (presumably it's generating the data, not writing the file that's the bottleneck). Particularly annoying because this includes autosaves, so this can introduce long pauses the player doesn't control.

Secondary, minor impact is that big .sfs files are less easy to edit, share, etc.

eggrobin:
More data:
> [22:04]RurouniDonut:Deleting old vessals just made my save go from 252 mb to 6 mb

lpgagnon:
https://drive.google.com/open?id=15bLvwZiRsIiOOg602r4tfFR7fna_4v4P

162MB sfs, of which ~155MB is from 2 vessels ~1500d old.

pleroy:
I have managed to read the save and produce a `.wl` file for the vessel trajectories.  The code to do this is in https://github.com/pleroy/Principia/tree/2400 (some of that is throw-away, but some will ultimately be checked in).

The output file (776 MiB...) is on [Google Drive](https://drive.google.com/open?id=1JQh1Za2Mavh_71-Di3q2BCzwR0LWSjYE) and shared with @eggrobin.  Good luck trying to read it in *Mathematica*.

pleroy:
> Notable (in both linked saves): simple zipping reduces the size by a good third. So if nothing else, different compression algorithms could be worth investigating (though higher compression might come at the cost of slower saves & loads).

I noticed the same, but it mostly comes from the fact that the save uses base64, i.e., 6 bits per bytes.  It stands to reason that zipping would save 25%.  A better compression algorithm (zstd?) would probably not save more than ~5%.

damianreds:
hypervelocity from KSP forums here, providing a hefty persistent file for your analysis. I tried uploading it to gist but website kept crashing before finishing upload. Let me know if anything else is needed or if there's any other way I can help.

https://drive.google.com/file/d/1pxOinC2T4251Q8no2SX1sds2FfulqMsU/view?usp=sharing

pleroy:
[...]

PR #2485, opened 2020-03-08
Use ZFP for compressing discrete trajectories

pleroy:
This seems to buy us a factor 4 in size, while complying with the downsampling tolerance of 10 m.

Helps with #2400.


physics/discrete_trajectory_body.hpp
```diff
@@ -494,12 +503,64 @@ void DiscreteTrajectory<Frame>::WriteSubTreeToMessage(
...
+
+  // Times are exact.
+  ZfpCompressor time_compressor(0);
+  // Lengths are approximated to the downsampling tolerance if downsampling is
+  // enabled, otherwise they are exact.
+  Length const length_tolerance =
+      downsampling_.has_value() ? downsampling_->tolerance() : Length();
+  ZfpCompressor length_compressor(length_tolerance / Metre);
+  // Speeds are approximated based on the length tolerance and the average step
+  // in the timeline.
```
eggrobin: Let us not average things over the timeline; this is meaningless (the timeline may behave completely differently at different times) and it might result in weird edge cases that are hard to find and test, e.g., an interplanetary trajectory (large Δt, thus requiring a small speed tolerance to keep the error of the Hermite interpolation in check) getting mangled by a long time spent in low orbit, or even just by a lot of dense intervals (low Δt, 10 s for dense intervals, driving the mean Δt down).

Use the max Δt.

pleroy: Done, PTAL.

serialization/physics.proto
```diff
@@ -89,6 +89,12 @@ message DiscreteTrajectory {
   repeated int32 fork_position = 3;
   // Added in 陈景润.
   optional Downsampling downsampling = 4;
+
+  // ZFP support added in Frobenius.
+  optional int32 zfp_codec_version = 7;
+  optional int32 zfp_library_version = 8;
+  optional bytes zfp_timeline = 5;
+  optional int32 zfp_timeline_size = 6;
```
eggrobin: These should be wrapped in a message, since they are really optional and none or all should be present.

pleroy: Done, PTAL.

PR #2496, opened 2020-03-15
Add support for D-dimensional zfp compression

pleroy:
Still sticking to 2D as it is the sweet spot.  Some changes to the ZFP header to avoid size limitations (uncovered in 4D).


base/zfp_compressor_body.hpp
```diff
@@ -2,10 +2,82 @@
...
+                                         /*nw=*/v.size() / padding_),
+                            std::move(free));
+  } else {
+    static_assert(false, "Unsupported dimension");
+  }
+}
+
+template<int D>
+constexpr std::int64_t NDimensionalHelper<D>::RoundUp(std::int64_t const size) {
+  // Hacker's Delight, H. S. Warren, Jr., section 3-1.
```
eggrobin: Add this to the bibliography, refer to the generated abbreviation.

physics/ephemeris_test.cpp
```diff
@@ -1111,16 +1111,16 @@ TEST(EphemerisTestNoFixture, DiscreteTrajectoryCompression) {
...
   std::string compressed;
   auto compressor = google::compression::NewGipfeliCompressor();
   compressor->Compress(uncompressed, &compressed);
 
   // We want a change detector, but the actual compressed size varies depending
   // on the exact numerical values, and therefore on the mathematical library.
-  EXPECT_LE(16'974, compressed.size());
-  EXPECT_GE(16'974, compressed.size());
+  EXPECT_LE(16'935, compressed.size());
+  EXPECT_GE(16'935, compressed.size());
```
eggrobin: I suppose things are getting smaller because we lose the metadata header?

base/zfp_compressor_body.hpp
pleroy: Done.

physics/ephemeris_test.cpp
pleroy: Yes.  48 bits per series, I think, and then the vagaries of gipfeli.

PR #2497, opened 2020-03-16
Clang and MSVC have different opinions

pleroy:


eggrobin:
Does not really look good, but fixed by #2498.

PR #2500, opened 2020-03-17
Fuel angular momentum is lost as-is only from SRMs

eggrobin:
Other engines should also lose some angular momentum, but since liquid fuels go through pipes and injectors they should lose some of their angular momentum before they leave the rocket. This will be addressed in a future pull request.


base/zfp_compressor.cpp
```diff
@@ -15,7 +15,7 @@ namespace zfp_compressor_internal {
 // small for our purposes.  Therefore, we must not include the bit
 // ZFP_HEADER_META.  We still need a header, though, to record the compression
 // parameters.
-static constexpr uint header_mask = ZFP_HEADER_MODE | ZFP_HEADER_MAGIC;
+static constexpr uint header_mask = ZFP_HEADER_FULL;
```
pleroy: Please revert this change.

ksp_plugin/part.hpp
```diff
@@ -74,6 +74,10 @@ class Part final {
   Mass const& mass() const;
   void set_inertia_tensor(InertiaTensor<RigidPart> const& inertia_tensor);
   InertiaTensor<RigidPart> const& inertia_tensor() const;
+  // Whether this part is a solid rocket motor, whose lost mass is expelled with
```
pleroy: ... the lost mass of which ...

eggrobin: As discussed, this is fine.

PR #2516, opened 2020-03-31
Skip the points of the parts's psychohistory that are before the vessel's psychohistory fork

### LINKED ISSUE #2490: Appending before fork time
Log file created at: 2020/03/12 09:09:17
Running on machine: DESKTOP-68GQI37
Log line format: [IWEF]mmdd hh:mm:ss.uuuuuu threadid file:line] msg
    @   00007FFE864D6360  	google::LogMessageFatal::~LogMessageFatal [0x00007FFE864D635F+47]
    @   00007FFDF96788CC  	(No symbol) [0x00007FFDF96788CB]
    @   00007FFDF984E857  	principia__VesselVelocity [0x00007FFDF984E856+1276966]
    @   00007FFDF984DF85  	principia__VesselVelocity [0x00007FFDF984DF84+1274708]
    @   00007FFDF97F00B5  	principia__VesselVelocity [0x00007FFDF97F00B4+889988]
    @   00007FFDF96E7998  	principia__CatchUpLaggingVessels [0x00007FFDF96E7997+231]
    @   00000203C9FFAA17  	(No symbol) [0x00000203C9FFAA16]
F0312 09:09:17.425650 28596 discrete_trajectory_body.hpp:170] Check failed: this->is_root() || time > this->Fork()->time Append at +6.25980497443201148e+03 s which is before fork time +6.26290882815326040e+03 s

I get these errors when seperating stages. I'm running quite a few mods which I understand could be an issue. Running Frenet on 1.8.1

### LINKED ISSUE #2507: Undocking causes gamecrash
I have docked two similar ships and undocked them after some time (KSP was closed and started again between docking and undocking). Undocking causes gamecrash. I've removed principia, undocked the ships. Got principia back and the game crashes on loading of the gamesave. Other saves work fine. The "broken" save works fine without principia.

KSP.log: 
[KSP.log](https://github.com/mockingbirdnest/Principia/files/4390244/KSP.log)

INFO from glog/principia (undock button -> crash):
[INFO.20200326-204405.22416.log](https://github.com/mockingbirdnest/Principia/files/4390255/INFO.20200326-204405.22416.log)

INFO from glog/principia (loading a "broken" save):
[INFO.20200326-215121.18636.log](https://github.com/mockingbirdnest/Principia/files/4390256/INFO.20200326-215121.18636.log)

The save (those ships are R-III I and R-III II):
https://drive.google.com/file/d/1AXfjbyWinKW5u1tGSdQVJzsLcKAbPvCS/view?usp=sharing

pleroy:
Also write the Principia version to the journal.

Fix #2490.
Fix #2507.

### DIFF
--- a/journal/recorder.cpp
+++ b/journal/recorder.cpp
@@ -5,8 +5,10 @@
 
 #include "base/array.hpp"
 #include "base/hexadecimal.hpp"
-#include "glog/logging.h"
 #include "base/serialization.hpp"
+#include "base/version.hpp"
+#include "glog/logging.h"
+#include "journal/profiles.hpp"
 
 namespace principia {
 
@@ -34,6 +36,19 @@ void Recorder::WriteAtDestruction(serialization::Method const& method) {
 void Recorder::Activate(base::not_null<Recorder*> const recorder) {
   CHECK(active_recorder_ == nullptr);
   active_recorder_ = recorder;
+
+  // When the recorder gets activated, pretend that we got a GetVersion call.
+  // This will record the version at the beginning of the journal, which is
+  // useful for forensics.
+  serialization::Method method;
+  not_null<serialization::GetVersion*> const get_version =
+      method.MutableExtension(serialization::GetVersion::extension);
+  active_recorder_->WriteAtConstruction(method);
+  not_null<serialization::GetVersion::Out*> const out =
+      get_version->mutable_out();
+  out->set_build_date(base::BuildDate);
+  out->set_version(base::Version);
+  active_recorder_->WriteAtDestruction(method);
 }
 
 void Recorder::Deactivate() {
--- a/ksp_plugin/vessel.cpp
+++ b/ksp_plugin/vessel.cpp
@@ -237,11 +237,27 @@ void Vessel::AdvanceTime() {
   auto prediction = prediction_->DetachFork();
 
   history_->DeleteFork(psychohistory_);
-  AppendToVesselTrajectory(&Part::history_begin,
+  AppendToVesselTrajectory(/*fork_time=*/std::nullopt,
+                           &Part::history_begin,
                            &Part::history_end,
                            *history_);
   psychohistory_ = history_->NewForkAtLast();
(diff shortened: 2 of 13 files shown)


ksp_plugin/vessel.cpp
```diff
@@ -575,6 +591,7 @@ void Vessel::SwapPrognostication(
 }
 
 void Vessel::AppendToVesselTrajectory(
+    std::optional<Instant> const& fork_time,
```
eggrobin, line 594: We pass a `DiscreteTrajectory`, so we can compute this ourselves instead of making the caller do it.

Remove the parameter, and let
```C++
bool const after_fork_time =
    !at_end_of_part_trajectory && (trajectory.is_root() || first_time > trajectory().Fork()->time);
```
(or maybe hoist it as an optional out of the loop, but given that the ancestry is always shallow here this is probably not useful).

journal/recorder.cpp
```diff
@@ -34,6 +36,19 @@ void Recorder::WriteAtDestruction(serialization::Method const& method) {
 void Recorder::Activate(base::not_null<Recorder*> const recorder) {
   CHECK(active_recorder_ == nullptr);
   active_recorder_ = recorder;
+
+  // When the recorder gets activated, pretend that we got a GetVersion call.
```
eggrobin, line 40: Why pretend? Couldn’t we just call `principia__GetVersion` here instead of reimplementing it?

We might have to move it to its own interface_something file for cleaner includes, but splitting the main `interface` file tends to be something we want to do anyway…

ksp_plugin/vessel.cpp
pleroy: Done.  `is_root` is cheap, `Fork` is not.

journal/recorder.cpp
pleroy: I started along that path and it's not nice.  The API of the interface is not exactly designed for being called from C++.  In this case the parameters are `char const** const`, which would require awkward storage management.  It's much simpler to just set two fields of a proto.

PR #2518, opened 2020-04-02
Extend the flight plan when the last manœuvre is moved forward or extended

### LINKED ISSUE #852: Prolong flight plan when pushing the last burn forward

pleroy:
Also changed the default flight plan length to 1 hour from 1000 seconds (?).

Fix #852.

### DIFF
--- a/ksp_plugin/flight_plan.cpp
+++ b/ksp_plugin/flight_plan.cpp
@@ -173,8 +173,17 @@ Status FlightPlan::Replace(NavigationManœuvre::Burn const& burn,
   if (manœuvre.IsSingular()) {
     return Singular();
   }
-  if (!manœuvre.FitsBetween(start_of_previous_coast(index),
-                            start_of_next_burn(index))) {
+  if (index == number_of_manœuvres() - 1) {
+    // This is the last manœuvre.  If it doesn't fit just because the flight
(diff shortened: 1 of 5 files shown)


ksp_plugin/flight_plan.cpp
```diff
@@ -173,8 +173,17 @@ Status FlightPlan::Replace(NavigationManœuvre::Burn const& burn,
...
     return Singular();
   }
-  if (!manœuvre.FitsBetween(start_of_previous_coast(index),
-                            start_of_next_burn(index))) {
+  if (index == number_of_manœuvres() - 1) {
+    // This is the last manœuvre.  If it doesn't fit just because the flight
+    // plan is too short, extend the flight plan.
+    if (manœuvre.IsAfter(start_of_previous_coast(index))) {
+      desired_final_time_ =
+          std::max(desired_final_time_, manœuvre.final_time());
```
eggrobin, line 181: What happens if we are unable to reach `desired_final_time_`?

pleroy: We will bubble up a status to the C# and it will say something like "reached the maximum number of steps, try to increase it".  Which is exactly the behaviour if you had extended the trajectory by hand.

Issue #2339, opened 2019-09-27
Add support for optional fields in the auxiliary interface messages


eggrobin:
See https://github.com/mockingbirdnest/Principia/pull/2337#discussion_r328826119, https://github.com/mockingbirdnest/Principia/pull/2338#discussion_r328291586.

This probably involves generating the marshalers.

PR #2521, opened 2020-04-06
Add an intermediate class, MonoMarshaler, to hide Mono bugs

pleroy:
Rebase all the marshalers on this class, and unify `in` and `out` if needed.

First step towards addressing #2339.


ksp_plugin_adapter/mono_marshaler.cs
```diff
@@ -0,0 +1,63 @@
...
+
+  void ICustomMarshaler.CleanUpNativeData(IntPtr native_data) {
+    IntPtr actual_native_data = IntPtr.Zero;
+    lock (allocated_intptrs_) {
+      if (allocated_intptrs_.Contains(native_data)) {
+        actual_native_data = native_data;
+        allocated_intptrs_.Remove(native_data);
+      }
+    }
+    if (actual_native_data != IntPtr.Zero) {
```
eggrobin, line 36: Document, either here or on `allocated_intptrs_`, what we are doing (the comment at the top of the class documents what the issue is, but the solution is nontrivial and deserves a comment).

ksp_plugin_adapter/utf8_marshaler.cs
```diff
@@ -6,38 +6,28 @@
...
-// the caller.
-internal class InUTF8Marshaler : UTF8Marshaler {
-  // In addition to implementing the |ICustomMarshaler| interface, custom
-  // marshalers must implement a static method called |GetInstance| that accepts
-  // a |String| as a parameter and has a return type of |ICustomMarshaler|,
-  // see https://goo.gl/wwmBTa.
+// A marshaler for UTF-8 strings whose ownership is not taken from C++.
+internal class UnownedUTF8Marshaler : UTF8Marshaler {
   public static ICustomMarshaler GetInstance(string s) {
+    UnityEngine.Debug.LogError("UnownedUTF8Marshaler.GetInstance");
```
eggrobin, line 19: Remove the traces.

ksp_plugin_adapter/utf8_marshaler.cs
```diff
@@ -6,38 +6,28 @@
...
 }
 
-// A marshaler for in parameter UTF-8 strings whose ownership is not taken from
-// the caller.
-internal class InUTF8Marshaler : UTF8Marshaler {
-  // In addition to implementing the |ICustomMarshaler| interface, custom
-  // marshalers must implement a static method called |GetInstance| that accepts
-  // a |String| as a parameter and has a return type of |ICustomMarshaler|,
-  // see https://goo.gl/wwmBTa.
+// A marshaler for UTF-8 strings whose ownership is not taken from C++.
```
eggrobin, line 16: "Not taken from C++" is a property that only makes sense for results/out parameters: for in and in out parameters there is no C++ ownership to take.
Further, the marshaler name itself is misleading: for in and in out parameters, the string ownership is C#-side—and since the marshaler is doing the construction and destruction, it owns the string.

This really is a `NoTransferOfOwnershipUTF8Marshaler`: either the string may be owned by C++ or by C#, but it stays that way. That seems like too many words for an identifier; maybe `UncededUTF8Marshaler`?

ksp_plugin_adapter/utf8_marshaler.cs
```diff
@@ -46,62 +36,51 @@ internal class InUTF8Marshaler : UTF8Marshaler {
...
+  private static readonly UnownedUTF8Marshaler instance_ =
+      new UnownedUTF8Marshaler();
 }
 
-// A marshaler for out parameter or return value UTF-8 strings whose ownership
-// is taken by the caller.
-internal class OutOwnedUTF8Marshaler : OutUTF8Marshaler {
+// A marshaler for UTF-8 strings whose ownership is taken from C++.  Useful for
+// out parameters and returned values.
+internal class OwnedUTF8Marshaler : UnownedUTF8Marshaler {
```
eggrobin, line 67: By symmetry, `CededUTF8Marshaler`.

Note that deriving from the other one seems a bit weird now.

If we wanted to use it for in parameters, it would have to allocate via an interface function. We have no use for that at the moment, so it should probably `Log.Fatal` if we accidentally use it that way.

ksp_plugin_adapter/mono_marshaler.cs
pleroy: Done.

ksp_plugin_adapter/utf8_marshaler.cs
pleroy: Oops, done.

ksp_plugin_adapter/utf8_marshaler.cs
pleroy: The verb "cede" in this context reads weird.  Let's discuss the name.

ksp_plugin_adapter/utf8_marshaler.cs
pleroy: Granted, inheritance is a bit odd, but then composition doesn't work so well for classes that cannot have nonstatic members.

I don't understand the part about allocating via an interface.

pleroy: Done.

ksp_plugin_adapter/utf8_marshaler.cs
pleroy: Done.

ksp_plugin_adapter/mono_marshaler.cs
```diff
@@ -0,0 +1,67 @@
...
+    }
+    return result;
+  }
+
+  public abstract object MarshalNativeToManaged(IntPtr native_data);
+
+  // Since Mono calls CleanUpNativeData with IntPtrs that have not been
+  // allocated by MarshalManagedToNative, we track here all the values returned
+  // by MarshalManagedToNative and only process in CleanUpNativeData those are
+  // found in this set.
```
eggrobin, line 61: Missing relative pronoun.

ksp_plugin_adapter/utf8_marshaler.cs
```diff
@@ -40,47 +26,52 @@ internal class UnownedUTF8Marshaler : UTF8Marshaler {
...
-                               GetHashCode() + " " + buffer + " " +
-                               managed_object);
     return buffer;
   }
 
-  public override object MarshalNativeToManaged(IntPtr native_data) {
+  public  object MarshalNativeToManagedImplementation(IntPtr native_data) {
     int size;
-    for (size = 0; Marshal.ReadByte(native_data, size) != 0; ++size) {}
+    for (size = 0; Marshal.ReadByte(native_data, size) != 0; ++size) { }
```
eggrobin, line 13: Why the space?

ksp_plugin_adapter/utf8_marshaler.cs
```diff
@@ -6,102 +6,81 @@
...
+// out parameters and returned values.
+internal class OwnershipTransferUTF8Marshaler : UTF8Marshaler {
   public static ICustomMarshaler GetInstance(string s) {
     return instance_;
   }
 
-  public override void CleanUpNativeData(IntPtr native_data) {
-    Marshal.FreeHGlobal(native_data);
+  public override void CleanUpNativeDataImplementation(IntPtr native_data) {
+    throw Log.Fatal("use |MarshalAs(UnmanagedType.LPWStr)| for in parameters");
```
eggrobin, line 33: No, `LPWStr` is for UTF-16 strings.

pleroy:
retest this please

PR #2522, opened 2020-04-09
Generate marshalers for structs

pleroy:
Makes it possible to pass UTF-8 strings in structs.  Also fixed a typo in ~~marshaller~~ marshaler.  Tested with the following blueprint file in RSS:
```
principia_numerics_blueprint:NEEDS[RealSolarSystem] {
  ephemeris {
    ...
    fitting_tolerance = 1000 μm
    ...
  }
  history {
    ...
  }
  psychohistory {
    ...
    length_integration_tolerance = 10000000 Å
    ...
  }
}
```
As usual there are special commits showing the generated files before/after.

Another step towards solving #2433


quantities/parser_body.hpp
```diff
@@ -105,6 +105,8 @@ inline Unit ParseUnit(std::string const& s) {
   if (s == "") {
     return Unit(1.0);
   // Units of length.
+  } else if (s == u8"Å") {
+    return Unit(100 * si::Pico(si::Metre));
```
eggrobin, line 109: Use `Ångström`:
https://github.com/mockingbirdnest/Principia/blob/a28e05c5372005daee013e90b5a3166158fe592e/quantities/bipm.hpp#L20

PR #2523, opened 2020-04-10
Update install_deps.sh to use principia_make.sh from dependencies

ts826848:
zfp does not support in-tree CMake builds, so the additional `build` directory is needed in install_deps.sh.

~~I'm not entirely sure changing `#include "zfp/zfp.h"` to `#include "zfp.h"` won't break anything on Windows. I have to reinstall VS 2017 on my machine sometime tomorrow.~~

I still get build errors with these two commits, but they seem to be unrelated:

```
Undefined symbols for architecture x86_64:
  "google::FlagRegisterer::FlagRegisterer<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > >(char const*, char const*, char const*, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >*, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >*)", referenced from:
      __GLOBAL__sub_I_logging.cc in libglog.a(libglog_la-logging.o)
      __GLOBAL__sub_I_vlog_is_on.cc in libglog.a(libglog_la-vlog_is_on.o)
  "google::FlagRegisterer::FlagRegisterer<bool>(char const*, char const*, char const*, bool*, bool*)", referenced from:
      __GLOBAL__sub_I_logging.cc in libglog.a(libglog_la-logging.o)
      __GLOBAL__sub_I_utilities.cc in libglog.a(libglog_la-utilities.o)
  "google::FlagRegisterer::FlagRegisterer<int>(char const*, char const*, char const*, int*, int*)", referenced from:
      __GLOBAL__sub_I_logging.cc in libglog.a(libglog_la-logging.o)
      __GLOBAL__sub_I_vlog_is_on.cc in libglog.a(libglog_la-vlog_is_on.o)
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make: *** [bin/tools] Error 1
```

This is fixable by adding `-lgflags` to the Makefile since I already had gflags installed through Homebrew, but I'm not sure how you guys want this to be addressed. Should I specify gflags as a build dependency in install_deps.sh along with clang/libc++-dev and friends, or do you want it built as a third-party dep?

I still need to try this on Windows too to see if it's an issue on there.


base/zfp_compressor.hpp
```diff
@@ -5,7 +5,7 @@
 #include <vector>
 
 #include "base/not_null.hpp"
-#include "zfp/zfp.h"
+#include "zfp.h"
```
eggrobin, line 8: We have automation that builds for Macintosh, Linux, and Windows, although it is not based on `install_deps.sh`; the `#include` is fine, if you need to change it it means the `install_deps` does not do what it should.

install_deps.sh
```diff
@@ -107,3 +107,17 @@ git pull
 cmake -DCMAKE_C_COMPILER:FILEPATH=`which clang` -DCMAKE_CXX_COMPILER:FILEPATH=`which clang++` -DCMAKE_C_FLAGS="${C_FLAGS}" -DCMAKE_CXX_FLAGS="${CXX_FLAGS}" -DCMAKE_LD_FLAGS="${LD_FLAGS}" -DBENCHMARK_ENABLE_GTEST_TESTS=OFF
 make -j8
 popd
+
+if [ ! -d "zfp" ]; then
+  git clone "https://github.com/LLNL/zfp.git"
```
eggrobin, line 112: As for the other dependencies, we don’t live at head, we have our own fork: https://github.com/mockingbirdnest/zfp.

install_deps.sh
```diff
@@ -107,3 +107,17 @@ git pull
...
+pushd zfp
+git checkout master
+git pull
+rm -rf build
+mkdir build
+pushd build
+cmake -DCMAKE_C_COMPILER:FILEPATH=`which clang` -DCMAKE_CXX_COMPILER:FILEPATH=`which clang++` -DCMAKE_C_FLAGS="${C_FLAGS}" -DCMAKE_CXX_FLAGS="${CXX_FLAGS}" -DCMAKE_LD_FLAGS="${LD_FLAGS}" -DBUILD_SHARED_LIBS=OFF ..
+make -j8
+popd
+popd
```
eggrobin, line 123: So, looking at the pipelines, they do
```sh
mkdir build; cd build; cmake -DZFP_WITH_OPENMP=OFF -DBUILD_SHARED_LIBS=OFF ..; make
```
on both Linux and Mac, which seems bad (in particular it does not pass `DCMAKE_C_COMPILER` etc.).

We should really unify `install_deps.sh` with the code run by the pipelines in some way; maybe via https://docs.microsoft.com/en-us/azure/devops/pipelines/repos/multi-repo-checkout?view=azure-devops?
@pleroy, any thoughts?

install_deps.sh
ts826848: Fixed by ba56718

base/zfp_compressor.hpp
ts826848: Reverted in 53d3588 and fixed by using the right zfp repo.

pleroy:

ts826848:
The changes have been made. I've successfully built Principia on my mac, and all the tests that passed before still pass. Hopefully things will work fine on Linux too.

PR #2524, opened 2020-04-10
Marshaling of structs that include repeated fields

pleroy:
* Add a new class, `RepeatedMarshaler`, which is a helper referenced by the generated code that marshals a `T[]` into an array of `IntPtr`, seen on the C++ side as a `T**`.
* Remove support for the `(size)` field from the generator.  We'll use null termination everywhere.  The `(size)` field was the source of a lot of complexity in the generator, and it becomes unmanageable with custom marshalers.  While that removal simplifies the generator quite a bit, it changes many parts of it.
* Remove `InBodyParametersMarshaler`, it's now replaced by generated code.
* This was tested in game with RSS by checking the resulting geopotential and using `°` for angles.

As usual, there are separate commit for the generated code before/after the changes.


ksp_plugin_adapter/repeated_marshaler.cs
```diff
@@ -0,0 +1,56 @@
...
+    if (!(managed_object is T[] value)) {
+      throw new NotSupportedException();
+    }
+    int sizeof_intptr = Marshal.SizeOf(typeof(IntPtr));
+    IntPtr native = Marshal.AllocHGlobal(sizeof_intptr * (value.Length + 1));
+    for (int i = 0; i < value.Length; ++i) {
+      Marshal.WriteIntPtr(native,
+                          i * sizeof_intptr,
+                          t_marshaler_instance_.
+                              MarshalManagedToNative(value[i]));
```
eggrobin, line 36: Check that the resulting IntPtr is not null.

serialization/journal.proto
```diff
@@ -455,18 +450,18 @@ message DeserializePlugin {
     optional DeserializePlugin extension = 5050;
   }
   message In {
-    required string serialization = 1 [(size) = "serialization_size"];
+    required string serialization = 1;
     required fixed64 deserializer = 2
         [(pointer_to) = "PushDeserializer",
-         (is_consumed_if) = "serialization->empty()"];
+         (is_consumed_if) = "*serialization == '\\0'"];
```
eggrobin, line 456: `std::string_view(serialization).empty()`?

eggrobin: (We might want to change the API so that termination is signalled by `serialization == nullptr` but that is a job for another pull request.)

pleroy: Done, but note how this is less efficient (not that we care, I think it's only used when playing a journal).

PR #2525, opened 2020-04-10
principia_make

eggrobin:


principia_make.sh
```diff
@@ -0,0 +1,19 @@
+# It seems that protoc really wants its dependencies to be in /usr/local/lib.
+# In some setups, e.g., Azure pipelines, this does not work, so we need to help
+# it find its dynamic libraries.
+if [ "${AGENT_OS?}" == "Darwin" ]; then
+  install_name_tool \
+      -change /usr/local/lib/libprotobuf.17.dylib ./deps/protobuf/src/.libs/libprotobuf.17.dylib \
```
pleroy, line 6: Let's do one parameter per line for all these calls.

principia_make.sh
```diff
@@ -0,0 +1,19 @@
...
+      deps/protobuf/src/.libs/libprotoc.17.dylib
+  install_name_tool \
+      -change /usr/local/lib/libprotobuf.17.dylib ./deps/protobuf/src/.libs/libprotobuf.17.dylib \
+      deps/protobuf/src/.libs/protoc
+  install_name_tool \
+      -change /usr/local/lib/libprotoc.17.dylib ./deps/protobuf/src/.libs/libprotoc.17.dylib \
+      deps/protobuf/src/.libs/protoc
+elif [ "${AGENT_OS?}" == "Linux" ]; then
+  export LD_LIBRARY_PATH="./deps/protobuf/src/.libs:$LD_LIBRARY_PATH"
+fi
```
pleroy, line 16: Should this fail if `AGENT_OS` is something unexpected?

eggrobin: If we wanted to check that, the last target in the dependency chain would be the wrong place.

PR #2526, opened 2020-04-11
Support for marshaling optional fields in interchange messages

### LINKED ISSUE #2339: Add support for optional fields in the auxiliary interface messages

pleroy:
* On the C# side, these optionals are represented as `T?`.  On the C++ side, as `T*`.  Note that in C++ the pointers cannot default to `nullptr` (because POD), so the client must null these fields as needed (`{}` is your friend).
* If an interchange message has optional fields, it needs a generated marshaler and must therefore be a class.
* The basic class for marshaling optionals is `OptionalMarshaler`.  If an optional field `is_produced`, the `OptionalMarshaler` is wrapped into an `OwnershipTransferMarshaler`.
* Ownership transfer of interchange messages is achieved by giving them an `operator new` and deleting them with a new interface function, `DeleteInterchange`.
* When replaying a journal, optional fields need "storage" objects, just like repeated fields.  Previously this was hacked with "unique_ptr<T>" but that doesn't work in all cases.  So we now have proper storage, filled with lambdas.
* A message that is a C# class is now returned by `T*` on the C++ side, and ownership may be transferred to C#.

Fix #2339.


ksp_plugin_adapter/ownership_transfer_marshaler.cs
```diff
@@ -0,0 +1,44 @@
...
+
+  public override void CleanUpNativeDataImplementation(IntPtr native_data) {
+    t_marshaler_instance_.CleanUpNativeDataImplementation(native_data);
+  }
+
+  public override IntPtr MarshalManagedToNativeImplementation(
+      object managed_object) {
+    return t_marshaler_instance_.MarshalManagedToNativeImplementation(
+        managed_object);
+  }
```
eggrobin, line 25: This should have the same checks as `OwnershipTransferUTF8Marshaler` (though the error message should more accurately say that we do not support transfer of ownership in instead of suggesting something that might not work).

https://github.com/mockingbirdnest/Principia/blob/2d809b32aa1a9c3849a9b131185bd91779d768fc/ksp_plugin_adapter/utf8_marshaler.cs#L32-L39

ksp_plugin/interface_vessel.cpp
```diff
@@ -93,33 +93,30 @@ OrbitAnalysis __cdecl principia__VesselRefreshAnalysis(
...
-      analysis.elements.mean_eccentricity =
-          ToInterval(elements.mean_eccentricity_interval());
-      analysis.elements.mean_inclination =
-          ToInterval(elements.mean_inclination_interval());
-      analysis.elements.mean_longitude_of_ascending_nodes =
-          ToInterval(elements.mean_longitude_of_ascending_node_interval());
-      analysis.elements.mean_semimajor_axis =
-          ToInterval(elements.mean_semimajor_axis_interval());
+      analysis->elements = new OrbitalElements{
+          /*sidereal_period=*/elements.sidereal_period() / Second,
```
eggrobin, line 106: Can we use designated initializers here?

Same question elsewhere.

pleroy:
retest this please

Issue #2519, opened 2020-04-04
Spin-up on reentry under physics warp with FAR


eggrobin:
@nepphhh reported this issue, illustrated by the following video.
https://www.youtube.com/watch?v=dNHiZsGIurg

Some notes from experiments by @nepphhh:
- the issue is not visible prior to Frobenius;
- the issue is not visible without FAR;
- the issue is not visible unless physics-warp is used.

@Myshiko reports that this only occur on reentries with high deceleration, not, e.g., when flying an aeroplane.

Note that @ferram4 and @dkavolis confirm that FAR (and FAR continued) correctly applies forces and torques (via `Part.Add(Torque|Force[AtPosition])`, so the discrepancy in angular momentum is not due to FAR applying a torque via some nonconformant path for which we would not account.

@ferram4 suggests that this may be some minor error in, e.g., application points, getting amplified by FAR’s reentry forces being much greater than stock’s as well as physics warp making the Δt longer.

scimas:
Actually this has happened to me without FAR (basically almost no mods) in JNSQ during an Mk1 capsule reentry (maybe without timewarp? don't remember), but only once. I will see if I can recreate it and report with reproduction steps.

scimas:
It seems that any uncontrolled craft (ie without SAS or similar systems actively controlling attitude) start vibrating more and more throughout reentry. The more the initial deflection from perfect retrograde orientation, the more severe is the amplification. And timewarping does make it worse. You can activate one such attitude control system to bring it under control. But if you deactivate it, the amplifications start taking place again.

While trying to figure out reproduction steps I encountered a crash actually. The game simply crashed on a decouple event with principia's logs saying "time Append at xyz which is before fork time abc". Probably the same as #2507. The issue is marked resolved for next release, but I'm attaching the logs and save anyway in case they're of any use.
[decouple_crash.zip](https://github.com/mockingbirdnest/Principia/files/4462701/decouple_crash.zip)

eggrobin:
I might have found a way to reproduce the issue with no FAR, no RO, and no physics warp: 
[2519.zip](https://github.com/mockingbirdnest/Principia/files/4466280/2519.zip)
Launch, timewarp as soon as you leave the atmosphere, and watch the reentry (the parachutes are inoperant, but their casings affect the trajectory of the rocket).

pleroy:
[...]

PR #2533, opened 2020-04-17
A more stable angular momentum correction.

eggrobin:
This appears to significantly improve the behaviour reported in #2519.

It is not perfect. In the save from https://github.com/mockingbirdnest/Principia/issues/2519#issuecomment-612605437, while we get no uncontrolled spin-up, we get more oscillations than in stock, slowing the rocket down, so that it reaches the water without burning up.

Note that this adds back some traces and toggles for debugging. They will be hidden in a separate pull request (but I would like to keep the plumbing around for now; we may not be quite done with this issue yet).

### DIFF
--- a/ksp_plugin/interface_vessel.cpp
+++ b/ksp_plugin/interface_vessel.cpp
@@ -195,5 +195,24 @@ XYZ __cdecl principia__VesselVelocity(Plugin const* const plugin,
   return m.Return(ToXYZ(plugin->VesselVelocity(vessel_guid)));
 }
 
+void __cdecl principia__SetAngularMomentumConservation(
+    bool const conserve) {
+  journal::Method<journal::SetAngularMomentumConservation> m({conserve});
+  ksp_plugin::PileUp::conserve_angular_momentum = conserve;
+  return m.Return();
+}
+
+char const* __cdecl principia__VesselGetPileUpTrace(
+    Plugin const* const plugin,
+    char const* const vessel_guid) {
+  journal::Method<journal::VesselGetPileUpTrace> m({plugin, vessel_guid});
+  CHECK_NOTNULL(plugin);
+  char const* trace;
+  plugin->GetVessel(vessel_guid)->ForSomePart([&trace](ksp_plugin::Part& part) {
+    trace = part.containing_pile_up()->trace.data();
+  });
+  return m.Return(trace);
+}
+
 }  // namespace interface
 }  // namespace principia
--- a/ksp_plugin/interface_vessel.cpp
+++ b/ksp_plugin/interface_vessel.cpp
@@ -196,9 +196,14 @@ XYZ __cdecl principia__VesselVelocity(Plugin const* const plugin,
 }
 
 void __cdecl principia__SetAngularMomentumConservation(
-    bool const conserve) {
-  journal::Method<journal::SetAngularMomentumConservation> m({conserve});
-  ksp_plugin::PileUp::conserve_angular_momentum = conserve;
+    bool const correct_orientation,
+    bool const correct_angular_velocity,
+    bool const thresholding) {
+  journal::Method<journal::SetAngularMomentumConservation> m(
+      {correct_orientation, correct_angular_velocity, thresholding});
+  ksp_plugin::PileUp::correct_orientation = correct_orientation;
+  ksp_plugin::PileUp::correct_angular_velocity = correct_angular_velocity;
+  ksp_plugin::PileUp::thresholding = thresholding;
   return m.Return();
 }
 
--- a/ksp_plugin/pile_up.cpp
+++ b/ksp_plugin/pile_up.cpp
@@ -442,7 +442,8 @@ void PileUp::DeformPileUpIfNeeded(Instant const& t) {
               ApparentPileUp::origin,
               EquivalentRigidPileUp::origin,
               OrthogonalMap<ApparentPileUp, EquivalentRigidPileUp>::Identity()),
-          apparent_equivalent_angular_velocity,
+          conserve_angular_momentum ? apparent_equivalent_angular_velocity
+                                    : ApparentPileUp::nonrotating,
           ApparentPileUp::unmoving);
   RigidMotion<NonRotatingPileUp, EquivalentRigidPileUp> const
       actual_pile_up_equivalent_rotation(
@@ -451,7 +452,8 @@ void PileUp::DeformPileUpIfNeeded(Instant const& t) {
               EquivalentRigidPileUp::origin,
               OrthogonalMap<NonRotatingPileUp,
                             EquivalentRigidPileUp>::Identity()),
-          actual_equivalent_angular_velocity,
+          conserve_angular_momentum ? actual_equivalent_angular_velocity
+                                    : NonRotatingPileUp::nonrotating,
           NonRotatingPileUp::unmoving);
   RigidMotion<ApparentBubble, NonRotatingPileUp> const
       apparent_bubble_to_pile_up_motion =
@@ -468,6 +470,17 @@ void PileUp::DeformPileUpIfNeeded(Instant const& t) {
         part, apparent_bubble_to_pile_up_motion * apparent_part_rigid_motion);
   }
   apparent_part_rigid_motion_.clear();
+
+  std::stringstream s;
+  s << "Apparent: " << apparent_angular_momentum << "\n"
+    << "norm: " << apparent_angular_momentum.Norm() << "\n"
+    << "Actual: " << angular_momentum_ << "\n"
+    << "norm: " << angular_momentum_.Norm() << "\n"
+    << u8"Δ: "
+    << (angular_momentum_ - Identity<ApparentPileUp, NonRotatingPileUp>()(
+                                apparent_angular_momentum))
+           .Norm();
+  trace = s.str();
 
   MakeEulerSolver(Identity<ApparentPileUp, NonRotatingPileUp>()(inertia_tensor),
                   t);
@@ -594,6 +607,8 @@ PileUpFuture::PileUpFuture(not_null<PileUp const*> const pile_up,
                            std::future<Status> future)
     : pile_up(pile_up),
       future(std::move(future)) {}
+
+bool PileUp::conserve_angular_momentum = true;
 
 }  // namespace internal_pile_up
 }  // namespace ksp_plugin
--- a/ksp_plugin/pile_up.cpp
+++ b/ksp_plugin/pile_up.cpp
@@ -443,8 +443,7 @@ void PileUp::DeformPileUpIfNeeded(Instant const& t) {
   // We apply a rigid rotational correction to the motions of the parts coming
   // from the game (the apparent motions) so as to enforce the conservation of
   // the angular momentum (|angular_momentum_| is authoritative).
-  // Mapping L_apparent to L_actual by a rigid motion leaves can be done in many
-  // ways.
+  // Mapping L_apparent to L_actual by a rigid motion can be done in many ways.
   // We prefer doing some of that correction by a change of attitude, rather
   // than a solely by a change in angular velocity, since, under isotropic
   // conditions, a change in attitude does not alter the physical system (an in
--- a/ksp_plugin/pile_up.cpp
+++ b/ksp_plugin/pile_up.cpp
@@ -477,6 +477,7 @@ void PileUp::DeformPileUpIfNeeded(Instant const& t) {
   auto const α =
       2 * quantities::ArcTan(q.imaginary_part().Norm(), q.real_part());
 
+  // TODO(egg): Use the real Δt.
   if (thresholding && α > ω / (50 * quantities::si::Hertz)) {
     r_apparent = Rotation<ApparentPileUp, EquivalentRigidPileUp>::Identity();
     r_actual = Rotation<NonRotatingPileUp, EquivalentRigidPileUp>::Identity();
--- a/ksp_plugin/pile_up.cpp
+++ b/ksp_plugin/pile_up.cpp
@@ -22,8 +22,8 @@ using base::make_not_null_unique;
 using geometry::AngularVelocity;
 using geometry::BarycentreCalculator;
 using geometry::Bivector;
-using geometry::Frame;
 using geometry::Commutator;
+using geometry::Frame;
 using geometry::Identity;
 using geometry::NonRotating;
 using geometry::Normalize;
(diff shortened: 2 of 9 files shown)


ksp_plugin/pile_up.cpp
```diff
@@ -23,17 +23,24 @@ using geometry::AngularVelocity;
 using geometry::BarycentreCalculator;
 using geometry::Bivector;
 using geometry::Frame;
+using geometry::Commutator;
```
pleroy, line 26: Alphabetize.

ksp_plugin/interface_vessel.cpp
```diff
@@ -195,5 +195,29 @@ XYZ __cdecl principia__VesselVelocity(Plugin const* const plugin,
   return m.Return(ToXYZ(plugin->VesselVelocity(vessel_guid)));
 }
 
+void __cdecl principia__SetAngularMomentumConservation(
```
pleroy, line 198: Please create an issue to clean up these interfaces method as well as the static state in `PileUp`, listing all the stuff that needs to be removed.  Once we are satisfied with the fix, we want an easy way to remove the traces.

ksp_plugin/plugin.cpp
```diff
@@ -658,17 +658,24 @@ void Plugin::SetPartApparentRigidMotion(
     PartId const part_id,
     RigidMotion<RigidPart, World> const& rigid_motion,
     DegreesOfFreedom<World> const& main_body_degrees_of_freedom) {
-  // Define |ApparentBubble| as the reference frame with the axes of
-  // |Barycentric| centred on the current main body.
+  // As a reference frame, |ApparentBubble| differs from |World| only by having
+  // the same axes as |Barycentric| and being nonrotating.  However, there is
+  // another semantic distinction: |Apparent| coordinates are uncorrected data
```
pleroy, line 663: Should this say `ApparentBubble` or is there another confusingly named frame?

ksp_plugin/plugin.cpp
```diff
@@ -658,17 +658,24 @@ void Plugin::SetPartApparentRigidMotion(
     PartId const part_id,
     RigidMotion<RigidPart, World> const& rigid_motion,
     DegreesOfFreedom<World> const& main_body_degrees_of_freedom) {
```
pleroy, line 660: Please remove this parameter since it's now unused.

ksp_plugin_adapter/main_window.cs
```diff
@@ -239,8 +239,32 @@ internal class MainWindow : SupervisedWindowRenderer {
     }
     UnityEngine.GUI.DragWindow();
   }
+  
+  static bool correct_orientation = true;
+  static bool correct_angular_velocity = true;
+  static bool thresholding = true;
```
pleroy, line 245: Could we move these declaration at the place where we normally put data members, i.e., at the end of the class?  Also, explicit `private`.

Another point: it seems to me that these should be persisted in `Load` and `Save`.  As written, when you restart the game you will go back to the default values.  If we believe that users might want to toggle these parameters when they run into trouble, making them non-persistent is going to be horribly confusing.

ksp_plugin/pile_up.cpp
```diff
@@ -419,44 +426,153 @@ void PileUp::DeformPileUpIfNeeded(Instant const& t) {
...
+      apparent_angular_momentum,
+      Identity<NonRotatingPileUp, ApparentPileUp>()(angular_momentum_)));
+  auto const actual_correction_axis = NormalizeOrZero(Commutator(
+      Identity<ApparentPileUp, NonRotatingPileUp>()(apparent_angular_momentum),
+      angular_momentum_));
+
+  // We apply a rigid rotational correction to the motions of the parts coming
+  // from the game (the apparent motions) so as to enforce the conservation of
+  // the angular momentum (|angular_momentum_| is authoritative).
+  // Mapping L_apparent to L_actual by a rigid motion can be done in many ways.
```
pleroy, line 446: Should the `L`s have little hats?

ksp_plugin/pile_up.cpp
```diff
@@ -419,44 +426,153 @@ void PileUp::DeformPileUpIfNeeded(Instant const& t) {
...
+  // We prefer doing some of that correction by a change of attitude, rather
+  // than a solely by a change in angular velocity, since, under isotropic
```
pleroy, line 448: Remove "a".

ksp_plugin/pile_up.cpp
```diff
@@ -419,44 +426,153 @@ void PileUp::DeformPileUpIfNeeded(Instant const& t) {
...
+  // map L̂_apparent to L̂_actual, by rotation around the correction axis defined
+  // above, which is perpendicular to both, and to impart the appropriate
+  // angular velocity correction to correct the norm of L.
+  // This amounts to trusting the direction of the angular momentum with respect
+  // to the vessel as given to us by the game.
+  // However, mapping L̂_apparent to L̂_actual is essentially singular when either
+  // is 0.  We remedy to that by only performing an attitude correction if its
+  // angle would be less than ω Δt, where ω is the smaller of the two equivalent
+  // angular frequencies |L_actual / I| and |L_apparent / I|.
+  // As a result, as either L tends towards 0, so does the greatest attitude
```
pleroy, line 460: s/greatest/largest/?

ksp_plugin/pile_up.cpp
```diff
@@ -419,44 +426,153 @@ void PileUp::DeformPileUpIfNeeded(Instant const& t) {
...
+  // the angle goes to 0).  Since the computation ensured that
+  // |apparent_correction_axis.coordinates() ==
+  //  actual_correction_axis.coordinates()| exactly, we need not worry about the
+  // intermediate essential singularity, and only need to remove it.
+  using EquivalentRigidPileUp = Frame<enum class EquivalentRigidPileUpTag>;
+
+  auto const L̂_apparent = NormalizeOrZero(apparent_angular_momentum);
+  auto const L̂_actual = NormalizeOrZero(angular_momentum_);
+  // NOTE(egg):
+  // — The two sides of this disjunction are always equal;
```
pleroy, line 486: Should this be checked?

ksp_plugin/pile_up.cpp
```diff
@@ -419,44 +426,153 @@ void PileUp::DeformPileUpIfNeeded(Instant const& t) {
...
+      actual_correction_axis == Bivector<double, NonRotatingPileUp>{};
+  bool const on_essential_singularity =
+      L̂_apparent == Bivector<double, ApparentPileUp>{} ||
+      L̂_actual == Bivector<double, NonRotatingPileUp>{};
+
+  bool const trivial_rotations = !correct_orientation ||
+                                 on_removable_singularity ||
+                                 on_essential_singularity;
+
+  Rotation<ApparentPileUp, EquivalentRigidPileUp>
```
pleroy, line 499: `const` this rotation and the next.

ksp_plugin/pile_up.cpp
```diff
@@ -419,44 +426,153 @@ void PileUp::DeformPileUpIfNeeded(Instant const& t) {
...
+
+  // So far we have only dealt with the essential singularity by removing it.
+  // We now need to deal with its neighbourhood, by ensuring that the tentative
+  // correction is not too big compared to the angular velocities involved.
+  Quaternion const q = tentative_attitude_correction.quaternion();
+  // α is the angle of the tentative attitude correction.
+  Angle const α =
+      2 * quantities::ArcTan(q.imaginary_part().Norm(), q.real_part());
+  AngularFrequency const ω = std::min(apparent_equivalent_angular_velocity.Norm(),
+                          actual_equivalent_angular_velocity.Norm());
```
pleroy, line 537: Bad indent.

ksp_plugin/pile_up.cpp
```diff
@@ -419,44 +426,153 @@ void PileUp::DeformPileUpIfNeeded(Instant const& t) {
...
+  auto actual_equivalent_angular_velocity =
+      angular_momentum_ / tentative_attitude_correction(inertia_tensor);
+
+  // So far we have only dealt with the essential singularity by removing it.
+  // We now need to deal with its neighbourhood, by ensuring that the tentative
+  // correction is not too big compared to the angular velocities involved.
+  Quaternion const q = tentative_attitude_correction.quaternion();
+  // α is the angle of the tentative attitude correction.
+  Angle const α =
+      2 * quantities::ArcTan(q.imaginary_part().Norm(), q.real_part());
```
pleroy, line 535: Yuck.  We need a function to extract the angle (and while we are at it, the axis) of a rotation (and a quaternion since that's the same thing).  Please open an issue.

ksp_plugin/pile_up.cpp
```diff
@@ -419,44 +426,153 @@ void PileUp::DeformPileUpIfNeeded(Instant const& t) {
...
+    actual_pile_up_equivalent_rotation =
+        Rotation<NonRotatingPileUp, EquivalentRigidPileUp>::Identity();
+    actual_equivalent_angular_velocity =
+        angular_momentum_ /
+        Identity<ApparentPileUp, NonRotatingPileUp>()(inertia_tensor);
+  }
+
+  bool const correcting_orientation =
+      apparent_pile_up_equivalent_rotation.quaternion() != Quaternion(1) ||
+      actual_pile_up_equivalent_rotation.quaternion() != Quaternion(1);
```
pleroy, line 552: We should also be able to write `rotation != Identity()`.  More stuff for the above-mentioned issue.

ksp_plugin/pile_up.cpp
```diff
@@ -419,44 +426,153 @@ void PileUp::DeformPileUpIfNeeded(Instant const& t) {
...
-      apparent_pile_up_equivalent_rotation(
+      apparent_pile_up_equivalent_motion(
           RigidTransformation<ApparentPileUp, EquivalentRigidPileUp>(
               ApparentPileUp::origin,
               EquivalentRigidPileUp::origin,
-              OrthogonalMap<ApparentPileUp, EquivalentRigidPileUp>::Identity()),
-          apparent_equivalent_angular_velocity,
+              apparent_pile_up_equivalent_rotation.Forget<OrthogonalMap>()),
+          correct_angular_velocity ? apparent_equivalent_angular_velocity
+                                    : ApparentPileUp::nonrotating,
```
pleroy, line 561: Bad indent.

ksp_plugin/pile_up.cpp
```diff
@@ -419,44 +426,153 @@ void PileUp::DeformPileUpIfNeeded(Instant const& t) {
...
+      actual_pile_up_equivalent_motion(
           RigidTransformation<NonRotatingPileUp, EquivalentRigidPileUp>(
               NonRotatingPileUp::origin,
               EquivalentRigidPileUp::origin,
-              OrthogonalMap<NonRotatingPileUp,
-                            EquivalentRigidPileUp>::Identity()),
-          actual_equivalent_angular_velocity,
+              actual_pile_up_equivalent_rotation.Forget<OrthogonalMap>()),
+          correct_angular_velocity ? actual_equivalent_angular_velocity
+                                    : NonRotatingPileUp::nonrotating,
```
pleroy, line 570: Same.

ksp_plugin/pile_up.cpp
```diff
@@ -419,44 +426,153 @@ void PileUp::DeformPileUpIfNeeded(Instant const& t) {
...
           NonRotatingPileUp::unmoving);
   RigidMotion<ApparentBubble, NonRotatingPileUp> const
       apparent_bubble_to_pile_up_motion =
-          actual_pile_up_equivalent_rotation.Inverse() *
-          apparent_pile_up_equivalent_rotation *
+          actual_pile_up_equivalent_motion.Inverse() *
```
pleroy, line 574: Would this fit on the previous line?

ksp_plugin/pile_up.cpp
```diff
@@ -469,6 +585,35 @@ void PileUp::DeformPileUpIfNeeded(Instant const& t) {
...
+  std::stringstream s;
+  s << "Apparent: " << apparent_angular_momentum << "\n"
+    << "norm: " << apparent_angular_momentum.Norm() << "\n"
+    << "Actual: " << angular_momentum_ << "\n"
+    << "norm: " << angular_momentum_.Norm() << "\n"
+    << u8"|Lap-Lac|: "
+    << (angular_momentum_ - Identity<ApparentPileUp, NonRotatingPileUp>()(
+                                apparent_angular_momentum))
+           .Norm() << "\n"
+    << u8"|Lap|-|Lac|: "
```
pleroy, line 597: Why is there an `u8` here?  And 4 lines above?

ksp_plugin/pile_up.cpp
```diff
@@ -469,6 +585,35 @@ void PileUp::DeformPileUpIfNeeded(Instant const& t) {
...
+    << u8"α: " << α / quantities::si::Degree << u8"°\n"
+    << u8"|ωap|: "
+    << apparent_equivalent_angular_velocity.Norm() /
+           (2 * π * Radian / quantities::si::Minute)
+    << " rpm\n"
+    << u8"|ωac|: "
+    << actual_equivalent_angular_velocity.Norm() /
+           (2 * π * Radian / quantities::si::Minute)
+    << " rpm\n"
+    << (correcting_orientation ? "CORRECTING ORIENTATION" : u8"—");
```
pleroy, line 614: Do we need to yell?

ksp_plugin/plugin.cpp
eggrobin: I mean `Apparent` as in the prefix, both `ApparentBubble` and `ApparentPileUp`.

ksp_plugin_adapter/main_window.cs
eggrobin: We really don’t want users to be playing with this; many configurations lead to absurd behaviour.
We want to keep playing with it though (and we don’t want to persist them to their non-default state, we that hid a bug during Frobenius development already…).

ksp_plugin/pile_up.cpp
eggrobin: Not these. The rigid motion includes the angular velocity, so it maps the bareheaded angular momenta to each other.

ksp_plugin/pile_up.cpp
eggrobin: If we want to check something it is the equality of the `coordinates` commented above, not a random consequence thereof.

ksp_plugin/pile_up.cpp
eggrobin: No, it is not const.

ksp_plugin/pile_up.cpp
eggrobin: #2534.

ksp_plugin/pile_up.cpp
eggrobin: Added to #2534.

ksp_plugin/pile_up.cpp
eggrobin: YES

ksp_plugin/interface_vessel.cpp
eggrobin: #2535.

ksp_plugin/plugin.cpp
pleroy: Maybe `Apparent...` then.

ksp_plugin/pile_up.cpp
eggrobin: No.

nepphhh:
Tested the binaries in KSP v1.8.1 in a mess of an RP-1 career/dev mod environment. Thanks to egg for providing them. 

Results: https://gfycat.com/jubilantmeaslydartfrog

PR #2536, opened 2020-04-18
Display the flight planner and the orbit analyser in the tracking station

### LINKED ISSUE #2531: Make orbital analysis window available in tracking station
[Description by @eggrobin, this issue originally had none]
The orbit analyser is currently only available in map view.

We should make it available for the `PredictedVessel`, i.e., the selected vessel in the tracking station, and the active vessel in map view.

pleroy:
These two seem to make the tracking station much more useful.

The implementation involves a bit of code restructuring where we use `PredictedVessel` systematically in many places that were using `ActiveVessel`.  Overall this factors out a bit of logic and fixes a buglet where the orbit analyser in the tracking station would show a stump of a window.

There is a remaining know issue for a Kerbal in EVA: if a flight plan is constructed in the tracking station, that flight plan is lost when switching to the map view.  That's because the Kerbal becomes unready for a couple of frames.  It doesn't happen in the other direction.  We call this "the invariants of KSP".  The ultimate fix would be to keep some of the state of unmanaged vessels, but that's a long term effort.

Fix #2531.

### DIFF
--- a/ksp_plugin_adapter/ksp_plugin_adapter.cs
+++ b/ksp_plugin_adapter/ksp_plugin_adapter.cs
@@ -322,7 +322,11 @@ public partial class PrincipiaPluginAdapter
   }
 
   private Vessel PredictedVessel() {
-    Vessel vessel = FlightGlobals.ActiveVessel ?? space_tracking?.SelectedVessel;
+    if (!PluginRunning()) {
+      return null;
+    }
+    Vessel vessel =
+        FlightGlobals.ActiveVessel ?? space_tracking?.SelectedVessel;
     string vessel_guid = vessel?.id.ToString();
     if (vessel_guid != null && plugin_.HasVessel(vessel_guid)) {
       return vessel;
@@ -1591,7 +1595,7 @@ public partial class PrincipiaPluginAdapter
     // The only timing that satisfies these constraints is BetterLateThanNever
     // in LateUpdate.
     string main_vessel_guid = PredictedVessel()?.id.ToString();
-    if (MapView.MapIsEnabled && main_vessel_guid != null && PluginRunning()) {
+    if (MapView.MapIsEnabled && main_vessel_guid != null) {
       XYZ sun_world_position = (XYZ)Planetarium.fetch.Sun.position;
       RenderPredictionMarkers(main_vessel_guid, sun_world_position);
       string target_id =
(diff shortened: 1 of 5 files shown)


ksp_plugin_adapter/ksp_plugin_adapter.cs
```diff
@@ -1587,8 +1591,7 @@ class PartCentredForceHolder {
     // The only timing that satisfies these constraints is BetterLateThanNever
     // in LateUpdate.
     string main_vessel_guid = PredictedVessel()?.id.ToString();
-    if (MapView.MapIsEnabled && main_vessel_guid != null &&
-        PluginRunning() && plugin_.HasVessel(main_vessel_guid)) {
+    if (MapView.MapIsEnabled && main_vessel_guid != null && PluginRunning()) {
```
eggrobin, line 1594: `PredictedVessel` requires `PlugginRunning`, but we are only testing that after calling it; either it was, and remains, useless here, or this is a fatal failure.

Add `PluginRunning`  to the conjunction in `PredictedVessel`.

ksp_plugin_adapter/window_renderer.cs
```diff
@@ -260,6 +265,41 @@ public interface ISupervisor {
...
+                                           PredictedVessel predicted_vessel,
+                                           params UnityEngine.GUILayoutOption[]
+                                               options) : base(
+      supervisor,
+      options) {
+    predicted_vessel_ = predicted_vessel;
+  }
+
+  // A helper for implementing the RenderButton() method of the subclasses.
+  protected void RenderButton(string text) {
```
eggrobin, line 287: This does not seem specific to vessel-related windows: the plotting frame selectors have a button too.

Should there be `RenderButton` in the base class, overriden here (not necessarily in this pull request, but maybe file a cleanup issue).

ksp_plugin_adapter/ksp_plugin_adapter.cs
pleroy: Done.

ksp_plugin_adapter/window_renderer.cs
pleroy: Look at the code of this function: it is dependent on the existence of a predicted vessel.  That's why it belongs here.  This was horribly broken in the `OrbitAnalyser`, as the decision to hide was made in the middle of `RenderWindow`, a sure recipe for layout/repaint inconsistencies.  This proves that the logic is non-obvious and belongs to a common place.

PR #2539, opened 2020-04-20
Add a C# class PrincipiaTimeSpan to gather time formatting and parsing

### LINKED ISSUE #2520: Add daylength tweakable
I play with JNSQ, one of its minor effects is a 12 hour long day. The planner assumes day to be the "default" one - 6h per day. It brings some inconvenience in planning, I would like to have a 12 hours day in the planner too. It can be a number in a .cfg or a field in planner, does not really matter where we change this parameter. I guess it is a simple-to-do QoL and want to see it in the next release.

pleroy:
Make it use the correct KSP APIs to figure out the length of the day, etc.  Tested in game with stock and JNSQ.

Fix #2520.


ksp_plugin_adapter/time_span.cs
```diff
@@ -0,0 +1,138 @@
...
+      return false;
+    }
+    time_span = new PrincipiaTimeSpan(d * date_time_formatter.Day +
+                                      h * date_time_formatter.Hour +
+                                      min * date_time_formatter.Minute + s);
+    return true;
+  }
+
+  public static int day_duration => date_time_formatter.Day;
+  public static string day_symbol => is_stock_day ? "d6" : "d";
```
eggrobin, line 125:
```C#
  public static string day_symbol =>
      hour_divides_day && !is_si_time
          ? "d" + (int)(date_time_formatter.Day / date_time_formatter.hour)
          : "d";


  private static bool hour_divides_day =>
      (int)(date_time_formatter.Day / date_time_formatter.hour) * date_time_formatter.hour ==
          date_time_formatter.Day;
  private static is_si_time =>
      date_time_formatter.Minute == 60 &&
      date_time_formatter.Hour == 60 * 60 &&
      date_time_formatter.Day == 24 * 60 * 60;
```

ksp_plugin_adapter/orbit_analyser.cs
```diff
@@ -403,19 +359,21 @@ internal class OrbitAnalyser : VesselSupervisedWindowRenderer {
...
-      min_value        : 10,
-      max_value        : double.PositiveInfinity,
-      formatter        : Formatters.FormatMissionDuration,
-      parser           : Formatters.TryParseMissionDuration,
-      label_width      : 2,
-      field_width      : 5) {
-      value = 7 * 24 * 60 * 60
-  };
+  private readonly DifferentialSlider mission_duration_ =
+      new DifferentialSlider(label            : "Duration",
```
eggrobin, line 363: Can we break the line after the `(` to avoid folding the actual parameters?

ksp_plugin_adapter/time_span.cs
pleroy: Actually I made the change, and I foundthe output with JNSQ unnecessarily noisy:
```
014 d12 00 h 00 min
```
So I changed it again to only display one digit after `d`.  Thus `d3`, `d6`, `d9` but just `d` for JNSQ.

PR #2545, opened 2020-04-26
Try to fix problems with builds

pleroy:
1. The diurnal build tries to execute a test that has flags, let's make it secular.
1. The `make` file doesn't pull the `mathematica` objects, resulting in link failures.


pleroy:
retest this please

PR #2547, opened 2020-04-26
An RAII object for Mathematica logging

pleroy:


mathematica/mathematica_test.cpp
```diff
@@ -357,5 +357,13 @@ TEST_F(MathematicaTest, ExpressIn) {
 #endif
 }
 
+TEST_F(MathematicaTest, Logger) {
+  Logger logger(TEMP_DIR / "mathematice_test.wl");
```
eggrobin, line 361: Vocative?

*Quousque tandem abutere, mathematice, patientia nostra?*

mathematica/mathematica_test.cpp
```diff
@@ -357,5 +357,13 @@ TEST_F(MathematicaTest, ExpressIn) {
...
+  logger.Append("a", std::vector{1, 2, 3});
+  logger.Append("b", 4 * Metre / Second);
+  logger.Append("a", F::origin);
+  // Go check the file.
```
eggrobin, line 365: Could we maybe do that? Wrap the `logger` in a block so it is destroyed, then
```C++
EXPECT_THAT(
  (std::stringstream{} << std::ifstream(TEMP_DIR / "mathematice_test.wl").rdbuf()).str(),
  Eq(R"(\
(*Something equivalent to*)
a={{1, 2, 3}, {0, 0, 0}};
b=Quantity[4, "m/s"];
)"));
```

Issue #2548, opened 2020-04-27
KSP Loading Issue


CZ-81H-YZ5Block5:
I already installed the C++ 2015-2019 Redist that Principia required，but the Principia still can't load properly
 Here's the texts in the Principia Dialog:
The Principia DLL failed to load.
An unknown error occurred; detected OS Microsoft Windows NT 10.0.14393.0 64-bit; tried loading dll at 'GameData\Principia\x64\principia.dll'. Note that the Microsoft Visual C++ 2015-2019 Redistributable (x64) - 14.22.27821 are required.

Warning: don't load a Principia save before you have fixed this error; it might get damaged.

the glog file of principia are not create.
[KSP.log](https://github.com/mockingbirdnest/Principia/files/4537988/KSP.log)

pleroy:
I think we'll need more information to figure out what's happening.  It would be good if you could:
1. Download and install [Process Monitor](https://docs.microsoft.com/en-gb/sysinternals/downloads/procmon) from the Microsoft site.
1. Start it.
1. Start KSP.
1. Use the `Process` menu in Process Monitor to filter `Process Name contains KSP`.
1. Save the events in a file and send it to us.

CZ-81H-YZ5Block5:
https://share.weiyun.com/5VxiuUp
I don't know what format do you need,so here's the CSV Format.

pleroy:
Are you by any chance running KSP using a shortcut?  If so, what is the "Start in:" directory for the shortcut?

The search for the Principia DLL normally starts at the directory containing `KSP_x64.exe` and appends `GameData\Principia\x64\principia.dll`.  However, if using a shortcut, the search starts at the "Start in:" directory, which may not be the place where Principia is installed.

In your logs I see confusing references to two KSP installs named `E:\National Department of Space and Aeroplanes\Kerbal Space Program-Stock- 多体` (the one that contains Principia, I presume) and `E:\National Department of Space and Aeroplanes\Kerbal Space Program-Stock- 澶氫綋`.  I suspect that you have shortcuts that straddle the two installs.

CZ-81H-YZ5Block5:
NO,I don't start KSP by a shortcunt,and I moved the whole KSP to E:\Kerbal Space Program-Stock- 多体\KSP_x64.exe,but it still doesn's work.And here's the PML file :https://share.weiyun.com/5bHi2aF

CZ-81H-YZ5Block5:
and here's the KSP.log
[...]

PR #2549, opened 2020-04-28
Detect non-ASCII paths as they confuse Mono

pleroy:
Will help users who run into #2548.  Message:
```
The Principia DLL failed to load.
C:\Program Files\Kerbal Space Program\1.9.1- 多体 contains the non-ASCII character 多; this is known to confuse Mono.

Warning: don't load a Principia save before you have fixed this error; it might get damaged.
```

### DIFF
--- a/ksp_plugin_adapter/loader.cs
+++ b/ksp_plugin_adapter/loader.cs
@@ -48,6 +48,13 @@ internal static class Loader {
              string.Join("', '", possible_dll_paths) + "' in directory '" +
              Directory.GetCurrentDirectory() + "'.";
     }
+    foreach (char c in Directory.GetCurrentDirectory()) {
+      if (c >= 128) {
+        return Directory.GetCurrentDirectory() +
+               " contains the non-ASCII character " + c +
+               "; this is known to confuse Mono.";
+      }
+    }
     try {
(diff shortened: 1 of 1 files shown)


ksp_plugin_adapter/loader.cs
```diff
@@ -48,6 +48,13 @@ internal static class Loader {
              string.Join("', '", possible_dll_paths) + "' in directory '" +
              Directory.GetCurrentDirectory() + "'.";
     }
+    foreach (char c in Directory.GetCurrentDirectory()) {
+      if (c >= 128) {
+        return Directory.GetCurrentDirectory() +
+               " contains the non-ASCII character " + c +
+               "; this is known to confuse Mono.";
+      }
+    }
```
eggrobin, line 57: The bug could be Windows-only and is probably system-dependent (for instance, on recent versions of Windows, the code page can be 65001, in which case interpreting UTF-8 bytes in the current code page woould work).

Let’s flag this potential issue only if an error occurs, as we do with `is_cxx_installed`, instead of making it prevent us from even trying.

pleroy: #2554.

PR #2550, opened 2020-05-02
Add a PID to smoothen the apparent angular momentum given by KSP

pleroy:
With the default parameters, it avoids the glitches seen in `Fubini oscillator`.

Another attempt at alleviating #2519.


base/flags.hpp
```diff
@@ -8,6 +8,11 @@
 namespace principia {
 namespace base {
 
+// Example of usage: create a file zfp.cfg containing:
+// principia_flags
+// {
```
eggrobin, line 13: Why the newline?

ksp_plugin/pile_up.cpp
```diff
@@ -41,12 +43,38 @@ using physics::RigidMotion;
...
 using ::std::placeholders::_3;
 
+// NOTE(phl): The default values were tuned on "Fubini oscillator".
+constexpr double kp_default = 0.65;
+constexpr Inverse<Time> ki_default = 0.03 / Second;
+constexpr Time kd_default = 0.0025 * Second;
+
+// Configuration example:
+//   principia_flags
+//   {
```
eggrobin, line 65: Remove the newline.

ksp_plugin/pile_up.cpp
```diff
@@ -415,15 +451,19 @@ void PileUp::DeformPileUpIfNeeded(Instant const& t) {
     CHECK(Contains(apparent_part_rigid_motion_, part));
   }
 
-  using ApparentPileUp = Frame<enum class ApparentPileUpTag, NonRotating>;
+  Time const Δt = t - psychohistory_->back().time;
```
eggrobin, line 456: Clear the PID if Δt changes.

ksp_plugin/pile_up.cpp
```diff
@@ -41,12 +43,38 @@ using physics::RigidMotion;
...
+
+// Configuration example:
+//   principia_flags
+//   {
+//     kp = 0.65
+//     ki = 0.03 / s
+//     kd = 0.0025 s
+//   }
+template<typename Q>
+Q GetPidFlagOr(std::string_view const name, Q const default) {
```
eggrobin, line 71: Principia style is to keep acronyms in upper case in camel case identifiers, see `ICRS`, `ParseTT`, `ParseTAI`, etc. This should say PID.

I would want this to return an optional and drop the `Or` but obviously this does not work because we cannot deduce Q, unless we build a mechanism to declare flags. Open an issue to track that, we don’t need it yet but we likely will at some point, the flags are inevitably becoming useful.

ksp_plugin/pile_up.hpp
```diff
@@ -143,6 +145,9 @@ class PileUp {
   using AppendToPartTrajectory =
       void (Part::*)(Instant const&, DegreesOfFreedom<Barycentric> const&);
 
+  // TODO(egg): Document this frame.
+  using ApparentPileUp = Frame<enum class ApparentPileUpTag, NonRotating>;
```
eggrobin, line 149: The axes are those of `Barycentric`. The origin is the centre of mass of the pile up. This frame is distinguished from `NonRotatingPileUp` in that it is used to hold uncorrected (apparent) coordinates given by the game, before the enforcement of conservation laws; see also `ApparentBubble`.

ksp_plugin/pile_up.hpp
```diff
@@ -236,6 +241,12 @@ class PileUp {
   // Called in the destructor.
   std::function<void()> deletion_callback_;
 
+  // A PID used to smoothen the value of the apparent angular momentum obtained
+  // from KSP.
+  PID<Bivector<AngularMomentum, ApparentPileUp>,
+      /*horizon=*/25,
+      /*finite_difference_order=*/5> pid_;
```
eggrobin, line 248: `apparent_angular_momentum_`, or `apparent_angular_momentum_controller_`; no need to speak Hungarian, but it would be useful to document what is being controlled.

numerics/pid.hpp
```diff
@@ -0,0 +1,49 @@
...
+  PID(double kp, Inverse<Time> const& ki, Time const& kd);
+
+  // Clears the state of the PID.
+  void Clear();
+
+  // Adds the error between the two values to the state of the PID and
+  // returns an apparent value derived from the actual value and the control
+  // variable.
+  Value ComputeValue(Value const& apparent,
+                     Value const& actual,
```
eggrobin, line 30: Is `apparent` and `actual` the standard terminology here?

numerics/pid.hpp
```diff
@@ -0,0 +1,49 @@
+﻿
+#pragma once
+
+#include <deque>
+
+#include "quantities/named_quantities.hpp"
+#include "quantities/quantities.hpp"
+
+namespace principia {
+namespace numerics {
```
eggrobin, line 10: Should this be in `physics`? It doesn’t feel like numerical analysis (there would be a bad argument for putting it in `integrators`, since it has an equally bad quadrature method in the middle of the implementation, but let’s not go there).

numerics/pid_body.hpp
```diff
@@ -0,0 +1,64 @@
...
+  errors_.pop_front();
+  --size;
+
+  // The last error.
+  Value proportional = errors_.back();
+
+  // Compute the integral of the errors.  This is the dumbest possible algorithm
+  // because we do not care too much about the accuracy.
+  Product<Value, Time> const integral =
+      std::accumulate(errors_.cbegin(), errors_.cend(), Value{}) * Δt;
```
eggrobin, line 48: > we do not care too much about the accuracy.

Given how low the sample rate seems to be compared to the frequency, it may be beneficial to have a decent quadrature. Open an issue to track this so that, if we ever implement a quadrature, we remember to use it here.

ksp_plugin/pile_up.cpp
pleroy: #2551

ksp_plugin/pile_up.hpp
pleroy: `apparent_angular_momentum_controller_`

numerics/pid.hpp
pleroy: The standard terminology is "process variable/value/parameter" and "setpoint/set point/set-point".  Changing accordingly.

numerics/pid.hpp
pleroy: Dunno, it doesn't seem very physical.  Think of it as applying a Laplace transform.

numerics/pid_body.hpp
pleroy: #2552.

ksp_plugin/pile_up.cpp
pleroy: Done in the `PID`, but with a huge tolerance because I don't trust KSP/Unity to be super-precise.  What we are after is physics warp anyway.

numerics/pid.hpp
```diff
@@ -23,18 +23,21 @@ class PID {
...
+  // of the PID and returns an updated process variable derived from the
+  // set-point and the control variable.
+  Value ComputeValue(Value const& process_variable,
+                     Value const& set_point,
                      Time const& Δt);
 
 private:
   double const kp_;
   Inverse<Time> const ki_;
   Time const kd_;
```
eggrobin, line 37: The types become:
```C++
Derivative<ControlOutput, Controlled> kp_;
Derivative<ControlOutput, Product<Controlled, Time>> ki_;
Derivative<ControlOutput, Variation<Controlled>> kd_;
```

numerics/pid_test.cpp
```diff
@@ -0,0 +1,75 @@
...
+  // TODO(phl): The large relative error is because our integration is lame.
+  // Fix it, one day.
+  EXPECT_THAT(pid.ComputeValue(apparent(20), actual(20), 1 * Second),
+              RelativeErrorFrom(actual(20) + kp * proportional + ki * integral +
+                                    kd * derivative,
+                                IsNear(0.08_⑴)));
+}
+
+TEST_F(PIDTest, Geometry) {
+  PID<Velocity<F>, /*horizon=*/10, /*finite_difference_order=*/3> pid(
```
eggrobin, line 68: In this test, control a `Position`, and make the control output a `Current`.

quantities/parser_body.hpp
```diff
@@ -216,12 +216,16 @@ inline Unit ParseQuotientUnit(std::string const& s) {
     // Not a quotient.
     return ParseProductUnit(s);
   } else {
-    // A quotient.  Parse each half.
+    // A quotient.  Parse each half.  Note that there may not be a left half for
+    // input like 1.23 / s.
```
eggrobin, line 220: Update the grammar in the `hpp` and in https://github.com/mockingbirdnest/Principia/wiki/Principia-configuration-files.

numerics/pid.hpp
```diff
@@ -23,18 +23,21 @@ class PID {
...
 
-  // Adds the error between the two values to the state of the PID and
-  // returns an apparent value derived from the actual value and the control
-  // variable.
-  Value ComputeValue(Value const& apparent,
-                     Value const& actual,
+  // Adds the error between the process variable and the set-point to the state
+  // of the PID and returns an updated process variable derived from the
+  // set-point and the control variable.
+  Value ComputeValue(Value const& process_variable,
```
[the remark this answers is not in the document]
pleroy: Calling the types `Input` and `Output`, it's clearer.

numerics/pid_body.hpp
```diff
@@ -27,14 +27,24 @@ void PID<Value, horizon, finite_difference_order>::Clear() {
...
+    Clear();
+  }
+  previous_Δt_ = Δt;
+
+  errors_.push_back(process_variable - set_point);
   int size = errors_.size();
   if (size <= horizon) {
     // If we don't have enough error history, just return our input.
-    return apparent;
+    return process_variable;
```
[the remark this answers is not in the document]
pleroy: Experimenting is very time-consuming, and this part is used for 500 milliseconds at the beginning anyway.

numerics/pid.hpp
pleroy: I don't think these are derivatives, they are just quotients.  There is no derivation in the canonical PID formula.

eggrobin: OK, but note that then you need `Quotient<ControlOutput, Difference<Controlled>>`.

pleroy: Using `Normed` as discussed.

numerics/pid_test.cpp
pleroy: A vector of currents.

PR #2554, opened 2020-05-03
Don't protest on a non-ASCII path if the DLL can be loaded

pleroy:
An afterthought on #2549.

### DIFF
--- a/ksp_plugin_adapter/loader.cs
+++ b/ksp_plugin_adapter/loader.cs
@@ -48,11 +48,13 @@ internal static class Loader {
              string.Join("', '", possible_dll_paths) + "' in directory '" +
              Directory.GetCurrentDirectory() + "'.";
     }
+    string non_ascii_path_error = null;
     foreach (char c in Directory.GetCurrentDirectory()) {
       if (c >= 128) {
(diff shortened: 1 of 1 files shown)


ksp_plugin_adapter/loader.cs
```diff
@@ -48,11 +48,13 @@ internal static class Loader {
...
              Directory.GetCurrentDirectory() + "'.";
     }
+    string non_ascii_path_error = null;
     foreach (char c in Directory.GetCurrentDirectory()) {
       if (c >= 128) {
-        return Directory.GetCurrentDirectory() +
-               " contains the non-ASCII character " + c +
-               "; this is known to confuse Mono.";
+        non_ascii_path_error = Directory.GetCurrentDirectory() +
+                             " contains the non-ASCII character " + c +
```
eggrobin, line 55: This error message will look weird if `c` is a combining mark, and will be invalid if it is a surrogate.
Use https://docs.microsoft.com/en-gb/dotnet/api/system.globalization.stringinfo.gettextelementenumerator to iterate over the extended grapheme clusters of `Directory.GetCurrentDirectory()`, then look for a code unit above 127 within each cluster.

PR #2566, opened 2020-05-10
Three new API functions

### LINKED ISSUE #2542: API Request: query the history
Kerbalism Contracts implements a couple of new contract parameters that operate mostly on vessel positions in relation to other vessels and/or celestial bodies. The nature of some of those contracts makes it necessary to validate said parameters based on past information, this happens when the user is time warping at very high speeds, causing KSP to update at time intervals that exceed a sane maximum.

To do that, KerCon uses sub-sampling to recalculates past positions of vessels, celestial bodies and waypoints on those bodies based on the KSP orbits. In the presence of Principia, this will clearly provide wrong results, possibly wrong enough to matter.

Since Principia already has a history for vessels (and, I presume, celestial bodies) I'd like to have a way to query the following information from principia:

- The true position of a vessel at time T
- The true position of a celestial body at time T
- The true position of a waypoint (lat, lon) on the surface of a celestial body, at time T

Where T will always be `Math.Floor(Planetarium.GetUniversalTime() - d)`, and d will always be anywehere in the range `10 .. timestep` where timestep depends on the current warp speed.

The KSP equivalent using the stock orbit class is implemented in https://github.com/Kerbalism/KerbalismContracts/blob/master/src/KerbalismContracts/SubRequirements/Evaluation/EvaluationContext.cs

If principias history does not internally use a resolution high enough for one value for every second, I'd like to also have a way to query the available resolution so that I can choose my Ts accordingly.

eggrobin:
Fix #2542.

Tested in-game; if we want to add the test utility (a separate DLL which just calls those endpoints by reflection and dumps the output or exceptions to a window) we should do it in a separate pull request.

Note that despite the work on the UTF-8 marshalers we still lack the capability to produce strings as struct fields, so Status still lacks a message. This may be a lower-hanging fruit than it was when that TODO was written though.


ksp_plugin/interface_external.cpp
```diff
@@ -260,5 +351,42 @@ Status __cdecl principia__ExternalGetNearestPlannedCoastDegreesOfFreedom(
...
+  journal::Method<journal::ExternalVesselGetPosition> m{
+      {plugin,
+       vessel_guid,
+       time},
+      {position}};
+  if (plugin == nullptr) {
+    return m.Return(
+        MakeStatus(Error::INVALID_ARGUMENT, "|plugin| must not be null"));
+  }
+  Instant const t = FromGameTime(*plugin, time);
```
pleroy, line 368: Move down, closer to the place where you need it.

ksp_plugin/interface_external.cpp
```diff
@@ -43,6 +45,95 @@ Status OK() {
...
+  journal::Method<journal::ExternalCelestialGetPosition> m{
+      {plugin,
+       body_index,
+       time},
+      {position}};
+  if (plugin == nullptr) {
+    return m.Return(
+        MakeStatus(Error::INVALID_ARGUMENT, "|plugin| must not be null"));
+  }
+  Instant const t = FromGameTime(*plugin, time);
```
pleroy, line 62: Move down, closer to the place where you need it.

ksp_plugin/interface_external.cpp
```diff
@@ -43,6 +45,95 @@ Status OK() {
...
+       planetocentric_latitude_in_degrees,
+       planetocentric_longitude_in_degrees,
+       radius,
+       time},
+      {position}};
+  if (plugin == nullptr) {
+    return m.Return(
+        MakeStatus(Error::INVALID_ARGUMENT, "|plugin| must not be null"));
+  }
+  Instant const t = FromGameTime(*plugin, time);
```
pleroy, line 106: Move down, closer to the place where you need it.

ksp_plugin/interface_external.cpp
```diff
@@ -43,6 +45,95 @@ Status OK() {
...
+    return m.Return(
+        MakeStatus(Error::OUT_OF_RANGE,
+                   (std::stringstream{}
+                    << "|time| " << t << " does not lie within the domain ["
+                    << trajectory.t_min() << ", " << trajectory.t_max()
+                    << "] of the trajectory of " << celestial.body()->name())
+                       .str()));
+  }
+  using Surface = Frame<enum class SurfaceTag>;
+  auto const to_world_axes =
```
pleroy, line 124: Use a real type instead of `auto`.

pleroy:
retest this please

PR #2584, opened 2020-05-20
Point to the planned initial Frenet frame before ignition

### LINKED ISSUE #2577: The manœuvre node marker should show the ignition orientation prior to ignition for guided burns
Following the Frenet frame prior to ignition is useless, even for manual execution, and, now that MechJeb supports Principia burns (https://github.com/MuMech/MechJeb2/pull/1264), it leads to 10 min of confusion as MechJeb desperately attempts to remain aligned to a rotating target (interrupting its timewarp to do so) prior to ignition.

eggrobin:
On ignition, switch to tracking the current actual Frenet frame (current behaviour).

Fix #2577; all Principia burns can now be executed by MechJeb, and MechJeb warps properly.

### DIFF
--- a/ksp_plugin/interface_flight_plan.cpp
+++ b/ksp_plugin/interface_flight_plan.cpp
@@ -255,10 +255,14 @@ XYZ __cdecl principia__FlightPlanGetGuidance(Plugin const* const plugin,
     result = plugin->renderer().BarycentricToWorld(
                  plugin->PlanetariumRotation())(manœuvre.InertialDirection());
   } else {
-    result = plugin->renderer().FrenetToWorld(
-                 *plugin->GetVessel(vessel_guid),
-                 *manœuvre.frame(),
-                 plugin->PlanetariumRotation())(manœuvre.direction());
+    result = plugin->CurrentTime() < manœuvre.initial_time()
+                 ? plugin->renderer().BarycentricToWorld(
+                       plugin->PlanetariumRotation())(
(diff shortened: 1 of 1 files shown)


ksp_plugin/interface_flight_plan.cpp
```diff
@@ -255,14 +255,14 @@ XYZ __cdecl principia__FlightPlanGetGuidance(Plugin const* const plugin,
...
-                       *manœuvre.frame(),
-                       plugin->PlanetariumRotation())(manœuvre.direction());
+    result = (plugin->CurrentTime() < manœuvre.initial_time()
+                  ? plugin->renderer().BarycentricToWorld(
+                        plugin->PlanetariumRotation()) *
+                        manœuvre.FrenetFrame()
+                  : plugin->renderer().FrenetToWorld(
+                        *plugin->GetVessel(vessel_guid),
+                        *manœuvre.frame(),
+                        plugin->PlanetariumRotation()))(manœuvre.direction());
```
pleroy, line 265: Note that the two branches would be more similar (and the code more transparent) if we had a `FrenetToBarycentric` instead of a `FrenetToWorld` because the barycentric-to-world conversion could be factored out.

eggrobin: Good point.

This would also be far more readable if Manœuvre::FrenetFrame were called Manœuvre::InitialFrenetFrame.

Let’s do these cleanups in a subsequent pull request.

PR #2596, opened 2020-05-30
A better syntax for the selectors of RigidMotion and AcceleratedRigidMotion

### LINKED ISSUE #2593: RigidMotion::angular_velocity_of_to_frame is confusing
It should be renamed to `RigidMotion::angular_velocity_of` and templatized on the frame (which can be `ToFrame` or `FromFrame`.

pleroy:
Fix #2593.

### DIFF
--- a/physics/rigid_motion.hpp
+++ b/physics/rigid_motion.hpp
@@ -28,9 +28,30 @@ using geometry::RigidTransformation;
 using geometry::Vector;
 using geometry::Velocity;
 using quantities::Acceleration;
+using quantities::AngularAcceleration;
 using quantities::Length;
 using quantities::Variation;
 using quantities::si::Radian;
+
+// A trait to determine if Frame is FromFrame or ToFrame and return the other
+// one as the |type| member.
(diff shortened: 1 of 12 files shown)


physics/rigid_motion.hpp
```diff
@@ -28,19 +28,42 @@ using geometry::RigidTransformation;
...
 using quantities::Variation;
 using quantities::si::Radian;
 
+// A trait to determine if Frame is FromFrame or ToFrame and return the other
+// one as the |type| member.
+template<typename Frame, typename FromFrame, typename ToFrame>
+struct other_frame;
+
+template<typename Frame>
+struct other_frame<Frame, Frame, Frame> {
```
eggrobin, line 42: Do we need this one? `ToFrame` = `FromFrame` makes `angular_velocity_of<Frame>` rather confusing.

pleroy: We need this one.  There are tests that compute `r * r.Inverse()` and check the result.  Without this specialization the other two become ambiguous.

PR #2603, opened 2020-06-11
Automated primary detection

eggrobin:
When there is an ambiguity (e.g., an Earth orbit is a heavily perturbed Solar orbit), use the body for which the current osculating orbit is shortest. This appears to work well whenever there is a reasonably-Keplerian orbit.

For fancier ambiguous situations (e.g., the maccollo square), using the shortest sidereal period would likely be more robust, as with this method we will switch between bodies along the orbit; however, there is no obviously right answer in such cases anyway: neither orbit is even close to Keplerian, and the numbers produced by the analyser for either are of questionable use.
Computing the mean sidereal period for all bodies is very costly; let us pick the cheap option.


ksp_plugin/orbit_analyser.cpp
```diff
@@ -106,7 +107,9 @@ void OrbitAnalyser::RepeatedlyAnalyseOrbit() {
          trajectory.back().time <
          parameters->first_time + parameters->mission_duration;
          t += parameters->mission_duration / 0x1p10) {
+      auto const flow_status = ephemeris_->FlowWithFixedStep(t, *instance);
```
pleroy, line 110: Unused variable?  This code is weird.

ksp_plugin/orbit_analyser.hpp
```diff
@@ -67,12 +67,11 @@ class OrbitAnalyser {
...
 
    private:
-    Analysis(Instant const& first_time,
-             not_null<RotatingBody<Barycentric> const*> primary);
+    explicit Analysis(Instant const& first_time);
 
     Instant first_time_;
     Time mission_duration_;
-    not_null<RotatingBody<Barycentric> const*> primary_;
+    RotatingBody<Barycentric> const* primary_ = nullptr;
```
pleroy, line 74: Please fix the usages of this member.  In particular, `primary()` must probably return an optional and  the caller(s) need to change.

PR #2604, opened 2020-06-17
Look at Rigidbody.centerOfMass.

### LINKED ISSUE #2560: Take CoM offset into account
In Pricipia Fubini, offsets to to the center of mass (for example from the AdjustableCoMShifter of RealismOverhaul) of parts are not applied in flight.

Craft that use the AdjustableCoMShifter and appear to have 0 Nm of torque from thrust start rotating in flight, as if the CoM was not shifted.

eggrobin:
Fix #2560.


ksp_plugin/interface.cpp
```diff
@@ -811,6 +811,7 @@ void __cdecl principia__InsertOrKeepLoadedPart(
     PartId const part_id,
     char const* const name,
     double const mass_in_tonnes,
+    XYZ centre_of_mass,
```
pleroy, line 814: `const`

ksp_plugin_adapter/main_window.cs
```diff
@@ -239,6 +239,18 @@ internal class MainWindow : VesselSupervisedWindowRenderer {
 
   private void RenderKSPFeatures() {
     if (show_2519_debugging_ui) {
+      string offsets = "";
```
pleroy, line 242: Could we revert changes to this file?

ksp_plugin/part.hpp
```diff
@@ -170,10 +173,14 @@ class Part final {
        RigidMotion<RigidPart, Barycentric> rigid_motion,
        std::function<void()> deletion_callback);
 
+  static RigidMotion<RigidPart, EccentricPart> ToEccentric(
```
pleroy, line 176: 1. Please comment this function.
1. I find the name confusing.  In particular at the call site I read `ToEccentric(centre_of_mass)` as transforming the centre of mass into the eccentric frame.  Since it's a factory, make it clear in the name; since it's not called often, invest in a long name.  `MakeRigidToEccentricMotion` maybe?

ksp_plugin/part.cpp
```diff
@@ -89,6 +91,14 @@ Mass const& Part::mass() const {
   return mass_;
 }
 
+void Part::set_centre_of_mass(Position<EccentricPart> const& centre_of_mass) {
+  centre_of_mass_ = centre_of_mass;
+}
+
+RigidMotion<RigidPart, EccentricPart> Part::ToEccentric() const {
```
pleroy, line 98: I don't find the name very clear.  Maybe `rigid_to_eccentric_motion` would read better at the call site?

Also, I would invest in a member variable (and serialize it).

ksp_plugin_adapter/main_window.cs
eggrobin: Reverted `show_2519_debugging_ui`.

ksp_plugin/part.cpp
eggrobin: As discussed, `MakeRigidToEccentricMotion`, and no.

PR #2609, opened 2020-06-21
Flight plan rebasing

eggrobin:
The `Rebase` button discards past (and ongoing) manœuvres, and sets the flight plan initial state to the current state (in terms of mass as well as position and velocity).

---

Drive-by change: drop the guidance node for one frame when it switches manœuvres; this fixes MuMech/MechJeb2#1276 (though @lamont-granquist wants to leave that one open until the UX is cleaned up a little) and the stock equivalent, whereby guiding to the manœuvre through scheduled burnout would lead to a reorientation to the next manœuvre with the engines still firing.

After this change, at the scheduled end of the burn, MechJeb ceases burn execution (shutting down the engines), and the stock SAS switches to holding attitude.

### DIFF
--- a/ksp_plugin_adapter/flight_planner.cs
+++ b/ksp_plugin_adapter/flight_planner.cs
@@ -139,14 +139,6 @@ class FlightPlanner : VesselSupervisedWindowRenderer {
 
   private void RenderFlightPlan(string vessel_guid) {
     using (new UnityEngine.GUILayout.VerticalScope()) {
-      if (UnityEngine.GUILayout.Button("Rebase")) {
-        var status = plugin.FlightPlanRebase(
-            vessel_guid, predicted_vessel.GetTotalMass());
-        UpdateStatus(status, null);
-        if (status.ok()) {
-          UpdateVesselAndBurnEditors();
-        }
-      }
       if (final_time_.Render(enabled : true)) {
         var status = plugin.FlightPlanSetDesiredFinalTime(vessel_guid,
                                                           final_time_.value);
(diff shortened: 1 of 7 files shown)


ksp_plugin_adapter/flight_planner.cs
```diff
@@ -139,6 +139,14 @@ class FlightPlanner : VesselSupervisedWindowRenderer {
 
   private void RenderFlightPlan(string vessel_guid) {
     using (new UnityEngine.GUILayout.VerticalScope()) {
+      if (UnityEngine.GUILayout.Button("Rebase")) {
+        var status = plugin.FlightPlanRebase(
+            vessel_guid, predicted_vessel.GetTotalMass());
+        UpdateStatus(status, null);
+        if (status.ok()) {
+          UpdateVesselAndBurnEditors();
```
pleroy, line 147: This seems wrong.  You cannot call this function during `Repaint`.  Look at the existing call, it happens at `Layout` time.

I suspect that you can just drop lines 146-148 and immediately return from this function.  The right thing should happen at the next cycle.  Of course, this needs to be tested in game.

ksp_plugin_adapter/flight_planner.cs
```diff
@@ -139,6 +139,14 @@ class FlightPlanner : VesselSupervisedWindowRenderer {
 
   private void RenderFlightPlan(string vessel_guid) {
     using (new UnityEngine.GUILayout.VerticalScope()) {
+      if (UnityEngine.GUILayout.Button("Rebase")) {
```
pleroy, line 142: It seems odd to put this button at the very top of the window.  It doesn't affect the parameters that are immediately below (steps, tolerance).  I would put it closer to the manœuvres that it affects, i.e., near line 232.

PR #2620, opened 2020-06-30
Polynomial primitive and a few templates

pleroy:
Contributes to #2400.


numerics/polynomial.hpp
```diff
@@ -82,6 +83,11 @@ class PolynomialInMonomialBasis : public Polynomial<Value, Argument> {
       Derivative<Value, Argument, order>, Argument, degree_ - order, Evaluator>
   Derivative() const;
 
+  // The integration constant is zero.
```
eggrobin, line 86:
```C++
// The constant term of the result is zero.
```

PR #2631, opened 2020-07-06
ISO 18431-2:2004/Cor.1:2008(en)

pleroy:


eggrobin:
I don’t think we need the (en) to refer to the corrigendum.

PR #2635, opened 2020-07-09
Move a few bit-related operations to a common place

pleroy:
Also use names that reflect what they do...


base/bits.hpp
```diff
@@ -0,0 +1,15 @@
+#pragma once
+
+namespace principia {
+namespace base {
+
+// Floor log2 of n, or 0 for n = 0.  8 -> 3, 7 -> 2.
```
eggrobin, line 6: ↦

base/bits.hpp
```diff
@@ -0,0 +1,15 @@
...
+constexpr int FloorLog2(int n);
+
+// Greatest power of 2 less than or equal to n.  8 -> 8, 7 -> 4.
```
eggrobin, line 9: Same.

pleroy:
retest this please

PR #2637, opened 2020-07-11
A rough implementation of FFT

pleroy:
Will be useful for #2400.


numerics/fast_fourier_transform.hpp
```diff
@@ -0,0 +1,47 @@
...
+
+namespace principia {
+namespace numerics {
+namespace internal_fast_fourier_transform {
+
+using base::FloorLog2;
+
+// This class computes Fourier[{...}, FourierParameters -> {1, -1}] in
+// Mathematica notation.  (The "signal processing" Fourier transform.)
+template<typename Container, int size_>
```
eggrobin, line 19: We do not use Container outside the constructors: templatize on `Scalar` instead, templatize the constructors.

numerics/fast_fourier_transform.hpp
```diff
@@ -0,0 +1,47 @@
...
+class FastFourierTransform {
+ public:
+  // The size must be a power of 2.
+  static constexpr int size = size_;
+  static constexpr int log2_size = FloorLog2(size);
+  static_assert(size == 1 << log2_size);
+
+  using Scalar = typename Container::value_type;
+
+  explicit FastFourierTransform(Container const& container);
```
eggrobin, line 29: Document that `container.size()` must be equal to `size`.

Add an overload for `std::array`, so that we can write `FastFourierTransform(some_array)` using CTAD without having to explicitly specify the known size.

numerics/fast_fourier_transform.hpp
eggrobin: Template parameter sizes that interact with containers have to be `std::size_t`, otherwise deductions break down.

numerics/fast_fourier_transform.hpp
eggrobin: Godbolt outline: https://gcc.godbolt.org/z/9vsoaY.

numerics/fast_fourier_transform_body.hpp
```diff
@@ -0,0 +1,152 @@
...
+using quantities::Sin;
+using quantities::si::Radian;
+using quantities::si::Unit;
+namespace si = quantities::si;
+
+// This implementation follows Myrnyy, A Simple and Efficient FFT Implementation
+// in C++, Dr. Dobbs, 2007.
+
+template<int array_size_, int chunk_size_ = array_size_>
+class DanielsonLánczos {
```
eggrobin, line 26: This class could use a comment (and maybe a citation of *Some improvements in practical Fourier analysis and their application to X-ray scattering from liquids*).

Shouldn’t we have the names of Cooley and Tukey somewhere as well?

pleroy: Added stuff to the bibliography.  Cooley-Tukey (🆒🦃) is for any length (with small prime factors).  Danielson-Lánczos is for powers of 2.

PR #2638, opened 2020-07-12
Mention proper library versions in the loader error message

pleroy:


ksp_plugin_adapter/loader.cs
```diff
@@ -36,8 +36,8 @@ internal static class Loader {
             @"GameData/Principia/Linux64/principia.so",
             @"GameData/Principia/MacOS64/principia.so"};
         is_cxx_installed = null;
-        required_cxx_packages = "libc++ and libc++abi 6.0-2 (Linux) or " +
-                                "El Capitan or later (MacOS)";
+        required_cxx_packages = "libc++abi1-8 and libc++1-8 or later (Linux) " +
+                                "or El Capitan or later (MacOS)";
```
eggrobin, line 40: Principia no longer supports macOS El Capitan since [Descartes](https://github.com/mockingbirdnest/Principia/wiki/Change-Log#descartes).
This should say Sierra.

Issue #2607, opened 2020-06-20
Parachuting Kerbals behave unphysically in Fuchs


eggrobin:
Noted while testing #2604; independently reported by @darthpointer.

@darthpointer further reports that the Kerbals are fine in Frobenius, so this likely has something to do with the rigidity assumptions of Fuchs, rather than improper force or torque application.

If all else fails we could make them unmanageable (parachuting leads to unmanageability by grounding in the short term anyway).

DarthPointer:
I found it affecting not only the kerbal, but the vessel too if I get blown away from the ladder. Deploying the chute after being blown away also causes insane accelerations (2k gees recorded).

eggrobin:
It looks like we are seeing spurious collisions between the kerbal and various parts of the aircraft, which means we treat the kerbal+aircraft system as a (nearly rigid) pile up even as they drift apart, which leads to obvious brokenness.

Let’s see whether we can characterize the spurious collisions…

eggrobin:
Ignoring the spurious collisions fixes ejection, but the EVA parachute itself still has issues: the Kerbal orients itself in a parachute-down attitude, which looks silly and is generally not survivable.

PR #2639, opened 2020-07-12
Handle collisions ourselves

eggrobin:
It appears that `currentCollisions` is often wrong (reference counting never works).

This improves matters with respect to the ejection part of #2607: the elimination of spurious collisions means that the Kerbal is no longer considered part of its aircraft, and thus we no longer impart erratic velocity and position corrections after ejection.

The issue of the attitude under parachute remains, as we do not understand how the parachute acts upon the Kerbal’s attitude.

### DIFF
--- a/ksp_plugin_adapter/collision_reporter.cs
+++ b/ksp_plugin_adapter/collision_reporter.cs
@@ -1,0 +1,24 @@
+﻿using System.Collections.Generic;
+
+namespace principia {
+namespace ksp_plugin_adapter {
+
+internal class CollisionReporter : UnityEngine.MonoBehaviour {
+  private void FixedUpdate() {
+    collisions.Clear();
+  }
+
+  private void OnCollisionEnter(UnityEngine.Collision collision) {
+    collisions.Add(collision);
+  }
+
+  private void OnCollisionStay(UnityEngine.Collision collision) {
+    collisions.Add(collision);
+  }
+
+  public List<UnityEngine.Collision> collisions =
+      new List<UnityEngine.Collision>();
+}
+
+}  // namespace ksp_plugin_adapter
+}  // namespace principia
--- a/ksp_plugin_adapter/ksp_plugin_adapter.cs
+++ b/ksp_plugin_adapter/ksp_plugin_adapter.cs
@@ -1189,7 +1189,6 @@ public partial class PrincipiaPluginAdapter
             plugin_.ReportGroundCollision(
                 closest_physical_parent(part1).flightID);
           }
-#if KSP_VERSION_1_9_1
           var collision_reporter =
               part1.gameObject.GetComponent<CollisionReporter>();
           if (part1.gameObject.GetComponent<CollisionReporter>() == null) {
@@ -1200,9 +1199,6 @@ public partial class PrincipiaPluginAdapter
           }
           foreach (var collision in collision_reporter.collisions) {
             var collider = collision.collider;
-#elif KSP_VERSION_1_7_3
-          foreach (var collider in part1.currentCollisions) {
-#endif
             if (collider == null) {
               // This happens, albeit quite rarely, see #1447.  When it happens,
               // the null collider remains in |currentCollisions| until the next
--- a/ksp_plugin_adapter/ksp_plugin_adapter.cs
+++ b/ksp_plugin_adapter/ksp_plugin_adapter.cs
@@ -1155,11 +1155,11 @@ public partial class PrincipiaPluginAdapter
 
     plugin_.PrepareToReportCollisions();
 
-    // The collisions are reported and stored into |currentCollisions| in
-    // OnCollisionEnter|Stay|Exit, which occurred while we yielded.
-    // Here, the |currentCollisions| are the collisions that occurred in the
-    // physics simulation, which is why we report them before calling
-    // |AdvanceTime|.
+    // The collisions are reported by the
+    // CollisionReporter.OnCollisionEnter|Stay events, which occurred while we
+    // yielded.
+    // Here, the |CollisionReporter.collisions| are the collisions that occurred
+    // in the physics simulation.
     foreach (Vessel vessel1 in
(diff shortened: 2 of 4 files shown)


ksp_plugin_adapter/ksp_plugin_adapter.cs
```diff
@@ -1155,18 +1155,19 @@ class PartCentredForceHolder {
...
+    // Here, the |CollisionReporter.collisions| are the collisions that occurred
+    // in the physics simulation.
     foreach (Vessel vessel1 in
              FlightGlobals.Vessels.Where(v => !v.packed && is_manageable(v))) {
       if (plugin_.HasVessel(vessel1.id.ToString())) {
         if (vessel1.isEVA && (vessel1.evaController.OnALadder ||
                               is_clambering(vessel1.evaController))) {
           var vessel2 = vessel1.evaController.LadderPart?.vessel;
           if (vessel2 != null && !vessel2.packed && is_manageable(vessel2)) {
+            Log.Info("Reporting climbing a ladder");
```
pleroy, line 1170: Remove?

ksp_plugin_adapter/ksp_plugin_adapter.cs
```diff
@@ -1188,11 +1189,16 @@ class PartCentredForceHolder {
             plugin_.ReportGroundCollision(
                 closest_physical_parent(part1).flightID);
           }
-#if KSP_VERSION_1_9_1
-          foreach (var collider in part1.currentCollisions.Keys) {
-#elif KSP_VERSION_1_7_3
-          foreach (var collider in part1.currentCollisions) {
-#endif
+          var collision_reporter =
+              part1.gameObject.GetComponent<CollisionReporter>();
```
pleroy, line 1193: Is this going to be expensive?  I think we had performance problems with this kind of operation in the past.

ksp_plugin_adapter/collision_reporter.cs
```diff
@@ -0,0 +1,24 @@
...
+
+namespace principia {
+namespace ksp_plugin_adapter {
+
+internal class CollisionReporter : UnityEngine.MonoBehaviour {
+  private void FixedUpdate() {
+    collisions.Clear();
+  }
+
+  private void OnCollisionEnter(UnityEngine.Collision collision) {
```
pleroy, line 11: I would vote for skipping the `Add` if `collision.collider` happens to be `null`.  That would simplify the client code somewhat.

ksp_plugin_adapter/ksp_plugin_adapter.cs
```diff
@@ -1188,11 +1189,16 @@ class PartCentredForceHolder {
...
+          if (part1.gameObject.GetComponent<CollisionReporter>() == null) {
+            // This would only happen if |part1| had been added after
+            // |BetterLateThanNever|, but we never know what the game will throw
+            // at us.
+            continue;
+          }
+          foreach (var collision in collision_reporter.collisions) {
+            var collider = collision.collider;
             if (collider == null) {
```
pleroy, line 1202: See my comment above and remove this statement.  The comment about #1447 can move to the `CollisionReporter` class.

ksp_plugin_adapter/ksp_plugin_adapter.cs
```diff
@@ -1229,6 +1237,8 @@ class PartCentredForceHolder {
                 // better ignore the collision, the Kerbal will soon become
                 // ready anyway.
               } else if (is_manageable(vessel2)) {
+                Log.Info($@"Reporting collision with collider {collider.name} ({
+                            (UnityLayers)collider.gameObject.layer})");
```
pleroy, line 1241: Remove?

ksp_plugin_adapter/main_window.cs
```diff
@@ -521,7 +521,7 @@ internal class MainWindow : VesselSupervisedWindowRenderer {
   // They must not be serialized: their non-default values can lead to absurd
   // behaviour.
   private static bool conserve_angular_momentum = true;
-  private static readonly bool show_2519_debugging_ui = false;
+  private static readonly bool show_2519_debugging_ui = true;
```
pleroy, line 524: Set back to `false`.

ksp_plugin_adapter/ksp_plugin_adapter.cs
eggrobin: We are getting a component of this specific game object, and it should almost always be there.
We do kind of thing every frame for the navball already:
https://github.com/mockingbirdnest/Principia/blob/aef945b1f8c1def933d3d8f74494abfefa6e1085/ksp_plugin_adapter/ksp_plugin_adapter.cs#L1725-L1731 https://github.com/mockingbirdnest/Principia/blob/aef945b1f8c1def933d3d8f74494abfefa6e1085/ksp_plugin_adapter/ksp_plugin_adapter.cs#L1836-L1847.


I think you are thinking of incidents involving [`FindObjectOfType`](https://github.com/mockingbirdnest/Principia/pull/1489/files), which looks at all objects; that is slow.

ksp_plugin_adapter/collision_reporter.cs
eggrobin: It is not immediately obvious that this would be equivalent: what if the collider dies during the physics step?
Note that Unity objects have custom comparison operators, so `!= null` checks not just that the `object` is not null (which would remain true as long as we keep a reference to it), but also that the `UnityEngine.Object` has not been destroyed.

ksp_plugin_adapter/ksp_plugin_adapter.cs
eggrobin: No, the provenance of the collision is useful information. The actual part-to-part collison report also logs on the C++ side (with IDs and part names, but without the added context), so it’s not like this makes logspam much worse.

ksp_plugin_adapter/ksp_plugin_adapter.cs
eggrobin: Same as the ladder, this is useful.

PR #2642, opened 2020-07-12
Log version and module information to ERROR

pleroy:
Some users (I'm looking at you, @lpgagnon) have been known to give us the `ERROR` file and it's a bit embarrassing that we cannot do anything with it because we don't have enough information there to decode the stack.

### DIFF
--- a/ksp_plugin/interface.cpp
+++ b/ksp_plugin/interface.cpp
@@ -653,13 +653,13 @@ void __cdecl principia__InitGoogleLogging() {
                               << message;
         });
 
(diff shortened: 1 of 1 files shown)


eggrobin:
This means that we would systematically create the ERROR and WARNING files even if we never log at those severities, which makes me sad.

On the other hand we have quite a bit of log spam at ERROR at the moment because of the `Reached maximum step count` business, the existence of the files is not a very useful signal; we can revisit this (by adding this information in the log file header on the glog side, instead of adding it as log lines) once we have dealt with that log spam.

PR #2644, opened 2020-07-12
Parachute torque

### LINKED ISSUE #2607: Parachuting Kerbals behave unphysically in Fuchs

eggrobin:
As @ferram4 said,
> consider it all cursed.

This fixes #2607.

### DIFF
--- a/ksp_plugin/interface.cpp
+++ b/ksp_plugin/interface.cpp
@@ -334,6 +334,48 @@ void __cdecl principia__ActivateRecorder(bool const activate) {
   }
 }
 
+XYZ __cdecl principia__AngularMomentumFromAngularVelocity(
+    XYZ world_angular_velocity,
+    XYZ moments_of_inertia_in_tonnes,
+    WXYZ principal_axes_rotation,
+    WXYZ part_rotation) {
+  journal::Method<journal::AngularMomentumFromAngularVelocity> m(
+      {world_angular_velocity,
+       moments_of_inertia_in_tonnes,
+       principal_axes_rotation,
+       part_rotation});
+  using PartPrincipalAxes = Frame<serialization::Frame::PhysicsTag,
+                                  Arbitrary,
+                                  Handedness::Left,
+                                  serialization::Frame::PRINCIPAL_AXES>;
+
+  auto const angular_velocity =
+      FromXYZ<AngularVelocity<World>>(world_angular_velocity);
+
+
+  static constexpr MomentOfInertia zero;
+  auto const moments_of_inertia =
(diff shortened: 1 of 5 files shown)


eggrobin:
The calculations could probably gain from being done on the C++ side, but they involve only C#-side inputs, so this would have to be a free interface function that takes Δω_world, the two quaternions, and the principal moments of inertia, types all of that, multiplies the inertia tensor by Δω, and untypes and returns ΔL.

ksp_plugin_adapter/ksp_plugin_adapter.cs
```diff
@@ -1494,8 +1513,40 @@ class PartCentredForceHolder {
...
+            // Instead of applying forces some distance from the centre of mass,
+            // or applying torques, EVA parachutes directly set the angular
+            // velocity of the Kerbal.  We register a torque equivalent to the
+            // change in angular momentum; failing to do so would cause the
+            // parachute to act on the linear motion but not the attitude,
+            // so that Kerbals would keep spinning under their parachute in
+            // stock, and would orient themselves head-down (and parachute-down)
+            // with FAR.  See #2607.
+            // To quote ferram4, “consider it all cursed”.
+            Vector3d Δω_world = part.rb.angularVelocity - old_angular_velocity;
```
pleroy, line 1530: I think I would really prefer if this code was in C++.  It's a bit annoying to pass things through the interface and retype all the data, but we would have a better confidence that everything is correct, including when we maintain it.  It could live directly in `interface.cpp`, no point in pushing it to some class.

eggrobin: Done. Considering how sloppy our handling of the MTS system was, I am not sure this change is less error-prone. I tried cleaning up the existing conversion constant.

ksp_plugin/interface.cpp
```diff
@@ -334,6 +334,48 @@ void __cdecl principia__ActivateRecorder(bool const activate) {
...
+       principal_axes_rotation,
+       part_rotation});
+  using PartPrincipalAxes = Frame<serialization::Frame::PhysicsTag,
+                                  Arbitrary,
+                                  Handedness::Left,
+                                  serialization::Frame::PRINCIPAL_AXES>;
+
+  auto const angular_velocity =
+      FromXYZ<AngularVelocity<World>>(world_angular_velocity);
+
```
pleroy, line 354: Remove one empty line.

PR #2651, opened 2020-07-17
Change ComputeRCSCharacteristics to use the sum of positive thrusts, …

### LINKED ISSUE #2571: RCS burns in editor are incorrect for multi-port thrusters
In https://github.com/mockingbirdnest/Principia/blob/05281ea21caf746a8e23cab5af667b1d1cdcbb2f/ksp_plugin_adapter/burn_editor.cs#L239
Thrusts are averaged across all thrust transforms.
For an _N_ port thruster this will tend to give a resulting thrust of _maxthrust_/_N_ assuming all ports are orthogonal, or some intermediate value otherwise.
Thrusters can actually deliver _maxthrust_ down any single port so the statement should return the sum of positive thrusts rather than the average.

RCrockford:
…rather than the average since multi-port thrusters appear to be able to deliver full thrust down any single port.

This fixes #2571.

### DIFF
--- a/ksp_plugin_adapter/burn_editor.cs
+++ b/ksp_plugin_adapter/burn_editor.cs
@@ -236,7 +236,7 @@ class BurnEditor : ScalingRenderer {
              (from transform in engine.thrusterTransforms
               select Math.Max(0,
                               Vector3d.Dot(reference_direction,
-                                           -transform.up))).Average()).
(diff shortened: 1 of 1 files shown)


pleroy:

eggrobin:
ok to test

PR #2652, opened 2020-07-17
Golden section search

pleroy:
Will be needed for #2400.

### DIFF
--- a/numerics/root_finders.hpp
+++ b/numerics/root_finders.hpp
@@ -19,6 +19,7 @@ using quantities::Derivative;
 // function agreeing with |f| on the values of |Argument|.
 // If |f(lower_bound)| and |f(upper_bound)| are both nonzero, they must be of
 // opposite signs.
+// TODO(phl): Use Brent's algorithm.
 template<typename Argument, typename Function>
 Argument Bisect(Function f,
                 Argument const& lower_bound,
@@ -26,6 +27,7 @@ Argument Bisect(Function f,
 
 // Performs a golden-section search to find the minimum of |f| between
 // |lower_bound| and |upper_bound|.
+// TODO(phl): Use Brent's algorithm.
 template<typename Argument,
          typename Function,
          typename Compare = std::less<
--- a/numerics/root_finders.hpp
+++ b/numerics/root_finders.hpp
@@ -23,6 +23,12 @@ Argument Bisect(Function f,
                 Argument const& lower_bound,
                 Argument const& upper_bound);
 
+//TODO(phl):comment
+template<typename Argument, typename Compare, typename Function>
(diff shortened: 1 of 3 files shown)


numerics/root_finders.hpp
```diff
@@ -18,11 +19,24 @@ using quantities::Derivative;
 // function agreeing with |f| on the values of |Argument|.
 // If |f(lower_bound)| and |f(upper_bound)| are both nonzero, they must be of
 // opposite signs.
+// TODO(phl): Use Brent's algorithm.
 template<typename Argument, typename Function>
 Argument Bisect(Function f,
                 Argument const& lower_bound,
                 Argument const& upper_bound);
 
+// Performs a golden-section search to find the minimum of |f| between
```
eggrobin, line 28: *A* minimum.

numerics/root_finders_body.hpp
```diff
@@ -56,6 +56,58 @@ Argument Bisect(Function f,
...
+      f_lower = f_lower_interior;
+      lower_interior = upper_interior;
+      f_lower_interior = f_upper_interior;
+      upper_interior = Barycentre<Argument, double>(
+          {lower, upper}, {lower_interior_ratio, upper_interior_ratio});
+      f_upper_interior = f(upper_interior);
+    }
+  }
+
+  return Barycentre<Argument, double>({lower, upper}, {1, 1});
```
eggrobin, line 108: This feels a little bit cheesy, but OK.

Can you compute an upper bound on the distance (in ULPs) between `lower` and `upper`, and document the error of the result accordingly in the header?

numerics/root_finders_test.cpp
```diff
@@ -53,6 +56,37 @@ TEST_F(RootFindersTest, SquareRoots) {
   }
 }
 
+TEST_F(RootFindersTest, GoldenSectionSearch) {
```
eggrobin, line 59: Add a test with a function that has multiple extrema on the search interval.

numerics/root_finders_body.hpp
pleroy: I could probably prove that `lower` and `upper` are a few ULPs apart when the loop exits.

But that is irrelevant, see my comment in the test.  Near its extremum the function is *constant* over tens of millions of ULPs.  So we don't really care how close `lower` and `upper` are to each other.  We would want to know how close they are to the extremum, but that's mostly unknowable.

The real solution here is to use Brent's method, which approximates the function by a parabola, and therefore gets a much better determination of its extremum.

numerics/root_finders_test.cpp
pleroy: Done.

PR #2662, opened 2020-08-02
Fix the collision warning for bodies with oceans

eggrobin:
Due to confusion between radii and altitudes, submarine periapsides were seen as a mere collision *risk*.


eggrobin:
retest this please

PR #2664, opened 2020-08-04
Collapsible manœuvres

eggrobin:
Once we add orbit analysis for coasts, the one-line coast descriptions will reflect that analysis (something like `Coast for 2 d 14 h in Earth orbit (5.2 rev.)`) and will be expandable to show:
- the number of sidereal revolutions, with ±1 buttons;
- the number of nodal revolutions, with ±1 buttons;
- the number of anomalistic revolutions, with ±1 buttons;
- in the distant future, the number of parent-body-synodic revolutions, with ±1 buttons;
- a button to open the full orbit analysis (at least for the final trajectory).

All manœuvres minimized:
screenshot19
One expanded:
screenshot20

### DIFF
--- a/ksp_plugin_adapter/burn_editor.cs
+++ b/ksp_plugin_adapter/burn_editor.cs
@@ -56,28 +56,44 @@ class BurnEditor : ScalingRenderer {
     ComputeEngineCharacteristics();
   }
 
+  public enum Event {
+    None,
+    Changed,
+    Deleted,
+    Minimized,
+    Maximized,
+  }
+
   // Renders the |BurnEditor|.  Returns true if and only if the settings were
   // changed.
-  public bool Render(string header,
+  public Event Render(string header,
                      bool anomalous,
-                     double burn_final_time,
-                     Action delete) {
+                     double burn_final_time) {
     bool changed = false;
     previous_coast_duration_.max_value = burn_final_time - time_base;
     using (new UnityEngine.GUILayout.HorizontalScope()) {
-      UnityEngine.GUILayout.Label(header);
-      string frame_info = "";
-      if (!reference_frame_selector_.FrameParameters().Equals(
+      if (UnityEngine.GUILayout.Button(
+              minimized ? "+" : "-", GUILayoutWidth(1))) {
+        minimized = !minimized;
+        return minimized ? Event.Minimized : Event.Maximized;
+      }
+      UnityEngine.GUILayout.Label(minimized ? $"{header} ({Δv():0.000} m/s)"
+                                            : header);
+      string info = "";
+      if (!minimized &&
+          !reference_frame_selector_.FrameParameters().Equals(
               adapter_.plotting_frame_selector_.FrameParameters())) {
-        frame_info = "Manœuvre frame differs from plotting frame";
+        info = "Manœuvre frame differs from plotting frame";
       }
       UnityEngine.GUILayout.Label(
-          frame_info,
+          info,
           Style.RightAligned(Style.Info(UnityEngine.GUI.skin.label)));
       if (UnityEngine.GUILayout.Button("Delete", GUILayoutWidth(2))) {
-        delete();
-        changed = true;
+        return Event.Deleted;
       }
+    }
+    if (minimized) {
+      return Event.None;
     }
     using (new UnityEngine.GUILayout.VerticalScope()) {
       // When we are first rendered, the |initial_mass_in_tonnes_| will just have
@@ -152,7 +168,7 @@ class BurnEditor : ScalingRenderer {
                                   Style.Warning(UnityEngine.GUI.skin.label));
       changed_reference_frame_ = false;
     }
-    return changed;
+    return changed ? Event.Changed : Event.None;
   }
 
   public double Δv() {
@@ -297,10 +313,12 @@ class BurnEditor : ScalingRenderer {
                               plugin.FlightPlanGetInitialTime(
                                   vessel_.id.ToString());
 
-  private double final_time => initial_time_ + duration_;
+  public double initial_time => initial_time_;
+  public double final_time => initial_time_ + duration_;
 
   private IntPtr plugin => adapter_.Plugin();
   public int index { private get; set; }
+  public bool minimized { private get; set; } = true;
   private BurnEditor previous_burn => get_burn_at_index_(index - 1);
 
   private bool is_inertially_fixed_;
--- a/ksp_plugin_adapter/burn_editor.cs
+++ b/ksp_plugin_adapter/burn_editor.cs
@@ -67,8 +67,8 @@ class BurnEditor : ScalingRenderer {
   // Renders the |BurnEditor|.  Returns true if and only if the settings were
   // changed.
   public Event Render(string header,
-                     bool anomalous,
-                     double burn_final_time) {
+                      bool anomalous,
+                      double burn_final_time) {
     bool changed = false;
     previous_coast_duration_.max_value = burn_final_time - time_base;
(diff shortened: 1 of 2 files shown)


ksp_plugin_adapter/burn_editor.cs
```diff
@@ -56,29 +56,45 @@ class BurnEditor : ScalingRenderer {
...
+    Changed,
+    Deleted,
+    Minimized,
+    Maximized,
+  }
+
   // Renders the |BurnEditor|.  Returns true if and only if the settings were
   // changed.
-  public bool Render(string header,
+  public Event Render(string header,
```
pleroy, line 69: Bad indent.

ksp_plugin_adapter/burn_editor.cs
```diff
@@ -56,29 +56,45 @@ class BurnEditor : ScalingRenderer {
...
+      UnityEngine.GUILayout.Label(minimized ? $"{header} ({Δv():0.000} m/s)"
+                                            : header);
+      string info = "";
+      if (!minimized &&
+          !reference_frame_selector_.FrameParameters().Equals(
               adapter_.plotting_frame_selector_.FrameParameters())) {
-        frame_info = "Manœuvre frame differs from plotting frame";
+        info = "Manœuvre frame differs from plotting frame";
       }
       UnityEngine.GUILayout.Label(
```
pleroy, line 88: The placement of this string next to the `Delete` button is weird, it seems to invite you to delete the manœuvre when the frames don't match.  I would do something like:
```
Manœuvre #1 <grey>frame differs from plotting frame</grey>
```
That is, left align the message and shorten it a bit.

ksp_plugin_adapter/burn_editor.cs
```diff
@@ -56,29 +56,45 @@ class BurnEditor : ScalingRenderer {
...
-          frame_info,
+          info,
           Style.RightAligned(Style.Info(UnityEngine.GUI.skin.label)));
       if (UnityEngine.GUILayout.Button("Delete", GUILayoutWidth(2))) {
```
pleroy, line 91: It might look nicer if this button was next to `+`/`-` but I can see the point of not having `Eject` and `Format` next to each other.

Could we use `🗑` (U+1F5D1) as a label for this button?  That would be clear enough.

ksp_plugin_adapter/burn_editor.cs
```diff
@@ -56,29 +56,45 @@ class BurnEditor : ScalingRenderer {
     ComputeEngineCharacteristics();
   }
 
+  public enum Event {
+    None,
```
pleroy, line 60: `None` is mysterious.  `Unchanged`?

ksp_plugin_adapter/burn_editor.cs
```diff
@@ -56,29 +56,45 @@ class BurnEditor : ScalingRenderer {
...
+
   // Renders the |BurnEditor|.  Returns true if and only if the settings were
   // changed.
-  public bool Render(string header,
+  public Event Render(string header,
                      bool anomalous,
-                     double burn_final_time,
-                     Action delete) {
+                     double burn_final_time) {
     bool changed = false;
```
pleroy, line 72: The code would be a bit cleaner if you removed this `bool` and had a local variable of type `Event`.

ksp_plugin_adapter/flight_planner.cs
```diff
@@ -1,6 +1,8 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Diagnostics.Tracing;
```
pleroy, line 3: Why?

ksp_plugin_adapter/flight_planner.cs
```diff
@@ -259,38 +261,42 @@ class FlightPlanner : VesselSupervisedWindowRenderer {
...
               UpdateStatus(status, null);
               burn_editors_[i].Close();
               burn_editors_.RemoveAt(i);
               UpdateBurnEditorIndices();
               Shrink();
               return;
-            } else {
+            }
+            case BurnEditor.Event.Minimized:
+            case BurnEditor.Event.Maximized:
```
pleroy, line 284: Put `{}` in all the `case` branches, it's cleaner.

ksp_plugin_adapter/flight_planner.cs
```diff
@@ -347,47 +353,61 @@ class FlightPlanner : VesselSupervisedWindowRenderer {
...
+      double start_of_coast = index == 0
+          ? plugin.FlightPlanGetInitialTime(vessel_guid)
+          : burn_editors_[index - 1].final_time;
+      string coast_description = index == burn_editors_.Count
+          ? "Final trajectory"
+          : $@"Coast for {
+                (burn_editors_[index].initial_time -
+                 start_of_coast).FormatDuration(show_seconds : false)}";
+      UnityEngine.GUILayout.Label(
+          coast_description);
```
pleroy, line 368: Fits on the previous line.

ksp_plugin_adapter/flight_planner.cs
```diff
@@ -347,47 +353,61 @@ class FlightPlanner : VesselSupervisedWindowRenderer {
...
+      if (UnityEngine.GUILayout.Button(
             "Add manœuvre",
```
pleroy, line 370: Same.

ksp_plugin_adapter/flight_planner.cs
```diff
@@ -347,47 +353,61 @@ class FlightPlanner : VesselSupervisedWindowRenderer {
...
-            vessel_guid, index));
-        burn_editors_.Insert(index, editor);
-        UpdateBurnEditorIndices();
+            GUILayoutWidth(4))) {
+        double initial_time;
+        if (index == 0) {
+          initial_time = plugin.CurrentTime() + 60;
+        } else {
+          initial_time = plugin.FlightPlanGetManoeuvre(
+                              vessel_guid,
```
pleroy, line 377: Weird indent.

ksp_plugin_adapter/flight_planner.cs
```diff
@@ -347,47 +353,61 @@ class FlightPlanner : VesselSupervisedWindowRenderer {
...
+                              index - 1).final_time + 60;
+        }
+        var editor = new BurnEditor(
+            adapter_,
+            predicted_vessel,
+            initial_time,
+            index,
+            get_burn_at_index : burn_editors_.ElementAtOrDefault);
+        editor.minimized = false;
```
pleroy, line 386: Could we just say that a burn editor is created maximized?  Shooting at the member variable is ugly.

ksp_plugin_adapter/burn_editor.cs
eggrobin: The font doesn’t support that.

ksp_plugin_adapter/burn_editor.cs
eggrobin: It’s `Event.None`, which makes some sense; `Event.Unchanged` is weird.

ksp_plugin_adapter/burn_editor.cs
eggrobin: We need a `bool` for the `|=`. We should probably refactor this quite a bit, in particular so that it doesn’t return `changed` on the first repaint, but not today.

ksp_plugin_adapter/flight_planner.cs
eggrobin: It is added maximized, but created minimized when loading an existing flight plan.

This is a property which is publicly write-only, not a member variable.

PR #2669, opened 2020-08-09
A first implementation of Кудрявцев projection

pleroy:
This is far from perfect (see the `TODO`s) but at least the basic algorithm and all the mechanics to generate Poisson series is in place and passes a test.

A significant contribution to #2400.


numerics/frequency_analysis_body.hpp
```diff
@@ -46,7 +149,100 @@ AngularFrequency PreciseMode(
...
+    // Only indices 0 to m are used in this array.  It contains Qₘⱼ.
+    std::array<Square<Value>, basis_size> Q;
+    for (int j = 0; j <= m; ++j) {
+      Q[j] = dot(basis[m], basis[j], weight);
+    }
+
+    // Only indices 0 to m - 1 are used in this array.  It contains Bⱼ⁽ᵐ⁾.
+    std::array<Value, basis_size> B;
+    for (int j = 0; j < m; ++j) {
+      Value accumulator;
```
eggrobin, line 203: Could we call this `Σ_αⱼₛ_Qₘₛ` (with `+=` and a `-` below)?
We do that in many places (`Σ_Bₗ_m_yˡ` in the elliptic integrals, `Σj_aʹ_ij_g_jk` in an integrator (we should switch to subscripts there someday), etc.), and `accumulator` is not very informative.

Same for the other accumulators.

This would make it much easier to check the code against the paper.

pleroy: Will do.  (Tofu, tofu, tofu, tofus tofui tofu, tofua tofua tofua, tofuum tofibus tofibus.)

PR #2679, opened 2020-08-18
An attempt at addressing ill-conditioned basis elements.

pleroy:
Ignore a basis element if it is sufficiently close to the span of the preceding basis elements (cancellation of at least 24 bits).  Not only does this yield better accuracy, it doesn't cause such huge tolerance changes between casanova and carpaccio (compare a5d483f445fe5e8eaf9efba1b57e28f1adfb973c and 34e915f61fb010aaa9de1004fae59084d008629c)

Contributes to #2400.

### DIFF
--- a/numerics/frequency_analysis_body.hpp
+++ b/numerics/frequency_analysis_body.hpp
@@ -258,7 +258,7 @@ IncrementalProjection(Function const& function,
           // here because our function is real and bounded.  But even if the
           // norm could be computed but was very small, we would end up with an
           // ill-conditioned solution.  Geometrically, we are in a situation
-          // where eₘ/ is very close to the space spanned by the (bₛ), that is,
+          // where eₘ is very close to the space spanned by the (bₛ), that is,
           // by the (eₛ) for i < m.  The fact that the basis elements and no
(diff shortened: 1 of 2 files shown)


numerics/frequency_analysis_body.hpp
```diff
@@ -249,17 +249,20 @@ IncrementalProjection(Function const& function,
...
-          // (eₛ) for i < m.  The fact that the basis elements and no longer
-          // independent when the degree increases is duely noted by [CV84].
-          // Given that eₘ effectively doesn't have benefit for the projection,
-          // we just drop it and continue with the algorithm.
+          // value of the square of that norm ends up negative, zero, or very
+          // small.  It makes no sense to have complex numbers (or infinities)
+          // here because our function is real and bounded.  But even if the
+          // norm could be computed but was very small, we would end up with an
+          // ill-conditioned solution.  Geometrically, we are in a situation
+          // where eₘ/ is very close to the space spanned by the (bₛ), that is,
```
eggrobin, line 261: stray /.

eggrobin (approved):

PR #2720, opened 2020-09-16
Tables for Gauss-Legendre quadrature

eggrobin:


pleroy:
Could we avoid the redundancy?  The algorithm could easily symmetricize the weights/nodes and that would make the tables twice as small.

pleroy:
#2400.

PR #2733, opened 2020-09-25
Much ado about nothing

eggrobin:
- Fix a bug in long-running golden section searches (typically those that find an extremum close to 0).
- Fix a bug whereby Brent would not terminate when finding a minimum close to 0 (which would be embarrassing for #2400).
- Test golden section search with a non-arithmetic comparator for which +0 is the minimum of [-1, 1].
- Add a default `eps` of √ϵ for Brent.


documentation/bibliography.bib
```diff
@@ -670,6 +670,18 @@ @article{OkunborSkeel1994
...
 
+@Article{Overholt1965,
+  author       = {Overholt, K. J.},
+  date         = {1965},
+  doi          = {10.1007/BF01937508},
+  journaltitle = {BIT},
+  number       = {4},
+  pages        = {284--286},
+  title        = {An Instability in the {F}ibonacci and Golden Section Search Methods},
+  note         = {Merged with the preceding article (starting p.~282) on the journal website},
```
pleroy, line 681: `note` before `number`.

PR #2734, opened 2020-09-25
Reduce heap usage in Poisson series

pleroy:
Use `vector` instead of `map`, avoid copies, do more `reserve`.  Overall this saves about 30% of the CPU.

Optimize #2400.


eggrobin (approved):

pleroy:
retest this please

PR #2745, opened 2020-10-03
Minor fixes for local macOS build

rnlahaye:
This PR contains fixes for some issues I ran into while building Principia locally.

Specifically:
* `principia_make.sh` has been marked as executable.
* `principia_make.sh` now defines AGENT_OS if it is not present.
* I tried building using `clang` 9, 10, and 6 and encountered build errors. Eventually I realized I needed `clang` 8. The macOS build documentation has been updated to make this more clear.
* Xamarin Studio for macOS is no longer a thing (it has been replaced with Visual Studio for Mac). I installed `msbuild` directly and was able to build the adapter using it. `msbuild` seems to be available for Linux as well, so I updated the makefile to use it. Build documentation has been updated accordingly.
* I added a macOS-only script to automatically create aliases to all installed KSP assemblies.

These changes should not affect the Azure pipelines.


principia_make.sh
```diff
@@ -1,3 +1,8 @@
+# Define AGENT_OS if not present.
```
eggrobin, line 1: `principia_make.sh` (which exists in the dependency repositories as well) is not meant to be run directly. It is used by the Azure pipelines that run the continuous integration, and here it mostly contains workarounds to please Azure (the term « agent » and the variable `AGENT_OS` both come from Azure).

FWIW the calling code is as follows.
```bash
if [ -f ./principia_variable_overrides.sh ]; then
  source ./principia_variable_overrides.sh
fi

if [ "${AGENT_OS?}" == "Darwin" ]; then
  C_FLAGS="${PRINCIPIA_C_FLAGS?} -mmacosx-version-min=${PRINCIPIA_MACOS_VERSION_MIN?} -arch x86_64"
  CXX_FLAGS="${PRINCIPIA_CXX_FLAGS?} ${PRINCIPIA_MACOS_CXX_FLAGS?}"
elif [ "${AGENT_OS?}" == "Linux" ]; then
  C_FLAGS="${PRINCIPIA_C_FLAGS?} -m64"
  CXX_FLAGS="${PRINCIPIA_CXX_FLAGS?}"
else
  C_FLAGS="${PRINCIPIA_C_FLAGS?}"
  CXX_FLAGS="${PRINCIPIA_CXX_FLAGS?}"
fi
LD_FLAGS="${C_FLAGS?} ${PRINCIPIA_LD_FLAGS?}"
CXX_FLAGS="${CXX_FLAGS?} ${LD_FLAGS?}"

# Stop on first failure.
set -e
source ./principia_make.sh
```

You probably want to run `make` directly when building Principia yourself.

Please revert the changes to this file.

add_ksp_assemblies_macos.sh
```diff
@@ -0,0 +1,33 @@
...
+  
+  # Get the KSP version.
+  KSP_VERSION=$(mdls -name=kMDItemVersion -raw "${KSP_APP}") 
+  echo "Version: ${KSP_VERSION}"
+
+  # Link the KSP app's dlls to the KSP Assemblies dir.
+  SOURCE_DIR="${KSP_APP}/Contents/Resources/Data/Managed"
+  TARGET_DIR="${KSP_ASSEMBLIES_DIR}/${KSP_VERSION}"
+  ln -s "${SOURCE_DIR}" "${TARGET_DIR}"
+done
```
eggrobin, line 33: New line at end of file.

eggrobin (changes requested):

eggrobin (approved):

PR #2756, opened 2020-10-08
Implement an automatic Clenshaw-Curtis quadrature and use it in the Poisson series inner product and norm

eggrobin:
#2400.


numerics/poisson_series_body.hpp
```diff
@@ -310,8 +310,11 @@ PoissonSeries<Value, degree_, Evaluator>::Norm(
     PoissonSeries<double, wdegree_, Evaluator> const& weight,
     Instant const& t_min,
     Instant const& t_max) const {
-  auto const integrand = PointwiseInnerProduct(*this, *this) * weight;
-  return Sqrt(integrand.Integrate(t_min, t_max) / (t_max - t_min));
+  auto integrand = [this, &weight](Instant const& t) {
+    return Hilbert<Value>::InnerProduct((*this)(t), (*this)(t)) * weight(t);
```
pleroy, line 314: Put a TODO in `hilbert.hpp` to provide a function `Norm²`.

numerics/quadrature.hpp
```diff
@@ -1,32 +1,53 @@
...
+// TODO(egg): The semantics of the |relative_tolerance| are unclear.
+template<int initial_points = 2, typename Argument, typename Function>
+Primitive<std::invoke_result_t<Function, Argument>, Argument>
+AutomaticClenshawCurtis(
+    Function const& f,
+    Argument const& lower_bound,
+    Argument const& upper_bound,
+    double relative_tolerance = 0x1p-16);
+
+// |points| must be of the form 2ᵖ + 1 for some p ∈ ℕ.
```
pleroy, line 34: It would be good if the comments explained the difference between this function and the previous one.

numerics/quadrature_body.hpp
```diff
@@ -24,35 +34,133 @@ Primitive<std::invoke_result_t<Function, Argument>, Argument> Gauss(
...
+  return Gauss<points>(f,
                        lower_bound,
                        upper_bound,
                        LegendreRoots[points],
                        GaussLegendreWeights[points]);
 }
 
+template<int points, typename Argument, typename Function>
+Primitive<std::invoke_result_t<Function, Argument>, Argument>
+AutomaticClenshawCurtis(Function const& f,
```
pleroy, line 58: It would be good to pick a different name, otherwise the `using` makes it visible outside the internal namespace.

numerics/quadrature_body.hpp
```diff
@@ -24,35 +34,133 @@ Primitive<std::invoke_result_t<Function, Argument>, Argument> Gauss(
...
+  constexpr int N = points - 1;
+  constexpr int log2_N = FloorLog2(N);
+  static_assert(N == 1 << log2_N);
+
+  Difference<Argument> const half_width = (upper_bound - lower_bound) / 2;
+
+  constexpr Angle N⁻¹π = π * Radian / N;
+
+  // We use a discrete Fourier transform rather than a cosine transform, see
+  // [Gen72c], (3).
```
pleroy, line 122: "equation (3)".

numerics/quadrature_body.hpp
```diff
@@ -24,35 +34,133 @@ Primitive<std::invoke_result_t<Function, Argument>, Argument> Gauss(
...
+  // TODO(egg): Consider a discrete cosine transform, ideally incrementally
+  // computed to improve the performance of the automatic quadrature.
+
+  // If we identify [lower_bound, upper_bound] with [-1, 1],
+  // f_cos_N⁻¹π[s] is f(cos πs/N).
+  std::vector<Value> f_cos_N⁻¹π;
+  f_cos_N⁻¹π.resize(2 * N);
```
pleroy, line 129: It would be more efficient to do `reserve` and then `push_back` in the loops: no need to zero-initialize the vector.

numerics/quadrature_body.hpp
```diff
@@ -24,35 +34,133 @@ Primitive<std::invoke_result_t<Function, Argument>, Argument> Gauss(
...
+  }
+  for (int s = N + 1; s <= 2 * N - 1; ++s) {
+    f_cos_N⁻¹π[s] = f_cos_N⁻¹π[2 * N - s];  // [Gen72c] (5).
+  }
+
+  auto const fft = std::make_unique<FastFourierTransform<Value, Angle, 2 * N>>(
+      f_cos_N⁻¹π, N⁻¹π);
+  auto const& a = *fft;
+
+  // [Gen72b] (7), factoring out the division by N.
```
pleroy, line 142: "equation (7)".

numerics/quadrature_body.hpp
```diff
@@ -24,35 +34,133 @@ Primitive<std::invoke_result_t<Function, Argument>, Argument> Gauss(
...
+  Value Σʺ{};
+  for (std::int64_t n = 0; n <= N; n += 2) {
+    // The notation g is from [OLBC10], 3.5.17.
+    int gₙ = n == 0 || n == N ? 1 : 2;
```
pleroy, line 146: `const`

Note that the `?:` is likely to involve a jump.  A slightly faster way to write it would be to extract `n = 0` and `n = N` from the loop, and loop from `n = 1` to `n < N`.

numerics/quadrature_body.hpp
eggrobin: Shrug; let’s remain close to the notation from the paper.

numerics/quadrature_body.hpp
eggrobin: Shrug.

PR #2792, opened 2020-11-22
Add support for finding the stop_token for the current thread

pleroy:


base/jthread.hpp
```diff
@@ -101,12 +103,24 @@ class jthread {
   std::thread thread_;
 };
 
+class this_jthread {
```
eggrobin, line 106: Not all jthreads can access themselves via `this_jthread`, so the name is misleading ; we need a dedicated one.

`this_cancellable_thread`, `MakeCancellableThread` at namespace scope, because `this_cancellable_thread::Make` is awkward and similarly misleading (you are not making this thread, you are making a new one).
Cancellable after the `CANCELLED` status, which we will use for cancellation propagation.

No need to reify a `cancellable_thread` type, but using unique wording makes it clear that this only applies to threads created by the appropriate factory.

eggrobin (approved):

pleroy:
retest this please

PR #2799, opened 2020-11-27
Interrupt the orbit analysis if the mission duration changes

eggrobin:
In order to do that:
- refactor the orbit analyser a bit so that its thread can be restarted;
- fix the `operator=` of our implementation of `jthread`;
- introduce `RETURN_IF_STOPPED`;
- pepper it throughout `OrbitalElements` and related functions;
- avoid copies of large vectors, wherein we cannot `RETURN_IF_STOPPED`;
- in particular, update `Status` for move semantics (what year is this).

The analyser can lag a little bit when updating the mission duration during the actual analysis (as opposed to the integration), I think the remaining large atomic operations are vector destructions; not much we can easily do about those.

Drive-by changes: take an exclusive lock to swap, not a reader lock, don’t flow twice to the same time.

### DIFF
--- a/astronomy/orbital_elements_body.hpp
+++ b/astronomy/orbital_elements_body.hpp
@@ -46,7 +46,7 @@ StatusOr<OrbitalElements> OrbitalElements::ForTrajectory(
   orbital_elements.radial_distances_ = RadialDistances(trajectory);
   auto const sidereal_period =
       SiderealPeriod(orbital_elements.osculating_equinoctial_elements_);
-  RETURN_IF_ERROR(sidereal_period.status());
+  RETURN_IF_ERROR(sidereal_period);
   orbital_elements.sidereal_period_ = sidereal_period.ValueOrDie();
   if (!IsFinite(orbital_elements.sidereal_period_)) {
     // Guard against NaN sidereal periods (from hyperbolic orbits).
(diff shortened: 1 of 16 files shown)


astronomy/orbital_elements_body.hpp
```diff
@@ -42,17 +44,22 @@ StatusOr<OrbitalElements> OrbitalElements::ForTrajectory(
...
     // Guard against NaN sidereal periods (from hyperbolic orbits).
     return Status(
         Error::OUT_OF_RANGE,
         "sidereal period is " + DebugString(orbital_elements.sidereal_period_));
   }
-  orbital_elements.mean_equinoctial_elements_ =
+  auto mean_equinoctial_elements =
       MeanEquinoctialElements(orbital_elements.osculating_equinoctial_elements_,
                               orbital_elements.sidereal_period_);
+  RETURN_IF_ERROR(mean_equinoctial_elements.status());
```
pleroy, line 60: We should change the macro to automatically extract the status when give a `base::StatusOr`.

base/jthread.hpp
```diff
@@ -123,6 +124,13 @@ class this_stoppable_thread {
   friend jthread MakeStoppableThread(Function&& f, Args&&... args);
 };
 
+#define RETURN_IF_STOPPED                                                     \
```
pleroy, line 127: I am wondering if this should take (optionally) a parameter to let the caller propagate a useful message.

I am also wondering if our macro names should systematically start with `PRINCIPIA_`.

pleroy (approved):

eggrobin:
retest this please

PR #2806, opened 2020-12-03
Stop unused analysers, reduce the analysis frequency in the background

eggrobin:


ksp_plugin/orbit_analyser.cpp
```diff
@@ -31,21 +31,22 @@ OrbitAnalyser::OrbitAnalyser(
       analysed_trajectory_parameters_(
           std::move(analysed_trajectory_parameters)) {}
 
-void OrbitAnalyser::Restart() {
-  analyser_ = MakeStoppableThread([this] { RepeatedlyAnalyseOrbit(); });
+void OrbitAnalyser::Interrupt() {
+  analyser_ = jthread();
```
pleroy, line 35: I would `request_stop` and `detach` (only if the thread is joinable, of course).  The thread would immediately become non-joinable, and could happily execute for a while until it notices the stop.  You could then drop the `analyzer_done_` boolean since it would effectively be subsumed by `joinable`.

ksp_plugin/orbit_analyser.cpp
```diff
@@ -82,8 +83,9 @@ Status OrbitAnalyser::RepeatedlyAnalyseOrbit() {
     {
       absl::MutexLock l(&lock_);
       if (!guarded_parameters_.has_value()) {
-        // No parameters, let's wait for them to appear.
-        continue;
+        // No parameters, stop.
+        analyser_done_ = true;
+        return Status::OK;
```
pleroy, line 88: If you detach, you don't need to change anything here, it will ultimately land on a `RETURN_IF_STOPPED`.

ksp_plugin/orbit_analyser.cpp
eggrobin: Rewrote as discussed.

ksp_plugin/orbit_analyser.cpp
eggrobin: Rewrote as discussed.

ksp_plugin/orbit_analyser.cpp
```diff
@@ -70,25 +74,19 @@ double OrbitAnalyser::progress_of_next_analysis() const {
...
-      if (!guarded_parameters_.has_value()) {
-        // No parameters, let's wait for them to appear.
-        continue;
-      }
-      std::swap(guarded_parameters, guarded_parameters_);
-    }
-
-    auto const& parameters = guarded_parameters->parameters;
+Status OrbitAnalyser::AnalyseOrbit(
+    GuardedParameters guarded_parameters) {
```
pleroy, line 78: Would fit on the previous line?

ksp_plugin/orbit_analyser.cpp
```diff
@@ -182,10 +180,12 @@ Status OrbitAnalyser::RepeatedlyAnalyseOrbit() {
     {
       absl::MutexLock l(&lock_);
       next_analysis_ = std::move(analysis);
+      // Take ownership of our own thread so we can safely detach.
+      analyser = std::move(analyser_);
```
pleroy, line 184: FWIW, the usual pattern to do this is to `swap`.  This avoids having to reason about moved-from objects (`analyzer_` becomes a well-defined, default-initialized `jthread`.

pleroy (approved):

PR #2807, opened 2020-12-03
Extended mission in the analyser

eggrobin:
This allows the analyser to try to see one full revolution if the `mission_duration` appears to be too short. This will be useful in the flight plan to figure out what orbit the vessel would be in if the orbit were not interrupted by the next manœuvre.

### DIFF
--- a/ksp_plugin/orbit_analyser.hpp
+++ b/ksp_plugin/orbit_analyser.hpp
@@ -89,6 +89,9 @@ class OrbitAnalyser {
     Instant first_time;
     DegreesOfFreedom<Barycentric> first_degrees_of_freedom;
     Time mission_duration;
+    // The analyser may compute the trajectory up to |extended_mission_duration|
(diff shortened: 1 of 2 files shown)


ksp_plugin/orbit_analyser.hpp
```diff
@@ -89,6 +89,9 @@ class OrbitAnalyser {
     Instant first_time;
     DegreesOfFreedom<Barycentric> first_degrees_of_freedom;
     Time mission_duration;
+    // The analyser may compute the trajectory up to |extended_mission_duration|
+    // to ensure that at least one revolution is analysed.
+    std::optional<Time> extended_mission_duration;
```
pleroy, line 94: I can't find a place that sets this parameter.  What am I missing?

eggrobin: It is currently unused.

It will be used in flight plan analysis.

PR #2832, opened 2020-12-30
Ifdef out a test that fails to compile with Clang

pleroy:


eggrobin (changes requested):
Remove the random txt files.

eggrobin (approved):

PR #2837, opened 2020-12-31
Update all predictions at the same time

### LINKED ISSUE #2824: Predicted trajectory updates very slowly whilst a target is selected
The predicted trajectory, and also the flight plan in map view update very slowly whilst a target is selected, but the view is not in target centric mode.
For example, selecting a target while performing a manoeuvre with the view in MCI will result in the prediction "freezing" or updating very slowly. Clearing the target will resume normal update rate.

eggrobin:
Fix #2824.
The renderer only has a target vessel when we are in the target frame; otherwise, from the point of view of the plugin, it is like any other vessel; by stopping all prognosticators but the main vessel’s and, immediately afterwards, all prognosticators but the target vessel’s, we stop the prognosticator of the main vessel—and thus its prediction is never recomputed.

### DIFF
--- a/ksp_plugin/interface.cpp
+++ b/ksp_plugin/interface.cpp
@@ -18,6 +18,7 @@
 #include <psapi.h>
 #endif
 
+#include "absl/strings/str_split.h"
 #include "astronomy/epoch.hpp"
 #include "astronomy/time_scales.hpp"
 #include "base/array.hpp"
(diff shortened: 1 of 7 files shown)


pleroy:
Could you explain in the description of the PR how it fixes the issue?

pleroy:
retest this please

pleroy:
retest this please

PR #2845, opened 2021-01-02
Support for repeated strings in the generator

### LINKED ISSUE #2838: Add support for repeated strings in the interface
#2837 uses unit separators like it’s 1963.

pleroy:
Fix #2838.


ksp_plugin_adapter/ksp_plugin_adapter.cs
```diff
@@ -423,9 +423,9 @@ class PartCentredForceHolder {
         plugin_.VesselSetPredictionAdaptiveStepParameters(
             target_id, adaptive_step_parameters);
         plugin_.UpdatePrediction(
-            string.Join("\x1F", main_vessel.id.ToString(), target_id));
+            new string[]{main_vessel.id.ToString(), target_id});
```
eggrobin, line 426: Could we use `params` to make the calling code nicer? https://docs.microsoft.com/en-nz/dotnet/csharp/language-reference/keywords/params

pleroy: I'd rather not.  It's not even clear from the documentation that `params` can have a custom marshaller, and we have seen in the past that even things that are documented don't work the way they should.  Not to mention our good friend Mono.

eggrobin (approved):

PR #2848, opened 2021-01-03
Plan analysis

eggrobin:
Internals: the `FlightPlan` analyses all coasts; more specifically, for non-final coasts, it may analyse their extension (using `extended_mission_duration` as if the next manœuvre did not happen (so that a coast for a partial revolution in Earth orbit is still analysed as a coast in Earth orbit).

UI: For non-final coasts, the UI uses that analysis to describe the coast, e.g., `Coast in circular Earth orbit for 3 h 00 min`.
For the final trajectory, the full orbit analysis UI is available at the press of a button.

screenshot21


ksp_plugin/flight_plan.cpp
```diff
@@ -67,6 +67,8 @@ FlightPlan::FlightPlan(
 
   // Create a fork for the first coasting trajectory.
   segments_.emplace_back(root_->NewForkWithoutCopy(initial_time_));
+  coast_analysers_.push_back(make_not_null_unique<OrbitAnalyser>(
+      ephemeris_, DefaultHistoryParameters()));
```
pleroy, line 71: Instead of managing all these arrays in parallel, could we create a small `struct Segment` containing a trajectory and (optionally) an analyser?  That would be less error-prone and more future proof.

ksp_plugin/flight_plan.cpp
```diff
@@ -410,6 +437,12 @@ Status FlightPlan::ComputeSegments(
         anomalous_segments_ = 1;
         anomalous_status_ = status;
       }
+      coast_analysers_[it - manœuvres_.begin()]->RequestAnalysis(
```
pleroy, line 440: It's odd that we are doing this even if the coast is anomalous.  Same comment below.  If a coast is anomalous its analyser should stop (and those of the following coasts).

ksp_plugin_adapter/orbit_analyser.cs
```diff
@@ -505,15 +505,78 @@ internal class OrbitAnalyser : VesselSupervisedWindowRenderer {
...
           value = 7 * 24 * 60 * 60
       };
-  private bool autodetect_recurrence_ = true;
-  private int revolutions_per_cycle_ = 1;
-  private int days_per_cycle_ = 1;
-  private int ground_track_revolution_ = 1;
+  protected bool autodetect_recurrence_ = true;
+  protected int revolutions_per_cycle_ = 1;
+  protected int days_per_cycle_ = 1;
+  protected int ground_track_revolution_ = 1;
```
pleroy, line 511: It would be good if we avoided protected fields.  I mean, we avoid them in C++, right?

eggrobin: Done.

ksp_plugin/flight_plan.cpp
eggrobin: No: if the flight plan’s synchronous computation of the coast times out before the end, it still makes sense to run the asynchronous analysis.

Analyses are not requested for coasts that are entirely past the anomaly though.

ksp_plugin/flight_plan.cpp
eggrobin: No. We keep the analys*ers* when we recompute the segments (and the analys*es*).  As analysis is asynchronous, we want to keep the old analysis around until the new one is ready (this is especially critical when using the sliders: continuous changes to the flight plan are reflected in the UI with a slight delay, instead of the UI being constantly blank until the slider is released).
The analysers work more like the manœuvres in that respect: we insert and erase them when a manœuvre is added or removed, instead of pushing and popping them every time the flight plan changes as we do for the segments.

ksp_plugin/orbit_analyser.hpp
```diff
@@ -98,6 +98,8 @@ class OrbitAnalyser {
                 Ephemeris<Barycentric>::FixedStepParameters
                     analysed_trajectory_parameters);
 
+  ~OrbitAnalyser();
```
pleroy, line 101: `virtual`

pleroy (approved):

pleroy:
Still LGTM.

Issue #2590, opened 2020-05-23
Vessels teleport/switch places upon EVA


gert7:
[Image](https://i.redd.it/q8cjpozv7k051.jpg)

1.9.1 Fuchs. This happened to my Apollo 11-style flight. When I noticed I hadn't moved my crew over to the lander I decided to EVA. The two undocked vessel sides switched places when I hit the EVA. I tried this again in an earlier quicksave and the same thing happened with a much greater distance between the vessels.

pleroy:
Are you using any mods besides Principia?  Can you give us your save and the logs at the time where the glitch happened?

gert7:
[Eva.zip](https://github.com/mockingbirdnest/Principia/files/4673099/Eva.zip)

Quicksave here can reproduce the bug after undocking the middle and switching to EVA. To me it seems to only move vessels along their trajectories, namely the vessel being EVA-d from moves forward some amount. The distance seems to vary, velocities and rotations are completely unaffected.

gert7:
And no, Principia is the only mod I have ever installed in this installation of KSP.

pleroy:
I have reproduced the problem and I have a journal.  This is going to be challenging to debug.

pleroy:
When the Kerbal goes in EVA, it is unready (and therefore unmanageable by Principia) for a short period (probably 1 frame).  If, during that period, it collides with the vessel, this makes the vessel unmanageable and Principia forgets about it.  Shortly thereafter the Kerbal becomes ready, and the vessel is re-created by Principia, *using the position provided by KSP*.  That position may differ significantly from the "real" one and the vessel appears to jump.

The solution is to ignore collisions with an unready Kerbal.  Unreadiness is a transient state anyway, so at the next frame the Kerbal will be ready and will nicely pile-up with the vessel.

There have been reports in the past of the history being lost when doing EVA.  That's probably the same cause (deletion and re-creation of a vessel).  I checked that the history is preserved in this instance.

PR #2852, opened 2021-01-08
KSP 1.11.0, and an ancient bug

eggrobin:
It looks like the « ex nihilo » path had a bug. Previously, this path only was only taken when Principia took control of a new vessel:
- on liftoff this happens with vanishing krakensbane;
- for EVA kerbals the mass of the kerbal was probably too small for the displacement of the vessel to be noticeable—and Kerbals have their own weirdnesses when spawning anyway.

The new welding mechanic creates parts ex nihilo, including fairly heavy ones, so this is far more visible.

This may explain the « teleportation » aspect in https://github.com/mockingbirdnest/Principia/issues/2590#issuecomment-639648219:
> the position provided by KSP […] may differ significantly from the "real" one

It probably differed because it was misinterpreted: KSP should largely be consistent with us over small time spans.

### DIFF
--- a/ksp_plugin_adapter/gl_lines.cs
+++ b/ksp_plugin_adapter/gl_lines.cs
@@ -149,7 +149,7 @@ internal static class GLLines {
     get {
       if (line_material_ == null) {
         line_material_ = new UnityEngine.Material(
-#if KSP_VERSION_1_10_1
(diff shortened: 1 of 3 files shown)


pleroy:
retest this please

PR #2868, opened 2021-01-21
Check that the version of the binary matches that of the journal

### LINKED ISSUE #2863: The journal for #2853 fails when replaying a DeleteInterchange message
`DeserializePointer` fails to find a pointer in the `pointer_map`.

pleroy:
The referenced bug was a pilot error: I was replaying with Germain (at head, really) a journal created with Гельфонд.  Confusion ensued.  I am adding a check to help me next time this happens.

As part of investigating this I realized that the calls to `SerializePlugin` that happen when replaying may not match those of the journal, presumably because the state of the plugin is different from that at the time of journaling.  This manifests itself as mismatching `DeleteString` not finding a pointer in the map.  I presume that this happens because differences in floating point computations lead to differences e.g., in the size of trajectories.  This is unfortunate, but hard to fix.

Fix #2863.

### DIFF
--- a/journal/player.hpp
+++ b/journal/player.hpp
@@ -8,6 +8,8 @@
 
 #include "serialization/journal.pb.h"
 
+#define PRINCIPIA_PLAYER_ALLOW_VERSION_MISMATCH 1
+
 namespace principia {
 namespace journal {
 
--- a/journal/player.hpp
+++ b/journal/player.hpp
@@ -8,7 +8,7 @@
 
 #include "serialization/journal.pb.h"
 
-#define PRINCIPIA_PLAYER_ALLOW_VERSION_MISMATCH 1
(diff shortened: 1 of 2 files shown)


journal/player.hpp
```diff
@@ -8,6 +8,8 @@
 
 #include "serialization/journal.pb.h"
 
+#define PRINCIPIA_PLAYER_ALLOW_VERSION_MISMATCH 1
```
eggrobin, line 11: Maybe default to 0 so you notice next time?

eggrobin (approved):

journal/player.hpp
pleroy: I set it to 1 to check that it was properly disabling the check and forgot to flip back.  Sigh.

PR #2869, opened 2021-01-23
Another test of diagonalization and fixes in the code generated for repeated strings

### LINKED ISSUE #2866: Crash when separating a stage with many struts or when there is a lot of debris.
Hello. First let me say that I really, really love this mod and have been able to use it with no other issues, for the most part. However, whenever I separate a stage with multiple struts or a ship is destroyed and produces a lot of debris, KSP (v. 1.10.1) crashes and produces a fatal error; an example would the 7th stage of the stock Acapello rocket.
   I thought this might be an issue with RAM usage, but according to Task Manager RAM usage does not increase above normal levels during the crash. I am able replicate this crash every time I try. Not sure whether this is a bug or a performance issue with my machine (Intel Core i7-9750, RTX 2060 6GB GPU, 8GB RAM).
   I can repeat this crash on a fresh install of KSP with no mods other than Principia. I have provided the info.log and fatal.log files.

Many Thanks.
[FATAL.20210119-202801.2084.log](https://github.com/mockingbirdnest/Principia/files/5838832/FATAL.20210119-202801.2084.log)
[INFO.20210119-202641.2084.log](https://github.com/mockingbirdnest/Principia/files/5838833/INFO.20210119-202641.2084.log)

pleroy:
Fix #2866.

### DIFF
--- a/geometry/symmetric_bilinear_form_test.cpp
+++ b/geometry/symmetric_bilinear_form_test.cpp
@@ -286,7 +286,6 @@ TEST_F(SymmetricBilinearFormTest, Serialization) {
 }
 
 TEST_F(SymmetricBilinearFormTest, Diagonalize) {
-#if 0
   // A simple test where it's clear what the eigensystem should be.
   {
     auto const f = MakeSymmetricBilinearForm<World>(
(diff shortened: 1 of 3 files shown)


eggrobin (approved):

pleroy:
retest this please

PR #2870, opened 2021-01-23
Fix keming in the bibliography

### LINKED ISSUE #2856: BibTeX, capitalization, and kerning
Some of our BibTeX entries use curly brackets to protect capitalization:
https://github.com/mockingbirdnest/Principia/blob/cc26d0fd2171b0fa4bc0ee1249759e1509deab4c/documentation/bibliography.bib#L1003 https://github.com/mockingbirdnest/Principia/blob/cc26d0fd2171b0fa4bc0ee1249759e1509deab4c/documentation/bibliography.bib#L21 https://github.com/mockingbirdnest/Principia/blob/cc26d0fd2171b0fa4bc0ee1249759e1509deab4c/documentation/bibliography.bib#L656
others do not:
https://github.com/mockingbirdnest/Principia/blob/cc26d0fd2171b0fa4bc0ee1249759e1509deab4c/documentation/bibliography.bib#L412

In theory, the former style is preferable since it is compatible with lowercasing style; however, it appears than when this is desired, `{Runge}` is the proper form, not `{R}unge`, since the latter interferes with kerning, see, e.g., https://tex.stackexchange.com/a/140071.

We could switch to {Runge}, {ODEs}, etc. On the other hand, our convention for case of bibliographic entries is “as in the document cited”, not “titlecased”, much like we preserve any orthographic quirks, e.g., Kahan’s space before the question mark in the title of _Mindless_, so our .bib file is incompatible with titlecasing styles anyway. Perhaps we should simply get rid of those curly brackets, it seems a bit odd to put in that work to support some of the styles we don’t use. In any case, we should avoid messing with kerning.

pleroy:
Fix #2856.


pleroy:
retest this please

pleroy:
retest this please

PR #2894, opened 2021-02-10
do a small change to make the describe text more precise

CindyRIng:


eggrobin:
Thank you for your contribution.
Note that this change will be in Goldbach (to be released on 2021-03-13) rather than Gödel (tomorrow), since we have already built the Gödel release at https://github.com/mockingbirdnest/Principia/tree/2021021119-Gödel.

PR #2897, opened 2021-02-18
Number manœuvres idiomatically, tweak time-related words

eggrobin:
As alluded to in #2890,
> Some day we may want to change `轨道机动 #𝑛` to `第𝑛次轨道机动` (I suspect this is currently messing with the word order when it comes to referring to the final time of a manœuvre), but this will have to come in a subsequent release.

Use ordinals (in the linguistic sense, _i.e._, « 𝑛th manœuvre », rather than cardinals as in « manœuvre number 𝑛 ») as they don’t require any particular logic in Chinese (contrary to [English -st -nd -rd -th](https://unicode-org.github.io/cldr-staging/charts/37/supplemental/language_plural_rules.html#en)).

Reworded starting and ending point to starting and ending time when referring to times.

Reworded 加速 to 时间加速 for « [time] warp »; 加速 simply means « accelerate », which is a lot more overloaded here than the original « warp ». 加速至轨道机动起点 « accelerate until manœuvre starting point » could have led to confusion, for which
时间加速至轨道机动开始时刻 « accelerate time until manœuvre starting time » should leave no room.

As a result of the numbering and point-to-time changes `轨道机动终点 #<<1>>` becomes `第<<1>>次轨道机动结束时`.

@Zaikarion pointed out that the term 时基 « time base » does not mean what we want it to.
Admittedly it is dubious in English to start with, as we are lifting its meaning of « time zero » from some obscure Apollo-Saturn documents. 时基 only has the more common [electronics-related](https://en.wikipedia.org/wiki/Time_base_generator) sense of « time base ». @Zaikarion suggested 计时起点 « timing start point ».

Screenshot of the flight plan UI, with changes highlighted:
image

### DIFF
--- a/ksp_plugin_adapter/localization/zh-cn.cfg
+++ b/ksp_plugin_adapter/localization/zh-cn.cfg
@@ -156,13 +156,13 @@ Localization {
     #Principia_FlightPlan_TotalΔv = 共计 Δv：<<1>> m/s  // <<1>>: Δv.ToString("0.000").
     #Principia_FlightPlan_Delete = 删除当前计划
     #Principia_FlightPlan_Rebase = 变基
-    #Principia_FlightPlan_ManœuvreHeader = 轨道机动 #<<1>>
-    #Principia_FlightPlan_UpcomingManœuvre = 即將执行轨道机动 #<<1>>：
+    #Principia_FlightPlan_ManœuvreHeader = 第<<1>>轨道机动
+    #Principia_FlightPlan_UpcomingManœuvre = 即將执行第<<1>>轨道机动：
     #Principia_FlightPlan_IgnitionCountdown = 至点火时间 <<1>>
-    #Principia_FlightPlan_OngoingManœuvre = 进行中的轨道机动 #<<1>>：
+    #Principia_FlightPlan_OngoingManœuvre = 进行中的第<<1>>轨道机动：
     #Principia_FlightPlan_CutoffCountdown = 至关机时间 <<1>>
     #Principia_FlightPlan_ShowManœuvreOnNavball = 在导航球上显示
-    #Principia_FlightPlan_WarpToManœuvre = 加速至轨道机动起点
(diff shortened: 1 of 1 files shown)


ksp_plugin_adapter/localization/zh-cn.cfg
```diff
@@ -174,21 +174,21 @@ Localization {
     #Principia_FlightPlan_StatusMessage_MaxSegment = 请增加“最大迭代步长”或避免与天体碰撞
     #Principia_FlightPlan_StatusMessage_Singularity = 积分器遇到奇点（可能是天体的中心）<<1>>  // <<1>> timeout
     #Principia_FlightPlan_StatusMessage_AvoidingCollision = 请避免与天体的碰撞
-    #Principia_FlightPlan_StatusMessage_Infinite = 轨道机动 #<<1>> 会产生无限或不确定的速度  // <<1>> Error Manœuvre
-    #Principia_FlightPlan_StatusMessage_Adjust = 调整轨道机动的持续时间 #<<1>>  // <<1>> Error Manœuvre
-    #Principia_FlightPlan_StatusMessage_OutRange1 = 轨道机动 #<<1>> 与 <<2>> 重叠 或者 <<3>>  // TODO(phl): Added <<3>> at a random place.
+    #Principia_FlightPlan_StatusMessage_Infinite = 第<<1>>次轨道机动 会产生无限或不确定的速度  // <<1>> Error Manœuvre
+    #Principia_FlightPlan_StatusMessage_Adjust = 调整第<<1>>次轨道机动的持续时间  // <<1>> Error Manœuvre
+    #Principia_FlightPlan_StatusMessage_OutRange1 = 第<<1>>次轨道机动 与 <<2>> 重叠 或者 <<3>>  // TODO(phl): Added <<3>> at a random place.
```
pleroy, line 179: Can we remove the TODO now?

pleroy (approved):

ksp_plugin_adapter/localization/zh-cn.cfg
eggrobin: No, it is still at a random place.

PR #2898, opened 2021-02-22
Prevent Vessel::RepeatedlyFlowPrognostication from bypassing sleep call.

rnlahaye:
Previously, Vessel::RepeatedlyFlowPrognostication would loop without sleeping while prognosticator_parameters_ was empty (because the continue bypasses the std::this_thread::sleep_until call). This resulted in the thread spending 40% of its time calling std::chrono::steady_clock::now and another 20% dealing with the mutex (see attached screenshot).
too_much_now

This change moves the std::this_thread::sleep_until call to the beginning of the loop where it cannot be bypassed.

### DIFF
--- a/ksp_plugin/vessel.cpp
+++ b/ksp_plugin/vessel.cpp
@@ -577,9 +577,12 @@ void Vessel::StartPrognosticatorIfNeeded() {
 }
 
 Status Vessel::RepeatedlyFlowPrognostication() {
+  std::chrono::steady_clock::time_point wakeup_time =
+      std::chrono::steady_clock::now();
   for (;;) {
+    std::this_thread::sleep_until(wakeup_time);
     // No point in going faster than 50 Hz.
-    std::chrono::steady_clock::time_point const wakeup_time =
+    wakeup_time =
         std::chrono::steady_clock::now() + std::chrono::milliseconds(20);
 
     RETURN_IF_STOPPED;
@@ -607,8 +610,6 @@ Status Vessel::RepeatedlyFlowPrognostication() {
       absl::MutexLock l(&prognosticator_lock_);
       SwapPrognostication(prognostication, status);
     }
-
-    std::this_thread::sleep_until(wakeup_time);
   }
   return Status::OK;
 }
--- a/ksp_plugin/vessel.cpp
+++ b/ksp_plugin/vessel.cpp
@@ -577,7 +577,7 @@ void Vessel::StartPrognosticatorIfNeeded() {
 }
 
 Status Vessel::RepeatedlyFlowPrognostication() {
-  for (auto wakeup_time = std::chrono::steady_clock::now();;
+  for (std::chrono::steady_clock::time_point wakeup_time;;
        std::this_thread::sleep_until(wakeup_time)) {
     // No point in going faster than 50 Hz.
     wakeup_time =
--- a/ksp_plugin/vessel.cpp
+++ b/ksp_plugin/vessel.cpp
@@ -577,14 +577,10 @@ void Vessel::StartPrognosticatorIfNeeded() {
 }
 
 Status Vessel::RepeatedlyFlowPrognostication() {
-  std::chrono::steady_clock::time_point wakeup_time =
-      std::chrono::steady_clock::now();
-  for (;;) {
-    std::this_thread::sleep_until(wakeup_time);
-    // No point in going faster than 50 Hz.
-    wakeup_time =
-        std::chrono::steady_clock::now() + std::chrono::milliseconds(20);
-
+  // No point in going faster than 50 Hz.
+  for (auto wakeup_time = std::chrono::steady_clock::now();;
+       wakeup_time += std::chrono::milliseconds(20),
+            std::this_thread::sleep_until(wakeup_time)) {
     RETURN_IF_STOPPED;
 
     std::optional<PrognosticatorParameters> prognosticator_parameters;
--- a/ksp_plugin/vessel.cpp
+++ b/ksp_plugin/vessel.cpp
@@ -577,10 +577,11 @@ void Vessel::StartPrognosticatorIfNeeded() {
 }
 
 Status Vessel::RepeatedlyFlowPrognostication() {
-  // No point in going faster than 50 Hz.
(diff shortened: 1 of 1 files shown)


ksp_plugin/vessel.cpp
```diff
@@ -577,9 +577,12 @@ void Vessel::StartPrognosticatorIfNeeded() {
...
 
 Status Vessel::RepeatedlyFlowPrognostication() {
+  std::chrono::steady_clock::time_point wakeup_time =
+      std::chrono::steady_clock::now();
   for (;;) {
+    std::this_thread::sleep_until(wakeup_time);
     // No point in going faster than 50 Hz.
-    std::chrono::steady_clock::time_point const wakeup_time =
+    wakeup_time =
         std::chrono::steady_clock::now() + std::chrono::milliseconds(20);
```
pleroy, line 586: It would probably be better to write this:
```
wakeup_time += std::chrono::milliseconds(20);
```
This would save reading the clock each time, and it would avoid drifts in the wake-up time of the loop (if the thread is suspended between waking-up and reading the clock).

pleroy (commented):
Thanks for another great contribution, and egg on my face...

pleroy:

ksp_plugin/vessel.cpp
rnlahaye: Done (using eggrobin@'s suggestion).

ksp_plugin/vessel.cpp
```diff
@@ -577,9 +577,12 @@ void Vessel::StartPrognosticatorIfNeeded() {
 }
 
 Status Vessel::RepeatedlyFlowPrognostication() {
+  std::chrono::steady_clock::time_point wakeup_time =
+      std::chrono::steady_clock::now();
   for (;;) {
```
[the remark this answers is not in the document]
rnlahaye: Done.

rnlahaye:
ok to test

rnlahaye:
I have a question: what is the intended behavior when the body of the loop takes more than 20ms to execute? This case has different behavior between the old and new implementations—in the new implementation, if an iteration of the loop is slow then future iterations will run at faster than 50Hz frequency to catch up.

pleroy:
> I have a question: what is the intended behavior when the body of the loop takes more than 20ms to execute? This case has different behavior between the old and new implementations—in the new implementation, if an iteration of the loop is slow then future iterations will run at faster than 50Hz frequency to catch up.

Hmm, that's a good point.  If the computation is fast (< 20 ms) we don't want to go faster than the frame refresh rate (that would waste CPU).  If the computation is slow (> 20 ms) we want to do it as frequently as possible to get a decent user experience.  As you point out, the proposed code creates a "debt" in the latter case that makes no sense.  

So we need to go back to something like the original computation of `wakeup_time`, with some care because it's now happening at the *end* of the loop, not at the beginning as it previously did.  I believe that the following would work:

```
for (...;; wakeup_time = std::max(std::chrono::steady_clock::now(), wakeup_time + std::chrono::milliseconds(20)), ...) {...}
```

ksp_plugin/vessel.cpp
```diff
@@ -577,11 +577,10 @@ void Vessel::StartPrognosticatorIfNeeded() {
...
 Status Vessel::RepeatedlyFlowPrognostication() {
-  for (;;) {
-    // No point in going faster than 50 Hz.
-    std::chrono::steady_clock::time_point const wakeup_time =
-        std::chrono::steady_clock::now() + std::chrono::milliseconds(20);
-
+  // No point in going faster than 50 Hz.
+  for (auto wakeup_time = std::chrono::steady_clock::now();;
+       wakeup_time += std::chrono::milliseconds(20),
+            std::this_thread::sleep_until(wakeup_time)) {
```
pleroy, line 583: Nit: the indent is weird.  Probably move left by 1 position.

eggrobin:
> So we need to go back to something like the original computation of `wakeup_time`, with some care because it's now happening at the _end_ of the loop, not at the beginning as it previously did. I believe that the following would work:
> 
> ```C++
> for (auto wakeup_time = std::chrono::steady_clock::now();;
>      wakeup_time = std::max(std::chrono::steady_clock::now(), 
>                             wakeup_time + std::chrono::milliseconds(20)),
>          std::this_thread::sleep_until(wakeup_time)) {
> ```

We could also reorder the comma and keep the old computation:
```C++
for (auto wakeup_time =
         std::chrono::steady_clock::now() + std::chrono::milliseconds(20);;
     std::this_thread::sleep_until(wakeup_time),
         wakeup_time =
             std::chrono::steady_clock::now() + std::chrono::milliseconds(20))) {
```
but I think it is going to be clearer to make it actually happen at the beginning of the loop:
```C++
for (std::chrono::steady_clock::time_point wakeup_time;;
     std::this_thread::sleep_until(wakeup_time)) {
  wakeup_time = std::chrono::steady_clock::now() + std::chrono::milliseconds(20);
```

ksp_plugin/vessel.cpp
rnlahaye: It now looks a bit different since the wakeup_time computation has been moved into the loop body. Does it still look weird? FYI I'm using clang_format for formatting.

rnlahaye:
I moved the wakeup_time computation to the beginning of the loop body as suggested by @eggrobin.

ksp_plugin/vessel.cpp
```diff
@@ -577,11 +577,11 @@ void Vessel::StartPrognosticatorIfNeeded() {
 }
 
 Status Vessel::RepeatedlyFlowPrognostication() {
-  for (;;) {
+  for (auto wakeup_time = std::chrono::steady_clock::now();;
```
eggrobin, line 580: That initialization isn’t doing anything anymore (except deducing the type, but that’s a slightly confusing use of `now`).
`std::chrono::steady_clock::time_point wakeup_time;;`.

eggrobin (approved):

ksp_plugin/vessel.cpp
rnlahaye: Done.

Issue #2899, opened 2021-02-24
Principia is slow on macOS (possibly due to Unity's allocator)


rnlahaye:
TLDR: Unity seems to have a custom allocator that uses mutexes. Mutexes are slow on macOS. Consequently, on macOS the vast majority of Principia's time seems to be spent on memory management.

After getting back into KSP after a hiatus, I found the game to have poor performance. Running a trace led me to discover a bug in `Vessel::RepeatedlyFlowPrognostication` but even after fixing it (#2898), the poor performance persisted. Sometimes the game would even pause completely for several seconds. Further traces revealed the problem was due to four threads concurrently evaluating `OrbitAnalyser::AnalyseOrbit`. Profiling an AnalyseOrbit benchmark revealed nothing amiss.
However, I was able to get a flame graph ([attached](https://github.com/mockingbirdnest/Principia/files/6032965/principia_macos_flamegraph.svg.zip)) of a running game which revealed the problem:
![Screen Shot 2021-02-23 at 6 05 12 PM](https://user-images.githubusercontent.com/71856888/108934408-b8a81480-7601-11eb-922b-ba45d3bbe7ae.png)
The regions highlighted in magenta are mutexes used by functions in `UnityPlayer.dylib`. Based on the placement of the `UnityPlayer.dylib` functions in the call stack (and the fact that one of them is `operator new`), I suspect they are a custom allocator used by Unity. Unfortunately, the performance of the stock mutex on macOS is known to be bad (#1955). Thus, terrible performance. The reason this was not apparent in benchmarks is probably because the benchmarks are not run from Unity and hence use the system allocator. This isn't restricted to `AnalyzeOrbit` either; the rest of Principia was also affected. I estimate that over 80% of Principia's CPU cost on my machine is spent managing these mutexes!

Something similar was diagnosed and fixed in #1955. Unfortunately, this time the slow mutexes are not Principia's but Unity's mutexes. Replacing the misbehaving mutexes with `absl::Mutex` is not really an option here.

I will try to fix this by forcing Principia to use the system allocator even when run from Unity. Stay tuned.

Possibly related to #2247.

pleroy:
This is a very interesting finding, and it would explain why people keep complaining about performance on macOS (we don't play on that platform).

[...]

PR #2900, opened 2021-03-01
Use system allocator for containers on macOS (bypassing Unity's allocator)

rnlahaye:
First part of a two-part fix for #2899.

This change causes STL containers (std::vector and the like) to use the native system allocator on macOS instead of the Unity-provided allocator.

It does this by defining an allocator which avoids using of `new` and `delete`, and then creating aliases to `std::vector` and friends in the `principia::std` namespace using the new allocator. The header file containing this change is included via compiler argument, so this change is almost entirely transparent to the rest of the codebase (I had to add full qualification in two places).


pleroy:
FWIW, this PR generally looks good.  We'll do a more thorough review over the week-end.  Expect nits, but nothing earth-shattering.

Makefile
```diff
@@ -87,6 +87,7 @@ SHARED_ARGS   := \
 	-Wall -Wpedantic                                              \
 	-Wno-char-subscripts                                          \
 	-Wno-gnu-anonymous-struct                                     \
+	-Wno-c99-extensions                                           \
```
pleroy, line 90: Just out of curiosity, which warning does this suppress?

magic/malloc_allocator.hpp
```diff
@@ -0,0 +1,110 @@
+// This file contains two things:
```
pleroy, line 1: Among other things, the `magic` directly is going to interact poorly with VS.  I would organize things differently here:

1. A file `base/malloc_allocator.hpp` that declares and defines the class `MallocAllocator` in namespace `principia::base`.  This one is not magic in any way, and should be included in the usual manner.
2. A file `base/macos_allocator_replacement.hpp` that has all the stuff declared in `principia::std`.  This is the one that is magically included.  For safety, it should have preprocessor directives that ensure that it's never pulled for other platforms:
```c++
#if !OS_MACOSX
#error Only include for macOS
#endif 
... all the rest ...
```
You'll need to include `base/macros.hpp` in that second file.

magic/malloc_allocator.hpp
```diff
@@ -0,0 +1,110 @@
...
+template <typename Key, typename Compare = std::less<Key>,
+          typename Allocator = allocator<Key>>
+using multiset = ::std::multiset<Key, Compare, Allocator>;
+
+template <typename Key, typename T, typename Compare = std::less<Key>,
+          typename Allocator = allocator<std::pair<const Key, T>>>
+using multimap = ::std::multimap<Key, T, Compare, Allocator>;
+
+}  // namespace std
+
```
pleroy, line 109: Remove blank line.

rnlahaye: Done.

magic/malloc_allocator.hpp
rnlahaye: Done.

Makefile
rnlahaye:
```
./astronomy/orbital_elements_body.hpp:165:10: warning: designated initializers
      are a C99 feature [-Wc99-extensions]
        {.t = time,
         ^~~~~~~~~
./astronomy/orbital_elements_body.hpp:166:10: warning: designated initializers
      are a C99 feature [-Wc99-extensions]
         .a = *osculating_elements.semimajor_axis,
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
./astronomy/orbital_elements_body.hpp:167:10: warning: designated initializers
      are a C99 feature [-Wc99-extensions]
         .h = e * Sin(ϖ),
         ^~~~~~~~~~~~~~~
```
…and so on.

eggrobin: Ah, right, these are C++20 whereas the flags say 1z=17.

eggrobin (approved):

PR #2905, opened 2021-03-07
Some fixes to our LaTeX package

eggrobin:
Use the maths font (currently Cambria Math) rather than the text font (currently Linux Libertine) for mathrm and the like (in practice, operator names such as log, cos, etc.).
Note that the PDFs will need to be regenerated to reflect that.

Remove the now seemingly obsolete workaround for mathmode starting with a delimiter, which made typesetting exponential in the number of nested brackets.

Add a reference for the « well-known hack ».


pleroy:
If the changes to the fonts require to regenerate the PDFs, could we do that in this PR?  I would like to see the effect of the change.

eggrobin:
> If the changes to the fonts require to regenerate the PDFs, could we do that in this PR? I would like to see the effect of the change.

Done except for the three most ancient PDFs (`Hamiltonian Mechanics` and `ODEs and Runge-Kutta integrators` should probably be removed; `Composition methods` needs quite a bit of cleanup, _e.g._, it should refer to the bibliography; it also fails to compile).
Note that `Perspective hiding` had to be tweaked due to changes in `tkz-euclide`.

pleroy:
Regenerating the PDFs has the unfortunate consequence that all files now have the date "2021-03-08" in their text.  Could we try to stick to the original date, perhaps by using the last update timestamp of the `.tex` file?

eggrobin:
> Regenerating the PDFs has the unfortunate consequence that all files now have the date "2021-03-08" in their text. Could we try to stick to the original date, perhaps by using the last update timestamp of the `.tex` file?

Not easily as far as I can tell; perhaps we could make XeLaTeX call git and look at the last commit, but the timestamp of the file may not be what we want either anyway: 5315ee68b4770cb4b78f8eb2057c0b74a791c997 technically changed `Perspective hiding.tex`, although it had no visual effect; similarly if we replaced `i.e.` by `\emph{i.e.}` (or `\idest`) changing the date would be weird.

Perhaps we use a manual date for all these documents, with `\date`.

pleroy:
> Perhaps we should use a manual date for all these documents, with `\date`.

Ok, let's do that, using the date found in the current PDFs.

PR #2914, opened 2021-03-13
Bibliography cleanups

eggrobin:
This change:
- moves the `biblatex` package and its customizations to the `basestyle` template;
- moves bibliography printing at the end of the document to the `basestyle` template;
- ensures that bibliography labels are consistent across all documents—[Fuk11a] was previously being cited as [Fuk11] in `Fukushima.pdf`;
- customizes `labelalpha` so that it includes the custom `userd` (`user`-defined field, `d`isambiguator) field;
- ensures that no ambiguity is left once `userd` is factored in, _i.e._, that the automated disambiguator `extraalpha` is empty, ensuring stability of labels;
- sets `userd`:
  - consistently with the existing labels if those were cited (most cases) changing the `.bib` ID to match if needed, or,
  - consistently with the `.bib` IDs otherwise ([Fan91a], [Fan91b]);
- changes `@incollection` to `@inproceeding` where the collection is specifically conference proceedings;
- changes `@incollection` to `@article` for [Del90], since that collection is a periodical with an ISSN;
- adds `eventdate` and `venue` to all `@inproceeding` entries, as well as some other fields as appropriate.

In order to make diffing easier, this change does not re-alphabetize the entries; this will be done in a subsequent pull request using the usual script.


documentation/Geopotential.tex
```diff
@@ -1,8 +1,5 @@
 \documentclass[10pt, a4paper, twoside]{basestyle}
 
-\usepackage[backend=biber,firstinits=true,maxnames=100,style=alphabetic,maxalphanames=4,doi=true,isbn=false,url=false,eprint=true]{biblatex}
-\bibliography{bibliography}
-
```
pleroy, line 5: This changes the spacing between the bibliographical id and the bibliographical entry in the References section (previously the entry started roughly below the "n" of "References", now it's below the "e".  This less visually pleasant.  In particular in this document this leads to a [ridiculously overfull hbox](https://idf.social/system/media_attachments/files/000/003/016/original/a8cacb25eed2eeb3.png?1536755438) when in the `[Wes17]` reference.

documentation/bibliography.bib
```diff
@@ -1369,73 +1394,102 @@ @incollection{Halley1809
...
+  booktitle = {AIAA/AAS Astrodynamics Specialist Conference and Exhibit},
+  eventdate = {2006-08-21/2006-08-24},
+  date      = {2006-08-24T09:00:00},
   doi       = {10.2514/6.2006-6750},
-  series    = {Collection of Technical Papers},
+  note      = {AIAA 2006-6750},
+  publisher = {American Institute of Aeronautics and Astronautics},
+  series    = {Guidance, Navigation, and Control and Co-located Conferences},
   title     = {Repeat Ground Track Lunar Orbits in the Full-Potential Plus Third-Body Problem},
-  volume    = {3},
```
pleroy, line 1427: Why did we lose the volume?

eggrobin: As best I can tell it was nonsense.
_Guidance, Navigation, and Control and Co-located Conferences_ is unnumbered, so it was not a `number` (a lot of `volumes` were « volume within the `series` » which should go into `number` in `@inproceedings`).
_AIAA/AAS Astrodynamics Specialist Conference and Exhibit_ (2006) does not appear to be partitioned in `volume`s either: https://arc.aiaa.org/doi/book/10.2514/MAST06.

Note that the previous value of `series`, _Collection of Technical Papers_, seems nonsensical.

pleroy: This came from [the citation on ResearchGate](https://www.researchgate.net/publication/228830968_Repeat_Ground_Track_Lunar_Orbits_in_the_Full-Potential_Plus_Third-Body_Problem/citation/download).

eggrobin: Ah, it looks like in print these proceedings appeared in the three-volume ISBN 1563478226, _A collection of technical papers_:
1. https://www.tib.eu/en/search/id/TIBKAT:521489660/A-collection-of-technical-papers-AIAA-AAS-Astrodynamics?cHash=6349bf79113ed67885ab83701e652639;
2. https://www.tib.eu/en/search/id/TIBKAT:521489873/A-collection-of-technical-papers-AIAA-AAS-Astrodynamics?cHash=cd433c128bcf43589e005c9f781fdb37;
3. https://www.tib.eu/en/search/id/TIBKAT:521489903/A-collection-of-technical-papers-AIAA-AAS-Astrodynamics?cHash=cb1d3358a90d8b8658ca3bb358e99946.

See https://www.tib.eu/en/search/id/BLCP:CN065746138/Repeat-Ground-Track-Lunar-Orbits-in-the-Full-Potential?cHash=f086b9f29f698e135b8d8d7aa257c363.

This print version has page numbers 1935–1954 in that citation, whereas the ARC PDF is numbered 1–20…

If we want to cite the print book we should do the same for the other paper from the AIAA/AAS Astrodynamics Specialist Conference, [Jon12], in https://www.tib.eu/en/search/id/TIBKAT:769080677/AIAA-AAS-Astrodynamics-Specialist-Conference-2012?cHash=190c9597dcb0a79c120396334a3fb587—note that for that one the POD publisher is [Curran](http://www.proceedings.com/15920.html), and the AIAA would need to move from `publisher` to `organization`.

documentation/Geopotential.tex
eggrobin: This was because of the `nocite{*}` in another `refsegment` (whose bibliography is not printed) in the same `refsection` (the margin shifted to accomodate the widest possible label).

This was a way to ensure globally consistent labels with automated disambiguation; now that we have `userd` we no longer need that.
Note that this means that global uniqueness of the references will be checked only when compiling `bibliography` (but we should probably make the CI do that anyway).

Done.

documentation/bibliography.bib
eggrobin: Added the print metadata to both entries.

documentation/Geopotential.tex
pleroy: There are still differences in the way that the URLs get folded in `[PL10]` and `[Wes17]` (and the latter is still ugly).

eggrobin: Ah, this is because `eprint` now starts a line per the global style, and our current settings forbid mid-word linebreaking in a URL. I will try to tame the `hyperref` options.

eggrobin: Done, there’s a `biblatex` setting for that (found in https://tex.stackexchange.com/a/134281).

PR #2923, opened 2021-03-15
Uniformize layout

eggrobin:
Changes:
- Factor out the `\textdisplay` from the `\title` (note that this has the effect of setting the author and date in the display font as well).
- Make everything `twopage`.
- Use `extdash` and `\Endash` for linebreaking of Runge–Kutta[–Nyström] in the bibliography, instead of a mix of unbreakable `--` and literal `–`, and breaking but verbose `--\allowbreak `. The package `extdash` provides `\--` which would be less verbose, but which appears ot fail to break when used in the bibliography.
- Give the bibliography a `\title` instead of a `\section` title.


documentation/Composition methods.tex
```diff
@@ -25,7 +21,7 @@ \section{Differential equations.}
 \quad\parbox{.4\linewidth}{with $\exp h\vA$ and $\exp h\vB$ known and
 $\commutator{\vB}{\commutator{\vB}{\commutator{\vB}{\vA}}}=\nullvec$;}
 \label{general}\\
-&\quad\text{the above equation, with $\exp h\vA = \Identity+h\vA$,
+\span\parbox{.7\linewidth}{the above equation, with $\exp h\vA = \Identity+h\vA$,
```
pleroy, line 24: Please let's not go around tweaking pixels.  The original code was just fine AFAIAC.  Not sure why it had to be changed, but I don't want to have to write crud like that.

documentation/bibliography.bib
```diff
@@ -17,7 +17,7 @@ @article{AlonsoMalloCanoMoreta2006
   keywords     = {Runge–Kutta–Nyström methods;Stability},
   number       = {1-2},
   pages        = {120--131},
-  title        = {Stability of Runge–Kutta–Nyström methods},
+  title        = {Stability of Runge\Endash Kutta\Endash Nyström methods},
```
pleroy, line 20: It's hard to believe that these changes are an improvement.  We should strive to keep the source comprehensible.  In particular, I am pretty sure that the next time I reference RKN methods I will forget about the silly `\Endash`.

documentation/bibliography.tex
```diff
@@ -1,7 +1,11 @@
-\documentclass[10pt, a4paper]{basestyle}
+\documentclass[10pt, a4paper, twoside]{basestyle}
```
pleroy, line 1: I completely hate the new format.  It is way too narrow and makes the bibliography entries hard to read with multiple word breaks and many lines per entry (even ignoring the XVIIth century oddities, we have numerous entries that span 6-7 lines).

I'd be happy to keep this document one-sided.  Alternatively, organize the two-sidedness so that the body of the text keeps the same width and the `[XXX]` handles are hanging on the left.

documentation/Composition methods.tex
eggrobin: > The original code was just fine

The original code *was* doing pixel-tweaking: this line is not aligned with the equalities in the `align`, and `&\quad` was manually spacing it by one em with respect to the alignment point (whereas `\span` makes it unaligned).

Similarly `\text` relied on the text fitting on one line, whereas `\parbox` allows linebreaking.

documentation/bibliography.tex
eggrobin: > organize the two-sidedness so that the body of the text keeps the same width

Agreed, and done. The wide margins from `basestyle` make sense for the other documents where they can be used for figures or review annotations, but the bibliography doesn’t have figures and it is hard to imagine we will be doing much scribbling in its margins; this is made worse by the bibliography effectively having its own margin for the labels.

PR #2926, opened 2021-03-20
Smarter override of version check

pleroy:
This will ensure that users don't disable it forever.  Example:
```
principia_override_version_check {
  version = 0.24.2
  version = 1.11.2
  version = 2.0.666
}
```
Tested in game with 1.11.2.

### DIFF
--- a/ksp_plugin_adapter/ksp_plugin_adapter.cs
+++ b/ksp_plugin_adapter/ksp_plugin_adapter.cs
@@ -49,17 +49,17 @@ public partial class PrincipiaPluginAdapter : ScenarioModule,
     Vectors = 31,
   };
 
-  private const string principia_serialized_plugin_ = "serialized_plugin";
-  private const string principia_initial_state_config_name_ =
+  private const string principia_serialized_plugin = "serialized_plugin";
+  private const string principia_initial_state_config_name =
       "principia_initial_state";
-  private const string principia_gravity_model_config_name_ =
+  private const string principia_gravity_model_config_name =
       "principia_gravity_model";
-  private const string principia_numerics_blueprint_config_name_ =
+  private const string principia_numerics_blueprint_config_name =
       "principia_numerics_blueprint";
-  private const string principia_override_version_check_config_name_ =
+  private const string principia_override_version_check_config_name =
       "principia_override_version_check";
-  private const string principia_flags_ = "principia_flags";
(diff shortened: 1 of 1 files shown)


ksp_plugin_adapter/ksp_plugin_adapter.cs
```diff
@@ -305,10 +305,23 @@ class PartCentredForceHolder {
...
+      bool fail_version_check = true;
+      ConfigNode override_version_check =
+          GameDatabase.Instance.GetAtMostOneNode(
+              principia_override_version_check_config_name);
+      if (override_version_check != null) {
+        foreach (var version in override_version_check.GetValues("version")) {
+          if (version ==
+              $@"{Versioning.version_major}.{Versioning.version_minor}.{
+                Versioning.Revision}") {
+            fail_version_check = false;
```
eggrobin, line 319: This feels a bit too double-negativeish, and in both cases we failed the version check; we are just making it non-fatal.
`version_check_overriden`?

eggrobin (changes requested):

ksp_plugin_adapter/ksp_plugin_adapter.cs
```diff
@@ -49,17 +49,17 @@ private enum UnityLayers {
...
-  private const string principia_numerics_blueprint_config_name_ =
+  private const string principia_numerics_blueprint_config_name =
       "principia_numerics_blueprint";
-  private const string principia_override_version_check_config_name_ =
+  private const string principia_override_version_check_config_name =
       "principia_override_version_check";
-  private const string principia_flags_ = "principia_flags";
-  private const string principia_draw_styles_config_name_ =
+  private const string principia_flags = "principia_flags";
+  private const string principia_draw_styles_config_name =
```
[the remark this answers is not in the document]
pleroy: Reverted the renaming.

I am not sure that I buy the argument in #1554, actually.  The problem there was that the name `Δt` was just too short for something declared at the top level (and it's not even clear from the name what was its purpose).  Hard to believe that we'd run into hiding issues with `principia_override_version_check_config_name` .

Fine with doing that in a separate PR, but please let's come up with a naming convention for constants.  Search for `private const` and you'll find every.possible.style.

ksp_plugin_adapter/ksp_plugin_adapter.cs

eggrobin (approved):

PR #2932, opened 2021-04-02
Change the way that checkpoints are serialized to allow multiple checkpoints

pleroy:
Checkpoints now go in their separate message instead of in the parent message.

#2400, take two.

### DIFF
--- a/physics/checkpointer.hpp
+++ b/physics/checkpointer.hpp
@@ -26,23 +26,16 @@ using quantities::Time;
 // Instead, this class creates checkpoints that encapsulate all information
 // needed to reconstruct the timeline after a give point in time.  When
 // serializing a timeline, the pairs (time, data) are written up to the oldest
-// checkpoint, followed by the checkpoint itself.  When deserializing, the
-// timeline may be reconstructed as needed based on the checkpoint.
+// checkpoint, followed by the checkpoints themselves.  When deserializing, the
+// timeline may be reconstructed as needed based on the checkpoints.
 // Checkpoints must be created at regular intervals to ensure that the timeline
 // may be reconstructed fast enough.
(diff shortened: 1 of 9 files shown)


physics/checkpointer.hpp
```diff
@@ -25,62 +26,68 @@ using quantities::Time;
...
-                       Message const& message) EXCLUDES(lock_);
+  // recently, as specified by |max_time_between_checkpoints|.  Returns true iff
+  // a new checkpoint was created.
+  bool WriteToCheckpointIfNeeded(Instant const& t,
+                                 Time const& max_time_between_checkpoints)
+      EXCLUDES(lock_);
+
+  // Caller the |Reader| passed at construction to reconstruct an object using
+  // the oldest checkpoint.  Returns true iff this object contains at least one
+  // checkpoint.
```
eggrobin, line 68: Could we return an erroneous status, and check IsFinite(oldest_checkpoint()) beforehand if needed? I find calls to `ReadFromOldestCheckpoint` that could actually do nothing a bit unreadable.

pleroy: It really should not happen, at the time of deserialization, that we don't have a checkpoint (it can happen e.g., when the checkpointer was just created) because we create one when serializing.  So I am changing this method to `ReadFromOldestCheckpointOrDie`.  This requires changes to tests that were actually not putting the object in a consistent state.

eggrobin: I don’t think we should call our functions `OrDie`; doing anything with the `OldestCheckpoint` requires that the set of checkpoints be nonempty, so this is clearly a requirement from the name, and we check our requirements everywhere without calling everything `OrDie`.

eggrobin (approved):

PR #2935, opened 2021-04-03
Remove the shutdown bit from the prognosticator parameters

pleroy:
It's no longer needed now that we have stoppable threads.


ksp_plugin/vessel.cpp
```diff
@@ -583,10 +569,6 @@ Status Vessel::RepeatedlyFlowPrognostication() {
       std::swap(prognosticator_parameters, prognosticator_parameters_);
     }
 
-    if (prognosticator_parameters->shutdown) {
```
eggrobin, line 586: Replace by a RETURN_IF_STOPPED?

eggrobin (approved):

PR #2936, opened 2021-04-03
Downsample the analyser trajectories

### LINKED ISSUE #2919: The orbit analysers in the flight plan have unreasonable memory usage for long flight plans
I have an old RP-0 save using Gallai and just switched to use the latest Goldbach version. Now once I get the save loaded and even if I sit there doing nothing, the game crashes within one minute. This is consistent and occured everytime. I pulled out resource manager and can visually confirm a steady growth of commit memory until it exceeds my RAM limit (32GB). The below screenshot was taken just after game crashed. 

image

I read the version notes and seems that now it calculate the trajectory on scene change, so I wonder this might be the issue? I had about 12 spacecrafts in-flight, most of them being LEO satellites with no flight plan. But there were a couple deep space probes which had long flight plans going 2000+days ahead and utilizing multiple gravity assists. I guess they might be the root cause, but at least I was able to create these flight plans with the older Gallai version.

Logs (the journal alone is 1GB before compression)
https://public-data-share-3258311398.s3.amazonaws.com/crash_logs_0314.zip

eggrobin:
Fix #2919.

### DIFF
--- a/ksp_plugin/integrators.hpp
+++ b/ksp_plugin/integrators.hpp
@@ -15,8 +15,8 @@ using quantities::si::Metre;
 using quantities::si::Milli;
 
 // Parameters for downsamplng after fixed-step integration.
-constexpr std::int64_t max_dense_intervals = 10'000;
-constexpr Length downsampling_tolerance = 10 * Metre;
+constexpr std::int64_t MaxDenseIntervals = 10'000;
+constexpr Length DownsamplingTolerance = 10 * Metre;
 
 // Factories for parameters used to control integration.
 Ephemeris<Barycentric>::AccuracyParameters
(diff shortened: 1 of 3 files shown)


ksp_plugin/integrators.hpp
```diff
@@ -14,6 +14,10 @@ using quantities::Length;
 using quantities::si::Metre;
 using quantities::si::Milli;
 
+// Parameters for downsamplng after fixed-step integration.
+constexpr std::int64_t max_dense_intervals = 10'000;
```
pleroy, line 18: I think we agreed that public constants like this should be in `CamelCase`.

al2me6:
Thank you for fixing this so quickly!

PR #2943, opened 2021-04-05
Support for prepending a continuous trajectory to another one

pleroy:
#2400 take two.

### DIFF
--- a/physics/continuous_trajectory_body.hpp
+++ b/physics/continuous_trajectory_body.hpp
@@ -160,7 +160,14 @@ Status ContinuousTrajectory<Frame>::Prepend(
     // conceivable that small differences would occur when moving a save from
     // one machine to another.  Because of this, we cannot check that the
     // trajectories are "continuous" at the junction, whatever that means.
-    if (*first_time_ == trajectory.polynomials_.back().t_max) {
+    if (*first_time_ != trajectory.polynomials_.back().t_max) {
+      // Lenient, but not too much: we don't want the last polynomial of
+      // |trajectory| to start *after* our |first_time_|.
+      CHECK_LT(*trajectory.first_time_, *first_time_);
+      CHECK(trajectory.polynomials_.size() == 1 ||
+            trajectory.polynomials_[trajectory.polynomials_.size() - 2].t_max <
+                *first_time_);
+
       status =
           Status(Error::OUT_OF_RANGE,
                  absl::StrCat(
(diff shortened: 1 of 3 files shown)


physics/continuous_trajectory_body.hpp
```diff
@@ -131,6 +131,66 @@ Status ContinuousTrajectory<Frame>::Append(
   return status;
 }
 
+template<typename Frame>
+Status ContinuousTrajectory<Frame>::Prepend(
+    ContinuousTrajectory&& trajectory) {
```
eggrobin, line 136: Rename the parameter to `prefix`.

physics/continuous_trajectory_body.hpp
```diff
@@ -131,6 +131,66 @@ Status ContinuousTrajectory<Frame>::Append(
...
+    degree_age_ = trajectory.degree_age_;
+    polynomials_ = std::move(trajectory.polynomials_);
+    last_accessed_polynomial_ = trajectory.last_accessed_polynomial_;
+    first_time_ = trajectory.first_time_;
+    last_points_ = trajectory.last_points_;
+  } else {
+    // The polynomials must be aligned, but we are lenient because it's
+    // conceivable that small differences would occur when moving a save from
+    // one machine to another.  Because of this, we cannot check that the
+    // trajectories are "continuous" at the junction, whatever that means.
```
eggrobin, line 162: Should we really expect such a thing? The treatment of time with fixed steps (and continuous trajectories, which are based on fixed steps) uses only basic operations (and as we have remarked elsewhere, is often exact in practice).

Continuity in space is hopeless without correctly-rounded elementary functions, but that is another can of worms.

I would favour requiring exactness for simplicity and easier spotting of bugs, and then testing across machines with different numerical libraries (we have those).

eggrobin (changes requested):

physics/continuous_trajectory_test.cpp
```diff
@@ -171,6 +170,7 @@ class ContinuousTrajectoryTest : public testing::Test {
       // additions) because it results in a bit of jitter in the intervals,
       // which matters for continuity.
       Instant ti = time + (i + 1) * step;
+      LOG(ERROR)<<ti<<" "<<i;
```
eggrobin, line 173: Remove the debug trace, I think.

eggrobin (approved):

PR #2947, opened 2021-04-16
Fix a missed change from Goldbach to Grassmann

scimas:
Tiny change in the README.md


pleroy:
Thanks for the fix.  It's not the first time that I miss something when updating the README.

PR #2950, opened 2021-04-18
Full implementation of the reanimator

pleroy:
Tested (to some extent) in game.

#2400 take two.

### DIFF
--- a/physics/continuous_trajectory_test.cpp
+++ b/physics/continuous_trajectory_test.cpp
@@ -728,8 +728,9 @@ TEST_F(ContinuousTrajectoryTest, Serialization) {
   EXPECT_GE(100, checkpoint.degree_age());
   EXPECT_EQ(4, checkpoint.last_point_size());
 
-  auto const trajectory_read =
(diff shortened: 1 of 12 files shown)


physics/continuous_trajectory_test.cpp
```diff
@@ -729,7 +729,7 @@ TEST_F(ContinuousTrajectoryTest, Serialization) {
   EXPECT_EQ(4, checkpoint.last_point_size());
 
   auto const trajectory_read =
-      ContinuousTrajectory<World>::ReadFromMessage(message);
+      ContinuousTrajectory<World>::ReadFromMessage(InfiniteFuture, message);
```
eggrobin, line 732: Name the parameter whose value is `InfiniteFuture`; same below.

eggrobin (approved):

physics/continuous_trajectory_test.cpp
pleroy: Done, also in `ephemeris_test.cpp`.

PR #2978, opened 2021-05-08
Do not flow a flight plan segment that is before the Ephemeris t_min

### LINKED ISSUE #2975: [Grassmanian branch] Crash when loading a pre-existing save
Log attached; journal at https://drive.google.com/file/d/1-7io-pSEbs2zQm_aQpYJqD-xwXgB-ypI/view?usp=sharing.
[INFO.20210507-185204.26904.log](https://github.com/mockingbirdnest/Principia/files/6443193/INFO.20210507-185204.26904.log)

pleroy:
Request reanimation instead.

Fix #2975.

### DIFF
--- a/ksp_plugin/flight_plan.cpp
+++ b/ksp_plugin/flight_plan.cpp
@@ -345,20 +345,29 @@ Status FlightPlan::BurnSegment(
     not_null<DiscreteTrajectory<Barycentric>*> const segment) {
   Instant const final_time = manœuvre.final_time();
   if (manœuvre.initial_time() < final_time) {
-    if (manœuvre.is_inertially_fixed()) {
-      return ephemeris_->FlowWithAdaptiveStep(
-                             segment,
-                             manœuvre.InertialIntrinsicAcceleration(),
-                             final_time,
-                             adaptive_step_parameters_,
-                             max_ephemeris_steps_per_frame);
+    Instant const starting_time = segment->back().time;
+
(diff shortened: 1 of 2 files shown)


ksp_plugin/flight_plan.cpp
```diff
@@ -368,12 +377,21 @@ Status FlightPlan::BurnSegment(
...
+  if (ephemeris_->t_min() <= starting_time) {
+    return ephemeris_->FlowWithAdaptiveStep(
+        segment,
+        Ephemeris<Barycentric>::NoIntrinsicAcceleration,
+        desired_final_time,
+        adaptive_step_parameters_,
+        max_ephemeris_steps_per_frame);
+  } else {
+    ephemeris_->RequestReanimation(starting_time);
+    return Status::OK;
```
eggrobin, line 393: This is going to break all the invariants: we have a segment that has not been flowed, but we say it is OK. If we eventually hit a segment whose starting time is after t_min, we are going to try integrating that, and assuming this doesn’t crash, we will have a missing segment followed by a normal segment, which is not how anomaly works in the flight plan.
This should return an erroneous status so that the entire flight plan becomes anomalous.
This should probably also have a unit test.

eggrobin (changes requested):

ksp_plugin/flight_plan.cpp
pleroy: Done, PTAL.

PR #2989, opened 2021-05-18
Rewrite PolynomialInMonomialBasis to avoid specializations

pleroy:
This fixes an old code smell that dated back to VS2015.


eggrobin (approved):
Now that we can have that for all types, we should bake the origin into the Taylor polynomials in the elliptic functions. https://github.com/mockingbirdnest/Principia/blob/9d71cf0f117faf084fa948d7127c2c321672e871/numerics/elliptic_integrals.cpp#L2048-L2069

PR #2996, opened 2021-05-25
Additive operators for converting vector-valued and affine valued polynomials

pleroy:
For some mysterious reason one of the four causes MSVC to go down the path of insanity.  We'll live with the rest.

### DIFF
--- a/numerics/polynomial.hpp
+++ b/numerics/polynomial.hpp
@@ -332,6 +332,8 @@ operator*(
     PolynomialInMonomialBasis<RValue, Argument, rdegree_, Evaluator> const&
         right);
 
+// Additive operators polynomial ± constant.
+
 #if PRINCIPIA_COMPILER_MSVC_HANDLES_POLYNOMIAL_OPERATORS
 template<typename Value, typename Argument, int ldegree_,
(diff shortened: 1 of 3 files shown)


numerics/polynomial.hpp
```diff
@@ -201,6 +207,30 @@ class PolynomialInMonomialBasis : public Polynomial<Value_, Argument_> {
   friend operator*(
       PolynomialInMonomialBasis<L, A, l, E> const& left,
       PolynomialInMonomialBasis<R, A, r, E> const& right);
+#if PRINCIPIA_COMPILER_MSVC_HANDLES_POLYNOMIAL_OPERATORS
```
eggrobin, line 210: Add a blank line and a comment
```C++
// Operators polynomial ± constant.
```
So that the semantic of the `V` or `A` is clear.

eggrobin (approved):

numerics/polynomial.hpp
pleroy: Done, but not here, in the *real* declarations.  These are the friend declarations, and we keep them as compact/cryptic a possible.

PR #3007, opened 2021-05-29
Use hardware FMA

eggrobin:
…or [Voigt-Kampff/Dekker](https://twitter.com/rygorous/status/1021953667941859334) if we don’t have that. Avoid `std::fma`.


numerics/fma.hpp
```diff
@@ -0,0 +1,46 @@
...
+// The functions in this file unconditionally wrap the appropriate intrinsics.
+// The caller may only use them if |UseHardwareFMA| is true.
+#if PRINCIPIA_USE_FMA_IF_AVAILABLE
+inline bool const UseHardwareFMA = HasCPUFeatures(CPUFeatureFlags::FMA);
+#else
+inline bool const UseHardwareFMA = false;
+#endif
+
+// ⟦ab + c⟧.
+inline double FusedMultiplyAdd(double a, double b, double c);
```
pleroy, line 24: Could we have tests for these functions to verify that we are calling the intrinsics correctly?

numerics/double_precision.hpp
```diff
@@ -127,6 +135,7 @@ using internal_double_precision::Mod2π;
 using internal_double_precision::TwoDifference;
 using internal_double_precision::TwoProduct;
 using internal_double_precision::TwoSum;
+using internal_double_precision::VeltkampDekkerProduct;
```
pleroy, line 138: Why would any client ever want to use this function?  Shouldn't it be purely internal?

eggrobin: See the comments:
https://github.com/mockingbirdnest/Principia/blob/19073a6492441fedf411b2107e58edcbc6b105f8/numerics/double_precision.hpp#L55-L57

pleroy:
You've added references to a test file in the projects, but not the test file itself.

pleroy:
Also, why don't we change `cpuid.cpp` into `cpuid_body.hpp`?  That would avoid adding the `cpp` files to all the projects, and it seems to make sense to inline these operations anyway.

eggrobin:
It doesn’t really make sense to inline that ; we don’t want to do CPUID calls every time we use something, instead we do them once and refer to the static result (here UseHardwareFMA).

numerics/fma_test.cpp
```diff
@@ -0,0 +1,36 @@
...
+#include "gmock/gmock.h"
+#include "gtest/gtest.h"
+#include "testing_utilities/almost_equals.hpp"
+
+namespace principia {
+namespace numerics {
+
+using base::CPUFeatureFlags;
+using base::HasCPUFeatures;
+using ::testing::Test;
```
pleroy, line 14: We generally don't do a `using` of this class, we use the full name when declaring the fixture.  Also, we put the `using` of symbols that start with `::` at the end of the list.

PR #3010, opened 2021-05-30
-mfma

eggrobin:


eggrobin:
We need to revert this.
Setting -mfma causes clang to emit VEX-encoded instructions (`vaddsd` rather than `addsd`, &c) for every floating-point operation everywhere, which means that this change drops pre-Sandy Bridge support on Linux and Mac.

MSVC is happy emitting VEX-encoded FMA intrinsics while keeping SSE2 everywhere else, so we can move forward with FMA on Windows if we want, but doing so on the other platforms would mean producing a binary that requires AVX (at which point we can also fit `R3Element` in a single register, which is nice; we probably want that on Windows too; but if we want to retain support for users on Westmere and eralier that means two DLLs, and some code to load the correct one).

PR #3028, opened 2021-06-08
Fix the build

eggrobin:
Clang doesn’t want to sweep a defaulted template parameter under the rug when using the template as a template template parameter.

Drive-by change: remove an obsolete comment (the evaluator structs are really needed, if only to group Evaluate and EvaluateDerivative).

### DIFF
--- a/numerics/polynomial_evaluators.hpp
+++ b/numerics/polynomial_evaluators.hpp
@@ -11,13 +11,30 @@ namespace internal_polynomial_evaluators {
 using quantities::Derivative;
 using quantities::Square;
 
-// The structs below are only needed because the language won't let us have (1)
-// a partial specialization of a function (2) an explicit specialization of a
-// member function template in an unspecialized class template.  Sigh.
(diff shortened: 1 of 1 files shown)


numerics/polynomial_evaluators.hpp
```diff
@@ -11,21 +11,38 @@ namespace internal_polynomial_evaluators {
...
 
-// The structs below are only needed because the language won't let us have (1)
-// a partial specialization of a function (2) an explicit specialization of a
-// member function template in an unspecialized class template.  Sigh.
+template<typename Value, typename Argument, int degree, bool allow_fma>
+struct EstrinEvaluator;
+template<typename Value, typename Argument, int degree, bool allow_fma>
+struct HornerEvaluator;
+
+namespace exported {
```
pleroy, line 19: Would it work to close `internal_polynomial_evaluators` here and reopen it near line 33?  That might avoid all the references to `exported`.

PR #3032, opened 2021-06-12
Add a predicate to check if a vessel is in a state that allows collapsing its trajectory

pleroy:
#2400 for vessels.

### DIFF
--- a/ksp_plugin/vessel.cpp
+++ b/ksp_plugin/vessel.cpp
@@ -654,11 +654,9 @@ bool Vessel::IsCollapsible() const {
   for (const auto& [_, part] : parts_) {
     // We expect parts to be piled up.
     CHECK(part->is_piled_up());
-    // Not collapsible if any part has a force or a torque applied to it.
+    // Not collapsible if any part has a force applied to it (but a torque is
+    // fine).
     if (part->intrinsic_force() != Vector<Force, Barycentric>{}) {
-      return false;
-    }
-    if (part->intrinsic_torque() != Bivector<Torque, Barycentric>{}) {
       return false;
     }
     parts.insert(part.get());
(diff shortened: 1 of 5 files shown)


ksp_plugin/vessel.cpp
```diff
@@ -645,6 +648,44 @@ void Vessel::AttachPrediction(
...
+  for (const auto& [_, part] : parts_) {
+    // Not collapsible if any part has a force or a torque applied to it.
+    if (part->intrinsic_force() != Vector<Force, Barycentric>{}) {
+      return false;
+    }
+    if (part->intrinsic_torque() != Bivector<Torque, Barycentric>{}) {
+      return false;
+    }
+    // Not collapsible if any part is not piled up (not even sure when that can
+    // happen).
```
eggrobin, line 663: Maybe we should try to understand what is going in our own code, instead of treating it like KSP?
This should either be removed or be a CHECK; not being piled up is something transient when building pile ups, and we should definitely not be considering collapsibility when the pile ups are being determined.

ksp_plugin/vessel.cpp
```diff
@@ -645,6 +648,44 @@ void Vessel::AttachPrediction(
...
+    parts.insert(part.get());
+    if (containing_pile_up == nullptr) {
+      containing_pile_up = part->containing_pile_up();
+    } else {
+      // All the parts should be in the same pile-up.
+      CHECK_EQ(containing_pile_up, part->containing_pile_up());
+    }
+  }
+  // Not collapsible if this vessel is not in a pile-up (e.g., because it has no
+  // part).
```
eggrobin, line 676: This should also be a CHECK.

eggrobin (changes requested):

eggrobin (approved):

PR #3034, opened 2021-06-19
Legible instants

eggrobin:
This will be used in `operator<<(std::ostream&, Instant)` in a future PR.


astronomy/time_scales_body.hpp
```diff
@@ -12,18 +12,19 @@
 #include "geometry/named_quantities.hpp"
 #include "glog/logging.h"
 #include "quantities/si.hpp"
+#include "numerics/double_precision.hpp"
```
pleroy, line 15: Alphabetize.

astronomy/time_scales_body.hpp
```diff
@@ -570,6 +571,44 @@ inline Instant Parse北斗Time(std::string const& s) {
...
+  // compute it exactly and round by hand.
+  DoublePrecision<quantities::Time> const time_of_day =
+      TwoDifference(t, beginning_of_day);
+  quantities::Time const time_of_day_rounded_down =
+      time_of_day.error >= 0 * Second
+          ? time_of_day.value
+          : std::bit_cast<double>(
+                std::bit_cast<std::uint64_t>(time_of_day.value / Second) - 1) *
+                Second;
+  int second_of_day = std::floor(time_of_day_rounded_down / Second);
```
pleroy, line 604: `const`

astronomy/time_scales_body.hpp
```diff
@@ -570,6 +571,44 @@ inline Instant Parse北斗Time(std::string const& s) {
...
+    mjd_most_significant_digits /= 10;
+  }
+  return JulianDate::MJD(mjd, digits, /*fractional_digit_count=*/0).CalendarDay();
+}
+
+inline DateTime TTSecond(Instant const& t) {
+  auto const date = TTDay(t);
+  Instant const beginning_of_day = DateTimeAsTT(DateTime::BeginningOfDay(date));
+  // Close to J2000, Sterbenz’s lemma can fail to apply to this subtraction.  We
+  // compute it exactly and round by hand.
```
pleroy, line 595: This is an unfortunate property.  I wish we had taken, e.g., the origin of MJD as our zero.  (Nobody flies rockets in 1858.)

astronomy/time_scales_body.hpp
```diff
@@ -570,6 +571,44 @@ inline Instant Parse北斗Time(std::string const& s) {
...
+  DoublePrecision<quantities::Time> const time_of_day =
+      TwoDifference(t, beginning_of_day);
+  quantities::Time const time_of_day_rounded_down =
+      time_of_day.error >= 0 * Second
+          ? time_of_day.value
+          : std::bit_cast<double>(
```
pleroy, line 601: Comment these bit-casting.

astronomy/time_scales_body.hpp
```diff
@@ -570,6 +571,44 @@ inline Instant Parse北斗Time(std::string const& s) {
...
+                std::bit_cast<std::uint64_t>(time_of_day.value / Second) - 1) *
+                Second;
+  int second_of_day = std::floor(time_of_day_rounded_down / Second);
+  return DateTime(date,
+                  date_time::Time(/*hour=*/second_of_day / 3600,
```
pleroy, line 606: Should we have constants for 3600 and 60?

astronomy/date_time_body.hpp
```diff
@@ -1038,6 +1091,10 @@ constexpr DateTime operator""_DateTime(char const* const str,
                                     size - (index_of_T + 1)));
 }
 
+inline std::ostream& operator<<(std::ostream& out, DateTime const& date_time) {
+  return out << date_time.date() << "T" << date_time.time();
```
pleroy, line 1095: Where is this operator (and the previous ones) tested.?

astronomy/time_scales_body.hpp
eggrobin: JD0 would be the standard approach if the origin is really annoying, but I don’t think this really is a problem except for this kind of correct rounding or round-tripping trickery, and having the origin in the middle of the timespan we care about means we have better resolution, which seems like it should be more broadly useful.

astronomy/date_time_body.hpp
eggrobin: In #3040.

numerics/double_precision_body.hpp
```diff
@@ -251,7 +252,7 @@ DoublePrecision<Product<T, U>> TwoProduct(T const& a, U const& b) {
 }
 
 template<typename T, typename U>
-FORCE_INLINE(inline)
+constexpr FORCE_INLINE(inline)
```
pleroy, line 255: Should be written `FORCE_INLINE(constexpr)`.

PR #3040, opened 2021-06-27
Meeus

eggrobin:
Split off from #3034.
Use Meeus’s algorithms instead of attempting to hand-roll our own date arithmetic.
Add support for the (proleptic) Julian calendar; dates before 1583 must have a prefix `J` or `G` to indicate the calendar used.
Add support for a leading sign on years, to represent years before 0000 (1 BC).
Add an `operator<<`; check that its output parses correctly for all dates from JD0.5 to the Gröbner release.


astronomy/date_time_body.hpp
```diff
@@ -547,9 +526,21 @@ class DateParser final {
   int const w_index_;
 };
 
-constexpr Date DateParser::Parse(char const* const str,
-                                 std::size_t const size) {
-  return ReadToEnd(str, size).ToDate();
+constexpr Date DateParser::Parse(char const* str,
+                                 std::size_t size) {
```
pleroy, line 530: `const`

astronomy/date_time_body.hpp
```diff
@@ -547,9 +526,21 @@ class DateParser final {
   int const w_index_;
 };
 
-constexpr Date DateParser::Parse(char const* const str,
-                                 std::size_t const size) {
-  return ReadToEnd(str, size).ToDate();
+constexpr Date DateParser::Parse(char const* str,
```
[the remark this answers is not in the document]
eggrobin: No, it is incremented.

astronomy/date_time_body.hpp
eggrobin: No, it is decremented.

PR #3042, opened 2021-07-04
Make FitHermiteSpline interruptible

### LINKED ISSUE #3033: Game stutters when vessel change state from landed to in flight or vice versa
When a vessel hits the ground or takes off there is a noticeable stutter as Principia takes or releases control of the vessel.

This can easily be seen by bouncing a vessel against the floor repeatedly (Physx will often do this for you).
There will be a noticeable pause on each bounce; switching hack gravity on will result in smooth frame rates suggesting that it is Principia causing the issue.

eggrobin:
Fix  #3033.

### DIFF
--- a/numerics/fit_hermite_spline.hpp
+++ b/numerics/fit_hermite_spline.hpp
@@ -2,6 +2,7 @@
 
 #include <list>
 
+#include "absl/status/statusor.h"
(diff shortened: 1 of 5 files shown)


eggrobin:
Verified in-game.

PR #3048, opened 2021-07-05
Legible output for Instant

eggrobin:
Diff on top of #3050: https://github.com/mockingbirdnest/Principia/compare/9ce778f...eggrobin:tchouk-tchouk?expand=1.


geometry/instant_output.cpp
```diff
@@ -0,0 +1,49 @@
...
+  // sign of 0 for J2000 itself, and the second after J2000 has fractions of a
+  // second that are too small to be reasonably printed in fixed format.
+  if (t >= "JD0.5"_TT && t < "9999-12-31T24:00:00"_TT &&
+      !(t > J2000 - 1 * Second && t < J2000 + 1 * Second)) {
+    auto const tt_second = TTSecond(t);
+    Instant const start_of_second = DateTimeAsTT(tt_second);
+    // This subtraction is exact by Sterbenz’s lemma.
+    Time const remainder = t - start_of_second;
+    // |remainder| being the result of a subtraction of numbers greater than or
+    // equal to 1, it is a multiple of 2u ≈ 2⏨-16; 16 fractional decimal digits
```
pleroy, line 32: I have no idea what `2⏨-16` means.

geometry/instant_output_test.cpp
```diff
@@ -0,0 +1,141 @@
...
+using astronomy::operator""_UT1;
+using astronomy::operator""_UTC;
+using quantities::Time;
+using quantities::si::Second;
+using ::testing::Eq;
+
+class InstantOutputTest : public ::testing::Test {
+ protected:
+  Instant JustAfter(Instant const t) {
+    return J2000 + std::bit_cast<double>(
```
pleroy, line 27: Could this use `std::nextafter`?

geometry/instant_output_test.cpp
```diff
@@ -0,0 +1,141 @@
...
+using astronomy::operator""_TT;
+using astronomy::operator""_UT1;
+using astronomy::operator""_UTC;
+using quantities::Time;
+using quantities::si::Second;
+using ::testing::Eq;
+
+class InstantOutputTest : public ::testing::Test {
+ protected:
+  Instant JustAfter(Instant const t) {
```
pleroy, line 26: Document this function.  Also, `const&`.

geometry/instant_output_test.cpp
```diff
@@ -0,0 +1,141 @@
...
+    return J2000 + std::bit_cast<double>(
+                       std::bit_cast<std::uint64_t>((t - J2000) / Second) +
+                       Sign(t - J2000)) *
+                       Second;
+  }
+  Instant JustBefore(Instant const t) {
```
pleroy, line 32: Blank line before and same comments.

geometry/instant_output_test.cpp
```diff
@@ -0,0 +1,141 @@
...
+};
+
+TEST_F(InstantOutputTest, UniversalTime) {
+  // Note that since we are roughly twelve hours from J2000, we expect
+  // log₁₀(12 h / 1 s), between four and five, of the printed digits to be
+  // unrepresentable: our granularity is about seven picoseconds. This is
+  // obvious in the case of UTC, which is offset from TT(TAI) by an integer
+  // number of milliseconds at any time.
+  EXPECT_THAT((std::stringstream() << "2000-01-01T00:00:00"_UTC).str(),
+              Eq("2000-01-01T00:01:04,1840000000011059 (TT)"));
```
pleroy, line 49: The trailing `11059` is not significant, right?  It might change if the implementation changes.  I would use a regexp and leave these digits unspecified.

geometry/instant_output.cpp
eggrobin: https://en.wikipedia.org/wiki/Scientific_notation#:~:text=Decimal%20Exponent%20Symbol%20is%20part,Algol%2060%20and%20Algol%2068.

geometry/instant_output_test.cpp
eggrobin: Yes; I need to implement a constexpr IEEE 754 nextUp/nextDown to fix the clang build anyway.

geometry/instant_output.cpp
pleroy: Could we please use a comprehensible notation and not some prehistoric convention? `2E-16`, `2.0E-16`, `2×10⁻¹⁶`.

geometry/instant_output_test.cpp
eggrobin: The implementation (of operator""_UTC) should be correctly-rounded (it is not, see the TODO below, this was found by looking at those « insignificant » digits, so they test _something_).
Barring changes of the value being formatted, these digits must not change: the fractional part (stored in `remainder`) is computed exactly, and we print 17 digits of it; if we cannot trust the binary-to-decimal code on seventeen digits we cannot ever hope that it will survive round-tripping.

geometry/instant_output_test.cpp
eggrobin: OBE.

PR #3049, opened 2021-07-05
NextUp & NextDown

eggrobin:
We don’t have `std::bit_cast` in our *nix builds yet, and `absl::bit_cast` is not constexpr so it does not help.
Implement an IEEE 754 `nextUp` and `nextDown` the slow and ugly way, using arithmetic.
We will probably want to keep those even when we get C++20 everywhere, since `std::nextafter` does not have IEEE 754 semantics and is not constexpr anyway. We will want to reimplement them with `std::bit_cast` though.


astronomy/time_scales_body.hpp
```diff
@@ -12,6 +12,7 @@
 #include "geometry/named_quantities.hpp"
 #include "glog/logging.h"
 #include "numerics/double_precision.hpp"
+#include "quantities/elementary_functions.hpp"
```
pleroy, line 15: Why this include?  Shouldn't you get `next.hpp`?

numerics/log_b.hpp
```diff
@@ -0,0 +1,39 @@
...
+
+namespace principia {
+namespace numerics {
+
+// A constexpr implementation of the IEEE 754:2008 logB function.
+// Uses sourceFormat as logBFormat, which makes it easy to cleanly handle NaN,
+// infinity, and 0.
+template<typename SourceFormat,
+         typename = std::enable_if_t<std::is_floating_point_v<SourceFormat>>>
+constexpr SourceFormat LogB(SourceFormat const x) {
```
pleroy, line 13: Please separate the `_body.hpp` file from the `.hpp` file as we normally do.

numerics/log_b.hpp
```diff
@@ -0,0 +1,39 @@
+#pragma once
+
+#include <limits>
+
+namespace principia {
+namespace numerics {
+
+// A constexpr implementation of the IEEE 754:2008 logB function.
+// Uses sourceFormat as logBFormat, which makes it easy to cleanly handle NaN,
```
pleroy, line 9: Weird camelCase (2x).

numerics/log_b.hpp
```diff
@@ -0,0 +1,39 @@
...
+  }
+  if (x == 0) {
+    // Signal the divideByZero exception (by dividing by -0).
+    return 1 / -abs_x;
+  }
+  SourceFormat scaled_abs_x = abs_x;
+  SourceFormat result = 0;
+  while (scaled_abs_x < 1) {
+    scaled_abs_x *= std::numeric_limits<SourceFormat>::radix;
+    result -= 1;
```
pleroy, line 29: `--result`

numerics/log_b.hpp
```diff
@@ -0,0 +1,39 @@
...
+  }
+  while (scaled_abs_x >= std::numeric_limits<SourceFormat>::radix) {
+    scaled_abs_x /= std::numeric_limits<SourceFormat>::radix;
+    result += 1;
```
pleroy, line 33: `++result`

numerics/next.hpp
```diff
@@ -0,0 +1,65 @@
...
+#include "numerics/scale_b.hpp"
+
+namespace principia {
+namespace numerics {
+
+// A constexpr implementation of the IEEE 754:2008 nextUp and nextDown
+// functions.
+template<typename SourceFormat,
+         typename = std::enable_if_t<std::is_floating_point_v<SourceFormat>>>
+constexpr SourceFormat NextUp(SourceFormat const x) {
```
pleroy, line 16: Separate the body.

numerics/next.hpp
```diff
@@ -0,0 +1,65 @@
...
+    return -0.0;
+  }
+  if (x == 0) {
+    // Note that this is independent of the sign of 0, as specified by
+    // IEEE 754:2008, 5.3.1.  This differs from C++ std::nextafter, which maps
+    // -0 to +0.
+    return std::numeric_limits<SourceFormat>::denorm_min();
+  }
+  if (x == std::numeric_limits<SourceFormat>::infinity()) {
+    // TODO(egg): Should we be signalling the overflow exception?
```
pleroy, line 29: I don't think we ever use the floating-point exceptions...

numerics/next_test.cpp
```diff
@@ -0,0 +1,80 @@
...
+using geometry::Sign;
+using ::testing::Eq;
+
+class NextTest : public ::testing::Test {};
+
+TEST_F(NextTest, ExtendedRealLine) {
+  static_assert(NextUp(-std::numeric_limits<double>::infinity()) ==
+                -0x1.FFFFFFFFFFFFFp+1023);
+  static_assert(NextUp(-0x1.FFFFFFFFFFFFFp+1023) == -0x1.FFFFFFFFFFFFEp+1023);
+  // [...]
```
pleroy, line 21: No idea what this comment is trying to tell me.

pleroy (commented):
For `scale_b` too, we should separate the body.

numerics/log_b.hpp
eggrobin: IEEE 754 uses that case, and I am referring to names in there.

numerics/next_test.cpp
eggrobin: A few numbers were omitted in the progression from -∞ to +∞.

numerics/next.hpp
eggrobin: No, but when it’s trivial to get them right we might as well…

astronomy/time_scales_body.hpp
eggrobin: `numerics/next.hpp` only has the one for `double`.

PR #3059, opened 2021-07-11
Change downsampling to work for non-root trajectories

pleroy:
It turns out that this was as simple as removing checks.  Presumably someone was lazy and didn't feel like writing a test...

A modest contribution to #2400.


pleroy:
retest this please

PR #3063, opened 2021-07-14
Add a mechanism to exclude forks at serialization time

pleroy:
Previously, only the forks listed in the `forks` parameter were included; they were restored on deserialization.

This doesn't work too well if there is a large number of fork to serialize, because listing them explicitly requires cumbersome bookkeeping.  And really, if the history has 1000 forks, we don't want to know about them.

So I am changing the default so that all the forks rooted at a trajectory are serialized, except those listed in the `excluded` parameter.  The ones that we are interested in, i.e., the ones that we want to restore on deserialization, are listed in `tracked`.

### DIFF
--- a/serialization/physics.proto
+++ b/serialization/physics.proto
@@ -79,7 +79,7 @@ message ContinuousTrajectory {
 }
 
 message DiscreteTrajectory {
-  // A marker to indicate that a fork doesn't have it's position tracked.
+  // A marker to indicate that a fork doesn't have its position tracked.
   // Added in Groethendieck/Haar.
   enum ForkPosition {
     MISSING_FORK_POSITION = -1;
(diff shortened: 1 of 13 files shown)


serialization/physics.proto
```diff
@@ -79,6 +79,11 @@ message ContinuousTrajectory {
 }
 
 message DiscreteTrajectory {
+  // A marker to indicate that a fork doesn't have it's position tracked.
+  // Added in Groethendieck/Haas.
```
eggrobin, line 83: Haar, I don’t think we will have a [Haas](https://en.wikipedia.org/wiki/Ruth_Haas) release.

eggrobin (approved):

Issue #3072, opened 2021-07-18
Gather interesting trajectories and use them in tests


eggrobin:
Both to exercise save compatibility to avoid a repeat of #3039, and to have real data with which to test the physics and astronomy libraries.

Issue #3039, opened 2021-06-25
Green & Grassmann OK but Gröbner crashes with old RetroRendezvous save 


Growflavor:
Green & Grassmann successfully load in KSP 1.11.2 the [old save](https://www.dropbox.com/s/d9vbdw0wba0jkxv/rendezvous.zip) at:

https://forum.kerbalspaceprogram.com/index.php?/topic/161390-principia-challenge-munar-retrograde-rendezvous-122/

However I get the following crash when trying to open the same save in KSP 1.11.2 with Gröbner: 

[Principia_ERROR_20210625_223123_15380.zip](https://github.com/mockingbirdnest/Principia/files/6718957/Principia_ERROR_20210625_223123_15380.zip)

`E0625 22:31:23.531883 22048 interface.cpp:691] Principia version 2021061011-Gröbner-0-g05ed8f86927035b5820d611536d711b3176db3d4 built on 2021-06-05T16:53:53Z by Microsoft Visual C++ version 192930037 for Windows x86-64

F0625 22:31:40.448882 22072 continuous_trajectory_body.hpp:489] Check failed: (continuous_trajectory->checkpointer_->ReadFromCheckpointAtOrBefore( desired_t_min)) == ::absl::OkStatus() (NOT_FOUND: No checkpoint vs. OK)`

PR #3076, opened 2021-07-25
Test passage in the compatibility paths, add a save by Reach

eggrobin:
The new save is interesting in the sense of #3072. It is 294 kiB, much smaller than the existing one.


ksp_plugin_test/plugin_compatibility_test.cpp
```diff
@@ -113,17 +124,109 @@ class PluginCompatibilityTest : public testing::Test {
...
+        infnity->flight_plan().GetManœuvre(i).Δv().Norm());
+  }
+  // The flight plan only covers the inner solar system (this is probably
+  // because of #3035).
+  // It also differs from https://youtu.be/7BDxZV7UD9I?t=439.
+  // TODO(egg): Compute the flybys and figure out what exactly is going on in
+  // this flight plan.
+  EXPECT_THAT(manœuvre_ignition_tt_seconds_and_Δvs,
+              ElementsAre(Pair("1970-08-14T09:34:49"_DateTime,
+                               +3.80488671073918022e+03 * (Metre / Second)),
```
pleroy, line 196: No `+`.

ksp_plugin_test/plugin_compatibility_test.cpp
```diff
@@ -70,20 +87,19 @@ class PluginCompatibilityTest : public testing::Test {
...
     PullSerializer* serializer = nullptr;
     char const* b64 = nullptr;
 
     LOG(ERROR) << "Serialization starting";
     for (;;) {
-      b64 = principia__SerializePlugin(plugin.get(),
-                                       &serializer,
-                                       preferred_compressor,
-                                       preferred_encoder);
+      b64 = principia__SerializePlugin(
```
pleroy, line 101: Unreformat.

ksp_plugin_test/plugin_compatibility_test.cpp
```diff
@@ -96,13 +114,8 @@ class PluginCompatibilityTest : public testing::Test {
...
 
-  void CheckSaveCompatibility(std::filesystem::path const& filename,
-                              std::string_view const compressor,
-                              std::string_view const encoder) {
-    // Read a plugin from the given file.
-    auto plugin1 = ReadPluginFromFile(filename, compressor, encoder);
-
-    // Write that plugin back to another file with the preferred format.
+  static void WriteAndReadBack(not_null<std::unique_ptr<Plugin const>> plugin1) {
+    // Write the plugin to new file with the preferred format.
```
pleroy, line 118: ... to *a* new file ...

ksp_plugin_test/plugin_compatibility_test.cpp
```diff
@@ -18,14 +20,29 @@
...
+using ::testing::AllOf;
+using ::testing::ElementsAre;
+using ::testing::Eq;
+using ::testing::HasSubstr;
+using ::testing::IsFalse;
+using ::testing::IsTrue;
+using ::testing::Not;
 using ::testing::NotNull;
+using ::testing::Pair;
+using ::testing::internal::CaptureStderr;
```
pleroy, line 44: Let's not access the internals, please.  It's going to break whenever Google feel like changing their implementation.

The same effect could be achieved in a portable manner with `EXPECT_EXIT`.

ksp_plugin_test/plugin_compatibility_test.cpp
```diff
@@ -113,17 +126,109 @@ class PluginCompatibilityTest : public testing::Test {
...
+  EXPECT_THAT(test->name(), Eq("TEST"));
+  EXPECT_THAT(TTSecond(test->psychohistory().front().time),
+              Eq("1970-08-14T08:03:18"_DateTime));
+  EXPECT_THAT(TTSecond(test->psychohistory().back().time),
+              Eq("1970-08-14T08:47:05"_DateTime));
+  EXPECT_THAT(test->has_flight_plan(), IsFalse());
+
+  auto const infnity =
+      plugin->GetVessel("29142a79-7acd-47a9-a34d-f9f2a8e1b4ed");
+  EXPECT_THAT(infnity->name(), Eq("IFNITY-5.2"));
```
pleroy, line 178: So the variable is called `infnity` and the vessel `IFNITY`.  Hint: this is not the same.

ksp_plugin_test/plugin_compatibility_test.cpp
```diff
@@ -113,17 +126,109 @@ class PluginCompatibilityTest : public testing::Test {
...
+  EXPECT_THAT(TTSecond(infnity->psychohistory().front().time),
+              Eq("1970-08-14T08:03:46"_DateTime));
+  EXPECT_THAT(TTSecond(infnity->psychohistory().back().time),
+              Eq("1970-08-14T08:47:05"_DateTime));
+  EXPECT_THAT(infnity->has_flight_plan(), IsTrue());
```
pleroy, line 183: `EXPECT_TRUE`, or rather `ASSERT_TRUE` (the next statement would die if this is false.)

ksp_plugin_test/plugin_compatibility_test.cpp
```diff
@@ -3,12 +3,14 @@
...
 #include <vector>
 
+#include "astronomy/time_scales.hpp"
 #include "base/file.hpp"
 #include "base/not_null.hpp"
 #include "base/pull_serializer.hpp"
 #include "base/push_deserializer.hpp"
 #include "glog/logging.h"
 #include "gmock/gmock.h"
+#include "google/protobuf/io/coded_stream.h"
```
pleroy, line 13: Why is this needed?

ksp_plugin_test/plugin_compatibility_test.cpp
```diff
@@ -113,17 +126,109 @@ class PluginCompatibilityTest : public testing::Test {
                                       preferred_compressor,
                                       preferred_encoder);
   }
+
+  static void CheckSaveCompatibility(std::filesystem::path const& filename,
+                              std::string_view const compressor,
```
pleroy, line 131: Bad indent.

ksp_plugin_test/plugin_compatibility_test.cpp
```diff
@@ -113,17 +126,109 @@ class PluginCompatibilityTest : public testing::Test {
...
   CheckSaveCompatibility(
-      SOLUTION_DIR / "ksp_plugin_test" / "pre_cohen.proto.hex",
+      SOLUTION_DIR / "ksp_plugin_test" / "saves" / "3039.proto.hex",
       /*compressor=*/"",
       /*decoder=*/"hexadecimal");
+  EXPECT_THAT(
+      GetCapturedStderr(),
+      AllOf(
+          HasSubstr(
+              "pre-Cohen ContinuousTrajectory"),  // Regression test for #3039.
```
pleroy, line 154: Not convinced that the comments here add value.

eggrobin: At least this one seems useful, since this this is what #3039 was about, and it’s not particularly obvious from the phrase « pre-Cohen ContinuousTrajectory ».

ksp_plugin_test/plugin_compatibility_test.cpp
eggrobin: Not really, because then it needs to be a death test and this makes it tricky (and unreadable) to have more complicated tests like the Reach one.

The proper way to do this would be to take advantage of the fact that we have decided to diverge our glog and have it be instrumentable, instead of reaching all the way to standard error (either directly or even more circuitously via EXPECT_EXIT).

eggrobin: A `LogSink` does the job.

ksp_plugin_test/plugin_compatibility_test.cpp
```diff
@@ -47,6 +44,40 @@ using ::testing::internal::GetCapturedStderr;
 const char preferred_compressor[] = "gipfeli";
 const char preferred_encoder[] = "base64";
 
+class StringLogSink : google::LogSink {
+ public:
+  StringLogSink(google::LogSeverity const minimal_severity)
```
pleroy, line 49: The `minimal_severity` seems unused.

PR #3080, opened 2021-08-04
Do not test Reach’s save in debug

eggrobin:
While it is slower than maccollo’s, it is also much slower to load.


pleroy:
retest this please

PR #3092, opened 2021-08-12
Translate technical terms in the new frame selector UI

eggrobin:
Also tweak the English terms a bit (JSO rather than JCSO, consistent with, e.g., [Juno terminology](https://lasp.colorado.edu/home/mop/files/2015/02/CoOrd_systems12.pdf)).

Use acronyms (ECI, ECEF, JSO, etc.) for both the selector buttons and the navball text when patterns for them are provided; otherwise, use a generic short description on the buttons (Inertial frame, Body-fixed frame, Orbital frame), and the full names on the navball.

Use the localized celestial names (this matters in RSS).

The reference frame descriptions and tooltips have not yet been translated.


ksp_plugin_adapter/ksp_plugin_adapter.csproj
```diff
@@ -121,6 +120,7 @@
     <Compile Include="ksp_plugin_adapter.cs" />
     <Compile Include="interface.cs" />
     <Compile Include="loader.cs" />
+    <Compile Include="localizer_extensions.cs" />
```
pleroy, line 123: Localisation

pleroy (approved):

ksp_plugin_adapter/ksp_plugin_adapter.csproj
eggrobin: -ization even (oxendict).

PR #3098, opened 2021-08-16
Update zh-cn localization and add strings for the new frame selector

al2me6:
Notable changes:

* Added localized strings for the new plotting frame selector.
* Localized the word 'Pin'.
* Rewrote the update reminders to be more faithful to the en-us version. It now mentions lunation numbers, but uses a written-out explanation ("The \<n\>th new moon, as defined by Meeus's _Astronomical Algorithms_, has arrived...") as the concept does not appear to exist in Chinese.
* Corrected the patched conics toggle to actually use the term 'patched conics'. The original instead used 'original orbit plotting,' in the sense of being in the unmodded game.
* Changed a number of flight analyser error messages so that they no longer refer to the craft not being inside an SoI. (_N.B._ the term Egg originally suggested for 'gravitationally bound' was '重力束缚‘. However, this term is not attested on CNKI (other than one usage in an odd philosophical context), so I chose the more commonly attested (in the context of globular clusters and black holes) '引力束缚‘ instead.)
* Removed the term 'simulate' (模拟) from the three burn mode (engine/RCS/instant impulse) selectors.
* Changed 'predicted/planned' from '已预测/已计划' to '预测的/计划的'. I parse the former as '\<unspecified subject\> has predicted/planned,' which seems awkward. The new translation is the adjectival form.
* Removed a number of (IMHO) extraneous parentheses around body names.
* Made various smaller edits where the wording felt extraneous/awkward/inaccurate to me.

These should be correct to the best of my knowledge. However, my Chinese is somewhat rusty and I do not have much formal mathematics/physics education in Chinese. Thus, a second pair of eyes might be advisable.

### DIFF
--- a/ksp_plugin_adapter/localization/zh-cn.cfg
+++ b/ksp_plugin_adapter/localization/zh-cn.cfg
@@ -54,10 +54,10 @@ Localization {
     #Principia_ReferenceFrameSelector_Description_Heading = <<1>>参考系：
     // <<1>> and <<2>> are the names of the relevant bodies, with articles if appropriate, <<2>> being the parent of <<1>>.
     // For the target frame, <<1>> is the name of the target vessel.
-    #Principia_ReferenceFrameSelector_Description_Target = 固定目标载具，<<1>>。\n以它绕<<2>>的轨道为水平面。\n这个参考系跟随轨道旋转，以保持<<2>>的方向不变。
-    #Principia_ReferenceFrameSelector_Description_BodyCentredNonRotating = 固定<<1>>的质心。\n以它的赤道为水平面。\n这个参考系不旋转。
-    #Principia_ReferenceFrameSelector_Description_BodyCentredParentDirection = 固定<<1>>的质心。\n以它绕<<2>>的轨道为水平面。\n这个参考系跟随轨道旋转，以保持<<2>>的方向不变。
-    #Principia_ReferenceFrameSelector_Description_BodySurface = 固定<<1>>。\n以它的赤道为水平面。\n这个参考系与<<1>>共同旋转。  // <<1>>centred body name
+    #Principia_ReferenceFrameSelector_Description_Target = 目标载具<<1>>已固定，\n水平面为绕着<<2>>的轨道。\n这个参考系跟随轨道旋转，以保持<<2>>的方向不变。
+    #Principia_ReferenceFrameSelector_Description_BodyCentredNonRotating = <<1>>的质心已固定，\n水平面为它的赤道。\n这个参考系不旋转。
+    #Principia_ReferenceFrameSelector_Description_BodyCentredParentDirection = <<1>>的质心已固定，\n水平面为绕着<<2>>的轨道。\n这个参考系跟随轨道旋转，以保持<<2>>的方向不变。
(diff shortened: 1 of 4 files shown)


pleroy:

ksp_plugin_adapter/localization/zh-cn.cfg
```diff
@@ -186,7 +188,7 @@ Localization {
     #Principia_FlightPlan_StatusMessage_ChangeFlightPlan = 请更换轨道规划
     #Principia_FlightPlan_StatusMessage_FailedError = 计算失败，出现错误 <<1>> 错误信息 \"<<2>>\"  // <<1>> status_.error <<2>> status_.message
     #Principia_FlightPlan_StatusMessage_TimeOut = 在<<1>>之后  // <<1>> Timeout
```
eggrobin, line 190: Change this to <<1>> (it is weird that it is its own string anyway), we’ll deal with wrapping it below (and I might just get rid of it later on).

ksp_plugin_adapter/localization/zh-cn.cfg
```diff
@@ -186,7 +188,7 @@ Localization {
...
-    #Principia_FlightPlan_StatusMessage_MaxSteps = 积分器已达到最大步数<<1>>  // <<1>> Timeout
+    #Principia_FlightPlan_StatusMessage_MaxSteps = 积分器已<<1>>达到最大步数  // <<1>> Timeout
```
eggrobin, line 191: The existing string talks about the time taken by the computation, not the time reached by the calculation; Zethar recommends:
> 积分器无法在最大步数以内算出+<<1>>时刻

ksp_plugin_adapter/localization/zh-cn.cfg
```diff
@@ -186,7 +188,7 @@ Localization {
...
     #Principia_FlightPlan_StatusMessage_MaxSegment = 请增加“最大迭代步长”或避免与天体碰撞
     #Principia_FlightPlan_StatusMessage_Singularity = 积分器遇到奇点（可能是天体的中心）<<1>>  // <<1>> timeout
```
eggrobin, line 193: The existing string talks about the time taken by the computation, not the time of the singularity; Zethar recommends:
> 积分器在算+<<1>>时刻时遇到了奇点

eggrobin (commented):
Some suggestions from @zaikarion.

ksp_plugin_adapter/localization/zh-cn.cfg
al2me6: As in, just `#Principia_FlightPlan_StatusMessage_FailedError = <<1>>`?

eggrobin: Wrong line, but
```
#Principia_FlightPlan_StatusMessage_TimeOut = <<1>>
```
(That is what goes into the <<1>> of the other two strings below)

al2me6: Ah. That  `<<1>>` I figured, thought you were referring to something else for whatever reason.

eggrobin (approved):

PR #3101, opened 2021-08-17
Improve translation malleability

eggrobin:
No functional change; preparation for fr support.

Use Lingoona articles (^d) rather than an explicit « the ».

Merge some strings to allow further reordering.

Every time the English version uses the name of a celestial without an article if it has one, e.g. « Mun orbit », « Mun periapsis », or an initial, pass instead three parameters 3k+1, 3k+2, 3k+3 which are « Mun^d », « Mun », and « M ».

Introduce a navball name string and drop the ad hoc code for that in Chinese.

### DIFF
--- a/ksp_plugin_adapter/localization/celestial_strings.cfg
+++ b/ksp_plugin_adapter/localization/celestial_strings.cfg
@@ -21,17 +21,17 @@
     // orbit, sending the latitude of perijove further from the equator.”
     // in [KJL08].  We follow this usage.
 
-    // <<4>> is the sequence of properties, ending with a space.
-    #Principia_OrbitAnalyser_OrbitDescription(Sun).en-us = <<4>>heliocentric orbit
-    #Principia_OrbitAnalyser_OrbitDescription(Moon).en-us = <<4>>lunar orbit
-    #Principia_OrbitAnalyser_OrbitDescription_Synchronous(Earth).en-us = <<4>>geosynchronous orbit
+    // <<3>> is the sequence of properties, ending with a space.
+    #Principia_OrbitAnalyser_OrbitDescription(Sun).en-us = <<3>>heliocentric orbit
+    #Principia_OrbitAnalyser_OrbitDescription(Moon).en-us = <<3>>lunar orbit
+    #Principia_OrbitAnalyser_OrbitDescription_Synchronous(Earth).en-us = <<3>>geosynchronous orbit
     #Principia_OrbitAnalyser_OrbitDescription_Stationary(Earth).en-us = geostationary orbit
 
     #Principia_ReferenceFrameSelector_Name_BodyCentredNonRotating(Sun).en-us = Heliocentric Inertial
     #Principia_ReferenceFrameSelector_Abbreviation_BodyCentredNonRotating(Sun).en-us = HCI
     #Principia_ReferenceFrameSelector_Name_BodyCentredParentDirection(Earth,Sun).en-us = Geocentric Solar Ecliptic
     #Principia_ReferenceFrameSelector_Abbreviation_BodyCentredParentDirection(Earth,Sun).en-us = GSE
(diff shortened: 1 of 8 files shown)


ksp_plugin_adapter/localization/celestial_strings.cfg
```diff
@@ -20,15 +20,15 @@
     // “However, the argument of periapsis increases approximately 0.9° per
     // orbit, sending the latitude of perijove further from the equator.”
     // in [KJL08].  We follow this usage.
-    #Principia_OrbitAnalyser_OrbitDescription_Primary(Sun).en-us = heliocentric
-    #Principia_OrbitAnalyser_OrbitDescription_Primary(Moon).en-us = lunar
-    #Principia_OrbitAnalyser_OrbitDescription_SynchronousPrimary(Earth).en-us = geosynchronous
-    #Principia_OrbitAnalyser_OrbitDescription_StationaryPrimary(Earth).en-us = geostationary
+    #Principia_OrbitAnalyser_OrbitDescription(Sun).en-us = <<4>>heliocentric orbit
```
pleroy, line 23: I have no idea what `<<4>>` is.  Maybe a comment?

ksp_plugin_adapter/localization_extensions.cs
```diff
@@ -28,23 +29,51 @@ internal static class L10N {
...
-    return (Localizer.CurrentLanguage == english_us_ &&
-            body.displayName.StartsWith("The ")) ? body.name : body.displayName;
+  public static string StandaloneName(this CelestialBody celestial) {
+    return Standalone(celestial.NameWithoutArticle());
+  }
+
+  private static bool StartsWithCapitalizedDefiniteArticle(string s) {
+    return (Localizer.CurrentLanguage == english_us_ && s.StartsWith("The ")) ||
+           (Localizer.CurrentLanguage == french_ &&
+            (s.StartsWith("La ") || s.StartsWith("Le ")));
```
pleroy, line 39: What about `L'` or `Les`?

ksp_plugin_adapter/localization_extensions.cs
```diff
@@ -28,23 +29,51 @@ internal static class L10N {
...
 
   public static string NameWithArticle(this CelestialBody body) {
-    // TODO(egg): should we apped ^d instead of prepending "the"?  Certainly
-    // we would need to do that for other languages.
-    return (Localizer.CurrentLanguage == english_us_ &&
-            body.displayName.StartsWith("The ")) ? "the " + body.name
-                                                 : body.displayName;
+    return StartsWithCapitalizedDefiniteArticle(body.displayName)
+        ? body.displayName.Contains('^')
+              ? body.displayName.Split(new[]{' '}, 2)[1] + "d"
```
pleroy, line 68: Extract the `Split` into a local variable, no point in repeating a complicated expression.

ksp_plugin_adapter/localization_extensions.cs
eggrobin: This is there to work around stock because we must work around stock ; other mods should use `ornithorhynque^d`, not `L’ornithorhynque` (and should we handle APOSTROPHE or only RIGHT SINGLE QUOTATION MARK? if we go that way it never ends, we might want to handle the lower case as well…).

Stock only has `La Mune` and `Le Soleil`.

eggrobin:
On second thought after poking at lingoona some more, I think instead of `Sun^d`, `Moon^d`, `Mars` we should have `Sun, `Moon`, `Mars^N`, and then we can choose whether we want an article if appropriate using `<<A:1>>` (the Moon, Mars) or <<1>> (Moon, Mars).
Closing for now, will reopen if that works.

ksp_plugin_adapter/localization_extensions.cs
```diff
@@ -54,27 +54,28 @@ internal static class L10N {
...
-        ? LingoonaQualify(
-              body.displayName.Split(new[]{' '}, 2)[1],
-              LingoonaQualifiers(body.displayName).Replace("d", ""))
-        : body.displayName;
-  }
-
-  public static string NameWithArticle(this CelestialBody body) {
-    if (!StartsWithCapitalizedDefiniteArticle(body.displayName)) {
-      return body.displayName;
+  public static string Name(this CelestialBody body) {
```
pleroy, line 57: Please comment this function, I shouldn't have to read the code to know what it does.

PR #3105, opened 2021-08-21
Allow tracking the root trajectory (not only forks) during serialization

pleroy:
This makes it possible to track a pointer that may denote either the root or some fork.

You guessed it, #2400.

### DIFF
--- a/serialization/physics.proto
+++ b/serialization/physics.proto
@@ -97,11 +97,12 @@ message DiscreteTrajectory {
     required Point instant = 1;
     required Pair degrees_of_freedom = 2;
   }
-  message Child {
+  // All the trajectories forked at a given time.
+  message Brood {
     required Point fork_time = 1;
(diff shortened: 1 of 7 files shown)


serialization/physics.proto
```diff
@@ -97,13 +97,16 @@ message DiscreteTrajectory {
     required Point instant = 1;
     required Pair degrees_of_freedom = 2;
   }
-  message Litter {
+  message Child {
     required Point fork_time = 1;
     repeated DiscreteTrajectory trajectories = 2;
```
eggrobin, line 102: If this message is called Child this field cannot be repeated; revert this renaming.

eggrobin (approved):

serialization/physics.proto
pleroy: Brood.

pleroy:
retest this please

PR #3113, opened 2021-08-23
Лидов–古在

### LINKED ISSUE #1119: Лидов–古在 mechanism test

eggrobin:
This test showcases the Лидов–古在 mechanism, and thus fixes #1119.
The relevant elements vary as follows:
butcher_e_i
The eccentricity–inclination exchange is evidenced by the following graph:
butcher_c1
The other conserved quantity (which relates all three elements) is also constant: the argument of the periapsis governs that exchange.
butcher_c2


astronomy/butcher_mercury_orbiter_1119.hpp
```diff
@@ -0,0 +1,45 @@
+﻿#pragma once
```
pleroy, line 1: Just call this file `memory_orbiter.hpp`.  No point in recording who provided it or what bug it was associated with.  (You can have that in the comments, of course.)

astronomy/лидов_古在_test.cpp
```diff
@@ -0,0 +1,165 @@
...
+
+  SolarSystem<ICRS> solar_system_1950_;
+  not_null<std::unique_ptr<Ephemeris<ICRS>>> ephemeris_;
+  MassiveBody const& mercury_;
+  BodyCentredNonRotatingDynamicFrame<ICRS, MercuryCentredInertial>
+      mercury_frame_;
+};
+
+#if !_DEBUG
+TEST_F(Лидов古在Test, MercuryOrbiter1) {
```
pleroy, line 78: Why `1` in the test name?

astronomy/butcher_mercury_orbiter_1119.hpp
```diff
@@ -0,0 +1,45 @@
...
+using physics::DegreesOfFreedom;
+using quantities::si::Metre;
+using quantities::si::Second;
+
+// State of the spacecraft Mercury Orbiter 1 at the start of September, from a
+// save by Butcher given in issue #1119.  This is used both from a plugin
+// compatibility test which reads that save, and from an astronomical test which
+// looks for the Лидов–古在 mechanism in the orbit.
+
+constexpr Instant MercuryOrbiter1119InitialTime =
```
pleroy, line 24: Remove `1119` from the identifiers.

astronomy/лидов_古在_test.cpp
```diff
@@ -0,0 +1,165 @@
...
+using quantities::Pow;
+using quantities::Sin;
+using quantities::si::Metre;
+using quantities::si::Milli;
+using quantities::si::Minute;
+using quantities::si::Radian;
+using quantities::si::Kilo;
+using quantities::si::Second;
+using testing_utilities::operator""_⑴;
+using quantities::si::Degree;
```
pleroy, line 45: Alphabetize.

astronomy/лидов_古在_test.cpp
```diff
@@ -0,0 +1,165 @@
+﻿
+#include <memory>
+
+#include "astronomy/butcher_mercury_orbiter_1119.hpp"
+#include "astronomy/orbital_elements.hpp"
+#include "astronomy/frames.hpp"
+#include "gmock/gmock.h"
+#include "gtest/gtest.h"
+#include "ksp_plugin/integrators.hpp"
```
pleroy, line 9: `astronomy` should not depend on `ksp_plugin`, that's going to introduce horrible compilation dependencies.  Just replicate the parameters as needed.  (It's more robust too: if we decide to change a bit the way that we downsample, we should not break this test.)

astronomy/лидов_古在_test.cpp
```diff
@@ -0,0 +1,165 @@
...
+  auto const instance = ephemeris_->NewInstance(
+      {&icrs_trajectory},
+      Ephemeris<ICRS>::NoIntrinsicAccelerations,
+      {SymmetricLinearMultistepIntegrator<Quinlan1999Order8A, Position<ICRS>>(),
+       /*step=*/10 * Second});
+
+  for (int year = 1967; year < 1980; ++year) {
+    Instant const t = DateTimeAsTT(date_time::DateTime::BeginningOfDay(
+        date_time::Date::Ordinal(year, 1)));
+    LOG(ERROR) << "Flowing to " << t;
```
pleroy, line 93: Not really an `ERROR`.  Should we log to `INFO` and make sure that `INFO` goes to `stderr`?  Same at line 96.

astronomy/лидов_古在_test.cpp
```diff
@@ -0,0 +1,165 @@
...
+    auto const status = ephemeris_->FlowWithFixedStep(t, *instance);
+    if (!status.ok()) {
```
pleroy, line 95: Should this be a `CHECK`?

astronomy/лидов_古在_test.cpp
```diff
@@ -0,0 +1,165 @@
...
+
+  DiscreteTrajectory<MercuryCentredInertial> mercury_centred_trajectory;
+  for (auto const& [t, dof] : icrs_trajectory) {
+    mercury_centred_trajectory.Append(t,
+                                      mercury_frame_.ToThisFrameAtTime(t)(dof));
+  }
+
+  OrbitalElements const elements = OrbitalElements::ForTrajectory(
+      mercury_centred_trajectory, mercury_, MasslessBody{}).value();
+  mathematica::Logger logger(SOLUTION_DIR / "mathematica" / "1119.generated.wl",
```
pleroy, line 109: `лидов_古在.generated.wl`.

astronomy/лидов_古在_test.cpp
```diff
@@ -0,0 +1,165 @@
...
+    c₂.Include((1 - ε) * (2.0 / 5.0 - sin²_i * sin²_ω));
+    logger.Append("t", elements.time, mathematica::ExpressIn(Second));
+    logger.Append("a", elements.semimajor_axis, mathematica::ExpressIn(Metre));
+    logger.Append("e", elements.eccentricity);
+    logger.Append("i", elements.inclination, mathematica::ExpressIn(Radian));
+    logger.Append(R"(\[Omega])",
+                  elements.argument_of_periapsis,
+                  mathematica::ExpressIn(Radian));
+  }
+  // The elements e, i, and ω all vary quite a lot.
```
pleroy, line 130: The numbers would be easier to interpret if, instead of `min` and `max` (or in addition), you looked at the relative error between `measure` and `midpoint`.  It's like 75% for `e` and 17% for `c₁`.

astronomy/лидов_古在_test.cpp
eggrobin: No, we do encounter an erroneous status (the orbiter runs into the ground at the end of 1971).

astronomy/лидов_古在_test.cpp
eggrobin: I don’t think the relative errors are meaningful here, since the point (c₁, c₂) lies within a bounded region within [0, 1] × [-3/5, 2/5] (depicted in [Лид61, p. 29], [LC62, p. 745]).

pleroy: How do you demonstrate that something is constant if you have lots of measurements with noise?  By looking at its statistical elements.  If you are not happy with a cheesy relative error, please do a proper statistical analysis.

eggrobin: As discussed, what we care about is the absolute scale of the variations; they are not really noise, but they are very small compared to the size of the region given in Рис. 1.

eggrobin:
retest this please

PR #3122, opened 2021-09-05
A mechanism for prepending a Forkable to another Forkable

pleroy:
A very boring part of #2400.

### DIFF
--- a/physics/forkable.hpp
+++ b/physics/forkable.hpp
@@ -134,7 +134,11 @@ class Forkable {
   // This trajectory is still a valid pointer, but it may now be owned by the
   // parent of T.  The only pointer that the client might assume to own is
   // this->root().
-  void Prepend(Tr4jectory&& trajectory);
+  //TODO(phl):comments
+  static not_null<std::unique_ptr<Tr4jectory>> Prepend(
+    not_null<std::unique_ptr<Tr4jectory>> prefix_root,
+    Tr4jectory const& prefix,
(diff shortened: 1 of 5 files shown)


physics/forkable.hpp
```diff
@@ -118,6 +118,24 @@ class Forkable {
...
+  // 3. Attach the forks (strictly) before t to this trajectory.
+  // 4. If T is a root, delete the entire tree rooted at T.
+  // 5. Otherwise, attach this trajectory as a fork of the parent of T instead
+  //    of T and delete T.
+  // When returning from this method, it is generally not safe to touch any part
+  // of the tree containing |trajectory|, unless the client knows a lot about
+  // the structure of the various trajectories (that's the case in tests).
+  // This trajectory is still a valid pointer, but it may now be owned by the
+  // parent of T.  The only pointer that the client might assume to own is
+  // this->root().
```
eggrobin, line 136: Let’s make this a static method that takes both roots by `unique_ptr` as well as a pointer to `trajectory` (whose root must be the prepended one), and returns whichever `unique_ptr` remains; taking ownership of `this` is a bad idea (and likewise for `trajectory->root()`), and having a `unique_ptr` to the non-const root of `trajectory` will ensure that the `const_cast` in the implementation does not introduce UB.

eggrobin (changes requested):

eggrobin (approved):

Issue #3136, opened 2021-10-02
Forkable must die


pleroy:
[Design](https://docs.google.com/document/d/1POE2fXMCMndNPzhHelkrmKAS1TvFsZ7pBOZk_9sqtJs/edit?usp=sharing).

PR #3137, opened 2021-10-03
Headers for the new-style DiscreteTrajectory

pleroy:
First step of #3136.

### DIFF
--- a/physics/discrete_trajectory2.hpp
+++ b/physics/discrete_trajectory2.hpp
@@ -9,6 +9,7 @@
 #include "physics/degrees_of_freedom.hpp"
 #include "physics/discrete_trajectory_iterator.hpp"
 #include "physics/discrete_trajectory_segment_iterator.hpp"
+#include "physics/discrete_trajectory_segment_range.hpp"
 #include "physics/discrete_trajectory_types.hpp"
 #include "physics/trajectory.hpp"
 #include "serialization/physics.pb.h"
@@ -30,8 +31,13 @@ using physics::DegreesOfFreedom;
 template<typename Frame>
 class DiscreteTrajectory2 : public Trajectory<Frame> {
  public:
-  using Iterator = DiscreteTrajectoryIterator<Frame>;
+  using iterator = DiscreteTrajectoryIterator<Frame>;
+  using reverse_iterator = std::reverse_iterator<iterator>;
   using SegmentIterator = DiscreteTrajectorySegmentIterator<Frame>;
+  using ReverseSegmentIterator = std::reverse_iterator<SegmentIterator>;
+  using SegmentRange = DiscreteTrajectorySegmentRange<SegmentIterator>;
+  using ReverseSegmentRange =
+      DiscreteTrajectorySegmentRange<ReverseSegmentIterator>;
 
   DiscreteTrajectory2() = default;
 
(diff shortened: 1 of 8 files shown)


physics/discrete_trajectory2.hpp
```diff
@@ -0,0 +1,110 @@
...
+using base::not_null;
+using geometry::Instant;
+using geometry::Position;
+using geometry::Velocity;
+using physics::DegreesOfFreedom;
+
+template<typename Frame>
+class DiscreteTrajectory2 : public Trajectory<Frame> {
+ public:
+  using Iterator = DiscreteTrajectoryIterator<Frame>;
```
eggrobin, line 33: let’s call the result of `begin` `iterator`; the API is meant to look like a [_Container_](https://en.cppreference.com/w/cpp/named_req/Container) here (or a [`std::span`](https://en.cppreference.com/w/cpp/container/span) or somesuch).

physics/discrete_trajectory2.hpp
```diff
@@ -0,0 +1,110 @@
...
+  DiscreteTrajectory2(DiscreteTrajectory2&&) = default;
+  DiscreteTrajectory2& operator=(DiscreteTrajectory2&&) = default;
+  DiscreteTrajectory2(const DiscreteTrajectory2&) = delete;
+  DiscreteTrajectory2& operator=(const DiscreteTrajectory2&) = delete;
+
+    Iterator begin() const;
+  Iterator end() const;
+
+  Iterator rbegin() const;
+  Iterator rend() const;
```
eggrobin, line 48: I should hope the reverse iterator is not equal to the iterator, otherwise the container cannot contain more than one element.
`std::reverse_iterator<iterator>`, which should probably be named `using reverse_iterator = std::reverse_iterator<iterator>`.

physics/discrete_trajectory2.hpp
```diff
@@ -0,0 +1,110 @@
...
+  Iterator find(Instant const& t) const;
+
+  Iterator lower_bound(Instant const& t) const;
+  Iterator upper_bound(Instant const& t) const;
+
+  SegmentIterator segments_begin() const;
+  SegmentIterator segments_end() const;
+
+  SegmentIterator segments_rbegin() const;
+  SegmentIterator segments_rend() const;
```
eggrobin, line 59: Same comment as for rbegin, `std::reverse_iterator<SegmentIterator>`.

physics/discrete_trajectory2.hpp
```diff
@@ -0,0 +1,110 @@
...
+  Iterator rbegin() const;
+  Iterator rend() const;
+
+  Iterator find(Instant const& t) const;
+
+  Iterator lower_bound(Instant const& t) const;
+  Iterator upper_bound(Instant const& t) const;
+
+  SegmentIterator segments_begin() const;
+  SegmentIterator segments_end() const;
```
eggrobin, line 56: If we want to iterate over the segments, a `SegmentRange` type accessible as `trajectory.segments()` would be useful, so that we can write
```C++
for (auto const& segment: trajectory.segments()) {
```
rather than
```C++
for (auto segments_it = trajectory.segments_begin();
     segments_it != trajectory.segments_end();
     ++segments_it) {
  auto const& segment = *segments_it;

File: physics/discrete_trajectory2.hpp (hunk 1 above)
Comment (pleroy) [reply physics/discrete_trajectory2.hpp:33]: Done.

File: physics/discrete_trajectory2.hpp (hunk 2 above)
Comment (pleroy) [reply physics/discrete_trajectory2.hpp:48]: Done.

File: physics/discrete_trajectory2.hpp (hunk 3 above)
Comment (pleroy) [reply physics/discrete_trajectory2.hpp:59]: Done.

File: physics/discrete_trajectory2.hpp (hunk 4 above)
Comment (pleroy) [reply physics/discrete_trajectory2.hpp:56]: Adding `segments()` and `rsegments()` since it avoids all these `segment_` functions.

File: physics/discrete_trajectory2.hpp [hunk 5]
```diff
@@ -41,35 +47,31 @@ class DiscreteTrajectory2 : public Trajectory<Frame> {
...
+  iterator lower_bound(Instant const& t) const;
+  iterator upper_bound(Instant const& t) const;
 
-  SegmentIterator segments_begin() const;
-  SegmentIterator segments_end() const;
-
-  SegmentIterator segments_rbegin() const;
-  SegmentIterator segments_rend() const;
+  SegmentRange segments() const;
+  ReverseSegmentRange rsegments() const;
```
eggrobin, line 62: I think the Right Thing to do is to have only `segments` and use an adapter to iterate backward (in C++20 that would be https://en.cppreference.com/w/cpp/ranges/reverse_view, maybe absl has a thing in the meantime, or we can write one in `base`, we should have `make_reverse_iterator`).

eggrobin (approved):

physics/discrete_trajectory2.hpp
pleroy: Added a TODO to fix this once we have C++20.

PR #3180, opened 2021-10-30
Convert Part and its test to DiscreteTraject0ry

pleroy:
#3136.

### DIFF
--- a/ksp_plugin/part.cpp
+++ b/ksp_plugin/part.cpp
@@ -170,26 +170,26 @@ DiscreteTraject0ry<Barycentric>::iterator Part::history_end() {
 }
 
 DiscreteTraject0ry<Barycentric>::iterator Part::psychohistory_begin() {
-  if (!psychohistory_.has_value()) {
+  if (psychohistory_ == trajectory_.segments().end()) {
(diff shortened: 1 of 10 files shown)


ksp_plugin/part.cpp
```diff
@@ -161,56 +161,50 @@ RigidMotion<RigidPart, Barycentric> const& Part::rigid_motion() const {
...
+  trajectory_.Append(time, degrees_of_freedom);
 }
 
 void Part::ClearHistory() {
-  if (psychohistory_ != nullptr) {
-    history_->DeleteFork(psychohistory_);
-  }
-  prehistory_->DeleteFork(history_);
-  history_ = prehistory_->NewForkAtLast();
+  trajectory_ = DiscreteTraject0ry<Barycentric>();
```
eggrobin, line 207: Shouldn’t this set `psychohistory_` to `trajectory_.segments().end()`?

pleroy: Good catch, done.  This should also call `clear`, not do an assignment, to preserve `history_`.

eggrobin (approved):

PR #3185, opened 2021-10-30
Convert FlightPlan and its test to DiscreteTraject0ry

pleroy:
#3136.

### DIFF
--- a/ksp_plugin/flight_plan.cpp
+++ b/ksp_plugin/flight_plan.cpp
@@ -54,18 +54,16 @@ FlightPlan::FlightPlan(
       initial_time_(initial_time),
       initial_degrees_of_freedom_(std::move(initial_degrees_of_freedom)),
       desired_final_time_(desired_final_time),
-      root_(make_not_null_unique<DiscreteTrajectory<Barycentric>>()),
       ephemeris_(ephemeris),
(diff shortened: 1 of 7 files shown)


ksp_plugin/flight_plan.cpp
```diff
@@ -460,23 +457,23 @@ absl::Status FlightPlan::ComputeSegments(
...
+  segments_.emplace_back(trajectory_.NewSegment());
   if (anomalous_segments_ > 0) {
     ++anomalous_segments_;
   }
 }
 
 void FlightPlan::ResetLastSegment() {
-  segments_.back()->ForgetAfter(segments_.back()->Fork()->time);
+  auto const last_segment = segments_.back();
+  trajectory_.ForgetAfter(std::next(last_segment->begin()));
```
eggrobin, line 468: Since it has different semantics with respect to forgetting the given point, I think `ForgetAfter` should be renamed to something like `ForgetFrom`: the name `ForgetAfter` is misleading, I would have expected no `next`.
We can do that after migrating to DiscreteTraject0ry if that’s easier.

eggrobin (approved):

ksp_plugin/flight_plan.cpp
pleroy: I think some of the identifiers will need adjusting, but we'll do that at the end.

PR #3190, opened 2021-11-03
Convert Renderer and its test to DiscreteTraject0ry

pleroy:
This requires some fixes to `MockVessel`.

#3136.

### DIFF
--- a/ksp_plugin/renderer.cpp
+++ b/ksp_plugin/renderer.cpp
@@ -74,52 +74,52 @@ Vessel const& Renderer::GetTargetVessel() const {
   return *target_->vessel;
 }
 
-not_null<std::unique_ptr<DiscreteTrajectory<World>>>
+DiscreteTraject0ry<World>
 Renderer::RenderBarycentricTrajectoryInWorld(
     Instant const& time,
-    DiscreteTrajectory<Barycentric>::Iterator const& begin,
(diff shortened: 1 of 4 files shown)


ksp_plugin/renderer.cpp
```diff
@@ -74,52 +74,52 @@ Vessel const& Renderer::GetTargetVessel() const {
...
-  auto trajectory = make_not_null_unique<DiscreteTrajectory<Navigation>>();
+    DiscreteTraject0ry<Barycentric>::iterator const& begin,
+    DiscreteTraject0ry<Barycentric>::iterator const& end) const {
+  DiscreteTraject0ry<Navigation> trajectory;
   for (auto it = begin; it != end; ++it) {
     auto const& [time, degrees_of_freedom] = *it;
     if (target_) {
-      auto const& prediction = target_->vessel->prediction();
-      if (time < prediction.t_min()) {
+      auto const prediction = target_->vessel->prediction();
```
eggrobin, line 103: Why not `auto const& prediction = *target_->vessel->prediction();`?

eggrobin (approved):

ksp_plugin/renderer.cpp
pleroy: It really doesn't matter here, but I have tried to stay away from operating on segments directly, going through iterators instead.

PR #3201, opened 2021-11-13
Use static styles to prevent Unity leaks

### LINKED ISSUE #3064: "Flight Plan UI" associated steady memory increase (~1GB per 20 minutes) on some machines leading to KSP instability/unusual texture lighting/slow crash/system out of memory
Several users on Discord & the KSP forum have recently observed an unusual cluster of symptoms...this issue is to help identify the interaction/cause mix of Unity, KSP, user hardware, normal expected Principia memory use, or bugs in any of these.

Apparent observations reported so far: 

> - after 30+ minutes or so in game, KSP instability/slowness/unusual texture lighting/slow crash- 
> > - having a flight plan with one or more maneuvers open may be a significant factor 
> - Appears to have become apparent around the time of the release of Principia Gröbner
> - so far most users report using windows 10 & NVidia hardware though one Radeon card was mentioned.

Using just KSP 1.11.2 with Principia Gröbner & two test saves provided below, I have observed working set memory to increase from an order of 2GB to an order of 8GB (on a machine with 32GB total RAM) at which point KSP hung during a slow time warp while the orbit analyzer was open for a simple vessel in Mun orbit.

> I documented the above in [this sequence of screenshots found in this album link](https://photos.app.goo.gl/PqVDsw3eM9wbrAK99).

>Principia logs for this session: [Principia_INFO_20210714_161308_4088.zip](https://github.com/mockingbirdnest/Principia/files/6824684/Principia_INFO_20210714_161308_4088.zip)
 
> I also began screen recording for this 1st replication test at the point where the vessel textures lost illumination i.e. 'became dark' which is documented in the following video (working set memory was around 4GB at this point as shown by Process Explorer in the win10 desktop switches captured by the video): 
> 
> https://youtu.be/xUo4hYwPPIw

After the 1st KSP hang/crash, I then reloaded KSP & repeated the flight with journaling on & observed the same working set memory increase up to close to 8GB and then decided to delete the flight plan & set history to 0 & to build a new flight plan.  
> - Over a relatively short period of time while building a new flight plan working set memory returned to around 2.6GB. 
> - No KSP crash occurred, 
> - just in game lag, 
> - loss of text in the Principia flight plan plotting frame window UI,
> - times of strange vessel texture illumination which came & went,
> - as well as loss of main KSP menu textures visible after "quit to main menu" (shown as the last two screen captures in the album provided above for the 1st flight).
If needed I can post the large journal (9GB) for this second 'repeat' flight as well as the long corresponding screen recorded video (5.4GB).

test save used for the 1st replication & video: 
[PrincipiaMun60kFreeFlyBy.zip](https://github.com/mockingbirdnest/Principia/files/6823207/PrincipiaMun60kFreeFlyBy.zip)

test save (created during the 1st video WITHOUT journaling) but used with journaling ON for 2nd replication & video: 
[MunTransfer.zip](https://github.com/mockingbirdnest/Principia/files/6823593/MunTransfer.zip)

pleroy:
Before this PR, the save given in #3064 burns about 120 MiB/minute when the flight planner UI is visible.  After this PR, it uses about 10 MiB/minute.  Note that we do expect some memory growth due to Principia as the trajectories of the celestials and vessels are extended.

Fix #3064.

### DIFF
--- a/ksp_plugin_adapter/style.cs
+++ b/ksp_plugin_adapter/style.cs
@@ -2,6 +2,15 @@
 namespace ksp_plugin_adapter {
 
 internal static class Style {
+  static Style() {
+    horizontal_line_style_ =
+        new UnityEngine.GUIStyle(UnityEngine.GUI.skin.horizontalSlider);
+    horizontal_line_style_.fixedHeight /= 5;
+    horizontal_line_style_.normal.background = ultra_cool_grey_texture;
+    line_spacing_style_ = new UnityEngine.GUIStyle(UnityEngine.GUI.skin.label);
+    line_spacing_style_.fixedHeight /= 5;
+  }
+
   public static UnityEngine.Color Tangent { get; } = XKCDColors.NeonYellow;
   public static UnityEngine.Color Normal { get; } = XKCDColors.AquaBlue;
   public static UnityEngine.Color Binormal { get; } = XKCDColors.PurplePink;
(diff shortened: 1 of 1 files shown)


eggrobin (approved):

Growflavor:
Thank you very much for figuring out the issue!  & I love the meme ;-)  I shared the sentiment while collecting & observing the dozens of gflags traces under different conditions...

PR #3220, opened 2021-11-30
Cache Localizer.Format calls

### LINKED ISSUE #3149: Displaying the reference frame selector has a significant performance impact
TL;DR: On @growflavor’s machine, opening the reference frame selectors degrades the framerate from about 60 FPS to about 15 FPS. Opening other windows also impacts the framerate, albeit less so.

Original issue below.

----

I mentioned on the forum the [primary issue of UI lag with the recently revised reference frame selector UI even on my primary machine that uses an NVIDIA RTX 2070S GPU](https://forum.kerbalspaceprogram.com/index.php?/topic/162200-wip181-191-1101-1110%E2%80%932-1122-principia%E2%80%94version-hadamard-released-2021-10-06%E2%80%94n-body-and-extended-body-gravitation/&do=findComment&comment=4029075).

Per [eggrobin's suggestion at the forum](https://forum.kerbalspaceprogram.com/index.php?/topic/162200-wip181-191-1101-1110%E2%80%932-1122-principia%E2%80%94version-hadamard-released-2021-10-06%E2%80%94n-body-and-extended-body-gravitation/&do=findComment&comment=4029587) that there might be a way to reduce the UI's impact on performance, I am opening this issue to report/track significant Principia UI FPS performance impact primarily with the new reference frame selector.

I also suggest this FPS method as a simple proxy to observe the relative impact of various parts of a UI which might help to identify elements that currently produce the most impact.  For example, 'pinning' the desired frame row & collapsing the full option tree noticibly reduces the FPS impact of the new reference frame selector UI.

Consequently, I am also hopeful that this UI impact via FPS observation method might help lead to insights that may also reduce the performance impact of the Flight Plan UI.

Link to linuxgurugamer's handy FPS tool: https://forum.kerbalspaceprogram.com/index.php?/topic/172692-112x-showfps/

Here is an example screenshot of the FPS impact of one of the new reference frame selector UIs, examples of other UIs may be viewed in the album provided later in this post:

Album of screenshots showing the relative FPS impact of various Principia UIs or combinations of UIs:
https://photos.app.goo.gl/xGtHaFaZvUS26KBU6

Hopefully at least some amount of relative FPS impact replicates on your machines.  Please let me know if there is further information that may be useful to you.  Thank you for the new reference frame selector UIs.  I hope this concept of FPS metric ends up useful to help identify ways to reduce the impact of various Principia UIs on performace.

eggrobin:
Fix #3149. Note that we do not have System.Runtime.Caching, so we need to write our own cache.
The time spent in `RenderWindowAndRecordTooltip` for the reference frame selector changes as follows:
| Lines shown | Before | After |
|---|---|---|
| Sun, Kerbin, Mun, Minmus | 5.7 ms | 500 μs |
| Sun | 1.7 ms | 200 μs |
| All celestials | 22.5 ms | 2 ms |

### DIFF
--- a/ksp_plugin_adapter/localization_extensions.cs
+++ b/ksp_plugin_adapter/localization_extensions.cs
@@ -1,5 +1,6 @@
 ﻿using KSP.Localization;
 using System;
+using System.Collections.Generic;
 using System.Linq;
 using System.Text.RegularExpressions;
 
@@ -26,7 +27,7 @@ internal static class L10N {
   }
 
   public static string Standalone(string name) {
-    return Localizer.Format("#Principia_GrammaticalForm_Standalone", name);
+    return L10N.CacheFormat("#Principia_GrammaticalForm_Standalone", name);
   }
 
   public static string StandaloneName(this CelestialBody celestial) {
@@ -62,6 +63,9 @@ internal static class L10N {
   // Bodies that may not have an article (Saturn, Vénus) have an uppercase
   // gender tag (Saturn^N, Vénus^F).
   public static string Name(this CelestialBody body) {
+    if (names_.TryGetValue(body, out string result)) {
+      return result;
+    }
     string name = LingoonaUnqualified(body.displayName);
     string qualifiers = LingoonaQualifiers(body.displayName);
     if (qualifiers == "") {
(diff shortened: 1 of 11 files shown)


ksp_plugin_adapter/localization_extensions.cs
```diff
@@ -121,9 +129,72 @@ internal static class L10N {
...
        from name in new[]{body.Name(), body.Initial()}
        select name).Concat(from arg in args select arg.ToString()).ToArray();
-    return CelestialOverride(template, names, bodies) ??
-        FormatOrNull(template, names);
+    return GetCached(new StringCacheEntry(template, names),
+                     () => CelestialOverride(template, names, bodies) ??
+                           FormatOrNull(template, names));
+  }
+
+  public static string CacheFormat(string name, params object[] args) {
```
pleroy, line 137: Let's put the cache in a separate class/file.

PR #3225, opened 2021-12-04
Use new-style DiscreteTrajectory in Entwurf

pleroy:
1. Pass the downsampling parameters at construction and not when creating the history: they are used each time we add a new segment.  (Not serialized yet.)
2. Only serialize the tail of the trajectory (12'000 ending at the end of the backstory).
3. Use segments instead of forks everywhere.
4. Simplify the checkpointing code now that we can easily denote the backstory.
5. Simplify/complete the reanimation code using `Merge`.  (Not tested yet.)
6. Fix the tests to use the new trajectory factories, and add a test for tail serialization.

#2400.


eggrobin (approved):

pleroy:
retest this please

pleroy:
retest this please

PR #3250, opened 2021-12-25
Ungarble unsupported languages somewhat

### LINKED ISSUE #3245: The reference frame selector is garbled in unsupported languages
Spotted in a screenshot posted on Discord by @von-Kerman, who uses the Russian UI:

This is because while we fall back to English strings, we fail to find the optional strings since we are not using the `.en-us` suffix when looking them up.

eggrobin:
The result is not great (in particular in stock we end up with translated celestial names inside untranslated message templates, with any hope of sensible grammar long lost), but at least it is not saying `_Selector_` on every button in the reference frame selector.
Fix #3245.
image

### DIFF
--- a/ksp_plugin_adapter/localization_extensions.cs
+++ b/ksp_plugin_adapter/localization_extensions.cs
@@ -35,8 +35,8 @@ internal static class L10N {
   }
 
   private static bool StartsWithCapitalizedDefiniteArticle(string s) {
-    return (Localizer.CurrentLanguage == english_us_ && s.StartsWith("The ")) ||
-           (Localizer.CurrentLanguage == french_ &&
+    return (UILanguage() == english_us_ && s.StartsWith("The ")) ||
+           (UILanguage() == french_ &&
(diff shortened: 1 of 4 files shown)


ksp_plugin_adapter/localization_extensions.cs
```diff
@@ -139,35 +143,38 @@ internal static class L10N {
...
-          for (int n = 1; n <= resolved_args.Count; ++n) {
-            template.Replace($"<<X:{n}>>", resolved_args[n - 1]);
-          }
-          // These escapes are handled by KSP *after* formatting, contrary to
-          // the \u ones and the ｢｣ for {}, which are handled when the Localizer
-          // loads the tags.
-          return Lingoona.Grammar.useGrammar(template, resolved_args)
-              .Replace(@"\n", "\n").Replace(@"\""", "\"").Replace(@"\t", "\t");
-        });
+    string template = Localizer.GetStringByTag(name);
```
pleroy, line 157: Move this variable and the comment block closer to the point of use, near lien 167.

PR #3251, opened 2021-12-25
Initialize the styles more lazily

### LINKED ISSUE #3239: Crash when looking at trajectories without having looked at a window since the process was started
How I encountered this problem: 
Create a flight plan for a current loaded craft in space; (maneuver showed in navball)
Save the game;
Exit KSP and Restart;
Continue the game and immediately load the saved file;
Press M, game froze for seconds and crash.

If I delete the flight plan then press M, I can continue game.

eggrobin:
Fix #3239.

### DIFF
--- a/ksp_plugin_adapter/style.cs
+++ b/ksp_plugin_adapter/style.cs
@@ -2,15 +2,6 @@
 namespace ksp_plugin_adapter {
 
 internal static class Style {
-  static Style() {
-    horizontal_line_style_ =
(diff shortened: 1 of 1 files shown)


ksp_plugin_adapter/style.cs
```diff
@@ -96,8 +97,8 @@ internal static class Style {
   // It is very important to use *static* styles for these elements.  If we use
```
pleroy, line 97: Fix the comment now that the styles are not static.

eggrobin: The styles are still static.

PR #3257, opened 2021-12-26
NaN-aware consistency check

eggrobin:


physics/discrete_trajectory_body.hpp
```diff
@@ -760,13 +760,19 @@ absl::Status DiscreteTrajectory<Frame>::ConsistencyStatus() const {
                            i));
         } else if (timeline_rbegin->degrees_of_freedom !=
```
pleroy, line 761: Give names to these expressions, they show up 4 times.

PR #3260, opened 2021-12-28
Merge was not processing the segments of the target after the end of the source

pleroy:
This led to an inconsistent time-to-segment map (reference to a 1-point segment that was not the most recent).

#3136.


eggrobin (approved):

pleroy:
retest this please

PR #3270, opened 2022-01-01
SO(3) is not commutative

### LINKED ISSUE #3235: The camera rotates in the plotting frame when KSP uses a rotating reference frame
A long-standing bug in #2393; if the active vessel is low enough, the camera will not remain fixed in the plotting frame.

eggrobin:
…but since I tested #3249 on Kerbin with the KCI, KCKF, and KSO frames whose reference planes are all coplanar, I was working in SO(2).

Tested with RSS this time.

Fix #3235.


serialization/geometry.proto
```diff
@@ -160,6 +160,7 @@ message Frame {
     // BUBBLE = 13;
     CAMERA = 14;
     CAMERA_REFERENCE = 17;
+    CAMERA_COMPENSATED_REFERENCE = 21;
```
pleroy, line 163: Alphabetize.

pleroy:
retest this please

PR #3286, opened 2022-01-18
Remove custom optional implementation.

rnlahaye:
It is no longer needed as of #3279.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -121,8 +121,6 @@ endif
 ifeq ($(UNAME_S),Darwin)
     INCLUDES += \
 			-I$(DEP_DIR)compatibility/filesystem \
-			-I$(DEP_DIR)compatibility/optional \
-			-I$(DEP_DIR)Optional \
 			-include "base/macos_allocator_replacement.hpp"
     LIBS += -framework CoreFoundation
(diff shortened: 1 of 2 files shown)


pleroy:

rnlahaye:
ok to test

Makefile
```diff
@@ -121,8 +121,6 @@ endif
 ifeq ($(UNAME_S),Darwin)
     INCLUDES += \
 			-I$(DEP_DIR)compatibility/filesystem \
```
pleroy, line 123: Can we get rid of this one too?

rnlahaye: Filesystem requires 10.15, so no.

Clang 12 is willing to compile filesystem while targeting 10.12, but I haven't actually tried running the resulting binary on a pre-Catalina system; I suspect there might be runtime errors in that case.

eggrobin: Yeah now that I think of it there is a note to that effect in the Makefile (and a citation of Cicero for good measure):
https://github.com/mockingbirdnest/Principia/blob/c8bcc2d328afd07506864ccd5f94e2276a0f0b95/install_deps.sh#L33-L36

Issue #2966, opened 2021-04-30
Investigate if Davenport's q-method could be used in the pile-up


pleroy:
This would avoid the cheesy hacky selection of the "reference part".

See [this post](https://math.stackexchange.com/questions/1634113/davenports-q-method-finding-an-orientation-matching-a-set-of-point-samples) and the articles that it references for details.

pleroy:
**TL;DR: Davenport's Q method is more principled and makes the code much simpler than the old heuristic.  It doesn't, however, result in significantly different behaviours, except perhaps in situations involving high aerodynamic forces, which are anyway hard to analyze because they tend to be chaotic.**

The curves below give the angular frequency of the pile-up over time.  The blue curve is for Davenport's Q method, the yellow curve for the old heuristic.

### First test case: the [sounding rocket](https://github.com/mockingbirdnest/Principia/issues/2519#issuecomment-628810018) in #2519:
![rocket](https://user-images.githubusercontent.com/7423020/164542334-fe68e7d0-aaaa-418b-a01c-a59dab1b2f3f.png)
Clearly, there are no visible differences, other than Davenport keeping the rocket intact for a few more seconds.  This vessel is very rigid, so that's not unexpected.

### Second test case: the [extreme spin-up](https://github.com/mockingbirdnest/Principia/issues/2519#issuecomment-629881597) reported by @scimas in #2519.
Before the rapid unscheduled disassembly:
![scimas_stock](https://user-images.githubusercontent.com/7423020/164543067-0ccf8034-e776-4720-9974-f4475f6ff70d.png)
and after:
![scimas_stock2](https://user-images.githubusercontent.com/7423020/164543142-eaebfa00-b56e-44d2-bbb0-8419f7590d59.png)
The time shift in the second graph is believed to be an effect of the chaotic nature of the RUD.  Because the vessel is less rigid that in the previous test case, the curves differ somewhat, but they remain quite similar.  In particular, the absurdly fast spin-up has no physical reality (it should amortize) but it is probably due to KSP being insane and not to the way we handle rotations.

### Third test case: same as above, but with FAR.
The two graphs below show what happens *after* the RUD (the evolution before the RUD is not all that interesting) in two experiments:
![scimas_far](https://user-images.githubusercontent.com/7423020/164544081-28802e6d-bbc4-4cf0-8ac0-6c52216a4e62.png)
[...]

PR #3287, opened 2022-01-19
Add computation of Rayleigh quotient

pleroy:
This requires the addition of class `UnboundedMatrix` and of a variety of `operator*`.

#2966.

### DIFF
--- a/numerics/fixed_arrays.hpp
+++ b/numerics/fixed_arrays.hpp
@@ -42,10 +42,10 @@ class FixedVector final {
   std::array<Scalar, size_> data_;
 
   template<typename L, typename R, int s>
-  friend Product<L, R> operator*(FixedVector<L, s> const& left,
-                                 FixedVector<R, s> const& right);
+  friend constexpr Product<L, R> operator*(FixedVector<L, s> const& left,
+                                           FixedVector<R, s> const& right);
   template<typename L, typename R, int r, int c>
-  friend FixedVector<Product<L, R>, r> operator*(
+  friend constexpr FixedVector<Product<L, R>, r> operator*(
       FixedMatrix<L, r, c> const& left,
       FixedVector<R, c> const& right);
   template<typename S, int s>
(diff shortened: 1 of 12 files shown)


numerics/fixed_arrays.hpp
```diff
@@ -41,8 +41,11 @@ class FixedVector final {
  private:
   std::array<Scalar, size_> data_;
 
+  template<typename L, typename R, int s>
+  friend constexpr Product<L, R> operator*(FixedVector<L, s> const& left,
+                                           FixedVector<R, s> const& right);
```
eggrobin, line 46: Implicit transposition tends to be very messy; in particular we lose associativity: A(vv) is a matrix, (Av)v is a scalar.

We could have a DotProduct, but that is going to make the code look less like the references, which is bad from the point of view of formula translation. 
Let us go with
```C++
template<typename V>
struct TransposedView {
  V const& transpose;
};
```
In `FixedVector`:
```C++
TransposedView<FixedVector> Transpose() const {
  return {.transpose = *this};
}
```
And then this operator becomes:
```C++
  template<typename L, typename R, int s>
  friend constexpr Product<L, R> operator*(TransposedView<FixedVector<L, s>> const& left,
                                           FixedVector<R, s> const& right);
```

eggrobin: (Note: we may want to have a more general and more powerful `TransposedView` in the future as the result of all `Transpose` (with conversions to the appropriate matrix types if actual transposition of the data is needed), but this is a rabbit hole for another day; for now this draft is good enough.)

numerics/fixed_arrays_test.cpp
```diff
@@ -73,7 +73,7 @@ TEST_F(FixedArraysTest, Assignment) {
 }
 
 TEST_F(FixedArraysTest, Multiplication) {
-  EXPECT_EQ(35, v4_ * v4_);
+  EXPECT_EQ(35, TransposedView{v4_} * v4_);
```
eggrobin, line 76: `v4_.Transpose() * v4`.

numerics/matrix_computations_body.hpp
```diff
@@ -212,6 +227,13 @@ ForwardSubstitution(LowerTriangularMatrix const& L,
   return x;
 }
 
-}  // namespace internal_arrays
+template<typename Matrix, typename Vector>
+typename RayleighQuotientGenerator<Matrix, Vector>::Result RayleighQuotient(
+    Matrix const& A, Vector const& x) {
+  // [GV13], section 8.2.3.
+  return x * (A * x) / (x * x);
```
eggrobin, line 234: This should no longer compile.

eggrobin (approved):

eggrobin (approved):

PR #3289, opened 2022-01-20
Compatibility with latest version of Clang

rnlahaye:
Some fixes so that the code compiles and the tests pass under Clang 13.

### DIFF
--- a/base/macros.hpp
+++ b/base/macros.hpp
@@ -196,6 +196,14 @@ inline void noreturn() { std::exit(0); }
 #  define CONSTEXPR_INFINITY const
 #endif
 
+// Clang does not like FP arithmetic that yields a NaN in constexpr code.
+// https://github.com/llvm-mirror/clang/blob/aa231e4be75ac4759c236b755c57876f76e3cf05/lib/AST/ExprConstant.cpp#L2611-L2618
+#if PRINCIPIA_COMPILER_CLANG || PRINCIPIA_COMPILER_CLANG_CL
+#  define CONSTEXPR_NAN const
+#else
+#  define CONSTEXPR_NAN constexpr
+#endif
+
 #if PRINCIPIA_COMPILER_MSVC
 #define MSVC_ONLY_TEST(test_name) test_name
 #else
--- a/base/macros.hpp
+++ b/base/macros.hpp
@@ -197,7 +197,7 @@ inline void noreturn() { std::exit(0); }
 #endif
 
 // Clang does not like FP arithmetic that yields a NaN in constexpr code.
-// https://github.com/llvm-mirror/clang/blob/aa231e4be75ac4759c236b755c57876f76e3cf05/lib/AST/ExprConstant.cpp#L2611-L2618
+// https://github.com/llvm/llvm-project/blob/llvmorg-13.0.0/clang/lib/AST/ExprConstant.cpp#L2860-L2867
 #if PRINCIPIA_COMPILER_CLANG || PRINCIPIA_COMPILER_CLANG_CL
 #  define CONSTEXPR_NAN const
 #else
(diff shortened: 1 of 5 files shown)


pleroy:

rnlahaye:
ok to test

mathematica/mathematica_test.cpp
```diff
@@ -116,7 +118,12 @@ TEST_F(MathematicaTest, ToMathematica) {
     EXPECT_EQ("Minus[0]", ToMathematica(-0.0));  // Not that this does anything.
     EXPECT_EQ("Infinity", ToMathematica(Infinity<double>));
     EXPECT_EQ("Minus[Infinity]", ToMathematica(-Infinity<double>));
-    EXPECT_EQ("Minus[Indeterminate]", ToMathematica(Sqrt(-1)));
+    EXPECT_EQ("Indeterminate", ToMathematica(NaN<double>));
+    EXPECT_EQ("Minus[Indeterminate]", ToMathematica(-NaN<double>));
+
+    // The sign of √-1 varies between Clang versions.
```
eggrobin, line 124: I am somewhat puzzled: it should just call the intrinsic (`PRINCIPIA_USE_SSE3_INTRINSICS` is false in debug on Windows, but I don’t think we have a debug build on *nix).
Is this a Rosetta2 thing? Or is it some chip that has opinions on the sign of i?
https://github.com/mockingbirdnest/Principia/blob/88817a719d78efdfac1435b7e1ad6be1e3e43155/quantities/elementary_functions_body.hpp#L75-L81

physics/rigid_motion_body.hpp
```diff
@@ -57,8 +57,8 @@ RigidMotion<FromFrame, ToFrame>::angular_velocity_of() const {
   } else if constexpr (std::is_same_v<F, FromFrame>) {
     return Inverse().angular_velocity_of_to_frame_;
   } else {
-    static_assert(std::disjunction_v<std::is_same_v<F, ToFrame>,
-                                     std::is_same_v<F, FromFrame>>,
+    static_assert(std::disjunction_v<std::is_same<F, ToFrame>,
+                                     std::is_same<F, FromFrame>>,
```
eggrobin, line 61: It is disturbing that this ever compiled, let alone on both clang and MSVC!
I think it would be more readable to have `std::is_same_v<meow> || std::is_same_v<purr>`; not sure why I didn’t flag that when I reviewed the PR that added those `disjunction`s.

eggrobin (approved):

base/macros.hpp
```diff
@@ -196,6 +196,14 @@ inline void noreturn() { std::exit(0); }
 #  define CONSTEXPR_INFINITY const
 #endif
 
+// Clang does not like FP arithmetic that yields a NaN in constexpr code.
+// https://github.com/llvm-mirror/clang/blob/aa231e4be75ac4759c236b755c57876f76e3cf05/lib/AST/ExprConstant.cpp#L2611-L2618
```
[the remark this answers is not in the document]
rnlahaye: Done.

physics/rigid_motion_body.hpp
rnlahaye: Done.

mathematica/mathematica_test.cpp
rnlahaye: It looks like this was an x86 vs ARM difference which I mistakenly thought was a clang difference. I reverted it.

eggrobin (approved):

pleroy:
There are 264 commits on their way in as part of #3293.  It's likely that it won't build with Clang.  I'll fix whatever is uncovered by the Azure pipelines, which are on Clang 8.

PR #3304, opened 2022-01-23
Make not_null<foo*> trivially copyable.

rnlahaye:
This fixes an annoying -Wrange-loop-construct clang warning.

Done by creating a wrapper struct that uses the default move assignment operator for trivially move assignable pointers and the custom swap-based move assignment operator for others.

### DIFF
--- a/base/not_null_test.cpp
+++ b/base/not_null_test.cpp
@@ -270,22 +270,22 @@ TEST_F(NotNullTest, RValue) {
 
 TEST_F(NotNullTest, TypeTraits) {
   // not_null<foo*> is trivially copyable.
-  EXPECT_TRUE(std::is_trivially_copyable_v<not_null<int*>>);
-  EXPECT_TRUE(std::is_trivially_copy_constructible_v<not_null<int*>>);
-  EXPECT_TRUE(std::is_trivially_copy_assignable_v<not_null<int*>>);
-  EXPECT_TRUE(std::is_trivially_move_constructible_v<not_null<int*>>);
-  EXPECT_TRUE(std::is_trivially_move_assignable_v<not_null<int*>>);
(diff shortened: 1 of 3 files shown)


pleroy:
Lint errors:
```
  | not_null.hpp:104 | runtime/explicit | High | 1
  | not_null.hpp:117 | runtime/explicit | High | 1
  | not_null.hpp:123 | build/include_what_you_use | Normal | 1
  | not_null.hpp:218 | whitespace/line_length | Low | 1
  | not_null.hpp:97 | whitespace/operators | Normal | 1

```

base/not_null_test.cpp
```diff
@@ -262,5 +268,25 @@ TEST_F(NotNullTest, RValue) {
   EXPECT_EQ(5, *owner_const_int);
 }
 
+TEST_F(NotNullTest, TypeTraits) {
+  // not_null<foo*> is trivially copyable.
+  EXPECT_TRUE(std::is_trivially_copyable_v<not_null<int*>>);
```
eggrobin, line 273: Change all of those to `static_assert` (cf. https://github.com/mockingbirdnest/Principia/blob/master/numerics/next_test.cpp which is tests constexpr code).

eggrobin (changes requested):

base/not_null_test.cpp
rnlahaye: Done.

pleroy:
The linter is still unhappy (and somewhat dumb):
<html>
<body>


not_null.hpp:124 | build/include_what_you_use | Normal | 1
-- | -- | -- | --
Add #include <algorithm> for swap


</body>
</html>

eggrobin (approved):

PR #3316, opened 2022-02-03
Added time units to benchmarks

rnlahaye:
Because I don't need to know a multi-second benchmark to nanosecond precision.
For example, `BM_EphemerisKSPSystem` goes from `16353847667  ns` to `16.3 s`
Microseconds, milliseconds, and seconds are used where appropriate.
Also replaced `while (state.KeepRunning())` with the updated form `for (auto _ : state)`, and updated `state.range_x()`and `state.range_y()` to `state.range(0)` and `state.range(1)`.
New benchmark output is here: [bench.txt](https://github.com/mockingbirdnest/Principia/files/7991189/bench.txt).

### DIFF
--- a/benchmark_automation/benchmark_automation.cs
+++ b/benchmark_automation/benchmark_automation.cs
@@ -1,4 +1,4 @@
-using System;
+﻿using System;
 using System.Diagnostics;
 using System.IO;
 
@@ -84,7 +84,8 @@ class BenchmarkAutomation {
                 CommaSeparatedAppend(ref csv_means, μ.ToString());
               } else if (!has_repetitions) {
                 string benchmark_name = words[0];
-                double μ = double.Parse(words[1]);
+                double μ = double.Parse(words[1]) *
+                    TimeConversionFactor(words[2]);
                 CommaSeparatedAppend(
                     ref csv_benchmark_names,
                     "\"" + benchmark_name.Replace("\"", "\"\"") + "\"");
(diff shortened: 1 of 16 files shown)


eggrobin:
> Because I don't need to know a multi-second benchmark to nanosecond precision.

This is a good idea, but https://github.com/mockingbirdnest/Principia/blob/master/benchmark_automation/benchmark_automation.cs needs to be adapted to either understand or override that.

rnlahaye:
Is there any reason why `benchmark_automaton.cs` doesn’t use the JSON output mode of the benchmark?

pleroy:
Ah, but there is a reason.  It was written looong before JSON output was implemented in the benchmark library.  And updating tooling that works is not overly exciting, so it stayed that way...

benchmark_automation/benchmark_automation.cs
```diff
@@ -1,4 +1,4 @@
-﻿using System;
```
pleroy, line 1: I am not sure what happened here.  Our files start with a BOM when they contain non-ASCII characters.  Did we lose the BOM?

rnlahaye: Done.

benchmark_automation/benchmark_automation.cs
```diff
@@ -75,7 +75,8 @@ class BenchmarkAutomation {
                     words[0].Substring(
                         startIndex : 0,
                         length     : words[0].Length - mean_suffix.Length);
-                double μ = double.Parse(words[1]);
+                double μ = double.Parse(words[1]) *
+                    TimeConversionFactor(words[2]);
```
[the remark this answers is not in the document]
rnlahaye: Done.

pleroy:
retest this please

PR #3318, opened 2022-02-06
Cleanup and simplify the array APIs

pleroy:
Notably:
1. Change matrix indexing to use `A(i, j)` instead of `A[i][j]`.  It avoids having helper classes for the rows and we'll be able to convert it to `A[i, j]` in C++23.
2. Rename `dimension` into `size` because that's what it gives, the size of the storage (surely that the dimension of some vector space, but we don't care that much).
3. Add `!=` for all arrays.
4. Systematically declare `constexpr` functions `rows`, `columns` and `size`.


eggrobin (approved):

pleroy:
retest this please

PR #3333, opened 2022-02-19
Fix the clang builds

eggrobin:


quantities/quantities.hpp
```diff
@@ -104,8 +104,8 @@ class Quantity final {
   template<typename Q>
   friend constexpr Q SIUnit();
 
-  template<typename D>
-  friend __m128d ToM128D(Quantity<D> x);
+  template<typename Dimensions>
```
pleroy, line 107: Why don't we call the template parameter of `Quantity` `Dimensions_`?  That's what we do these days to avoid name clashes.

Issue #3358, opened 2022-05-02
Investigate the possibility to display equipotential lines


pleroy:
What I have in mind is [something like this](https://static.scientificamerican.com/sciam/cache/file/799B2D41-709C-4511-9D60A3E4B946DB55_source.png?w=590&h=800&5081446F-BF42-45E4-89D85528A2C70365).  There are many interesting challenges, though, in particular (1) how insanely costly is that going to be? (2) what kind of weird UI do we need to let the user pick what they want?

Of course, this may not pan out, but it seems rather interesting.

pleroy:
Equipotentials in the Earth-Centred reference frame that fixes the direction of the Moon (backward from the in-game MEO).
![equipotentials](https://user-images.githubusercontent.com/7423020/168495207-bfcd05ea-325a-4789-af49-6e952413aa32.gif)

eggrobin:
Note on visualization: the equipotentials move (and the energy of a free-falling vessel can change over time, because since the potential is time-dependent, so that Noether’s theorem does not apply).
This means, for instance, that if we draw the equipotential of the vessel at any point in time, the vessel may be seen to cross its own equipotential, defeating the point of drawing equipotentials. This is not unlike a trajectory going through the current position of the moving Moon. However, in that case, the motion of the moon is indicated by its trajectory (a faded line in the past, a solid line in the future).

We may want to similarly draw the equipotential of the vessel’s energy at multiple points in time, so that any motion is evident. As usual, it is best to work in a reference frame wherein things you care about don’t move, but that requires being able to see whether they move.

eggrobin:
Update: The equipotentials move too much in all our current frames to be of any practical use, and would thus only clutter and confuse.

In particular, in the MEO frame, the shape of the potential changes drastically over a month, obscuring most of the Lagrange points.

The animated GIF from https://github.com/mockingbirdnest/Principia/issues/3358#issuecomment-1127054011 is more promising: the Earth is close to the barycentre, so this potential is nearly the one that would have the Lagrange points, absent the Sun; they are still legible here. However, because the orbit is eccentric, the potential grows and shrinks over the course of a month, making trajectories that stay near a Lagrange point appear to move away and fall into the Moon's well.

It may be time to seriously look into the long-discussed rotating-pulsating frame, wherein the potential should look much the same.

eggrobin:
[...]

PR #3365, opened 2022-05-17
RotationFreeGeometricAccelerationAtRest

eggrobin:
For #3358.

### DIFF
--- a/physics/dynamic_frame.hpp
+++ b/physics/dynamic_frame.hpp
@@ -61,16 +61,17 @@ class DynamicFrame {
 
   // The acceleration of a particle at rest in |ThisFrame| at the given
   // |position| owing to non-inertial motion of |ThisFrame| and gravity,
-  // excluding components with a rotation; that is, if a is the vector field
-  // given by GeometricAcceleration, this function returns
-  //   a - (rot a) r / 2,
-  // where r is the radial vector.
-  // In a rotating reference frame, this may alternatively be expressed as the
-  // second derivative of position with respect to the parametrization on the
-  // angle θ rather than time (which eliminates the Euler acceleration),
-  // converted to an acceleration (in m/s² rather than m/rad²) by multiplying by
-  // the value of θ′² at t.
-  // This acceleration derives from a potential.
+  // excluding components with a rotation.
+  // Let r be the radial vector (from the origin of |ThisFrame|) corresponding
+  // to |position|.
+  // Let r ↦ r″ be the vector field  of free fall accelerations from a rest in
+  // |ThisFrame| at t. This function returns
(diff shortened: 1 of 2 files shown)


physics/dynamic_frame.hpp
```diff
@@ -59,6 +59,24 @@ class DynamicFrame {
       Instant const& t,
       DegreesOfFreedom<ThisFrame> const& degrees_of_freedom) const;
 
+  // The acceleration of a particle at rest in |ThisFrame| at the given
+  // |position| owing to non-inertial motion of |ThisFrame| and gravity,
+  // excluding components with a rotation.
+  // Let r be the radial vector (from the origin of |ThisFrame|) corresponding
+  // to |position|.
+  // Let r ↦ r″ be the vector field  of free fall accelerations from a rest in
```
pleroy, line 67: Remove extra space after `field`.

Also "from a rest" is weird.

physics/dynamic_frame.hpp
```diff
@@ -59,6 +59,24 @@ class DynamicFrame {
...
+  // |ThisFrame| at t. This function returns
+  //   a = r″ - (rot r″) r / 2.
+  // In a rotating reference frame, this may equivalently be expressed using
+  // the second derivative of position with respect to the parametrization on
+  // the angle θ rather than time, which eliminates the Euler acceleration:
+  //   a = θ′² d²r/dθ², starting from a rest defined as dr/dθ = 0.
```
pleroy, line 73: Can we have TEX documentation for this change of variable?  (And for the `rot` formula above.)

physics/dynamic_frame_body.hpp
```diff
@@ -85,6 +85,39 @@ DynamicFrame<InertialFrame, ThisFrame>::GeometricAcceleration(
   return gravitational_acceleration_at_point + fictitious_acceleration;
 }
 
+template<typename InertialFrame, typename ThisFrame>
+Vector<Acceleration, ThisFrame>
+DynamicFrame<InertialFrame, ThisFrame>::RotationFreeGeometricAccelerationAtRest(
```
pleroy, line 90: I would really prefer if we factored out this code by having an auxiliary function that is mostly a copy of `GeometricAcceleration` with 5 out parameters for the 5 components of the acceleration.  `GeometricAcceleration` would just call that function and sum the results; and `RotationFreeGeometricAccelerationAtRest` would call it with a null velocity (because "at rest") and drop the Euler acceleration.

physics/dynamic_frame.hpp
eggrobin: Not today.

eggrobin: Not today. #3366.

PR #3384, opened 2022-06-21
A close window button.

### LINKED ISSUE #2295: Windows should have a close button
ZAJC3W said on the forum: "May i have GUI suggestion? Add close window buttons(to all windows) so we don't have to search for button that opened them. It can get tedious when planning manoeuvres."

eggrobin:
Not the prettiest thing, but it should help with usability when there are lots of windows around.

image

A potentially confusing quirk is that since closing the main window has the exact same effect as toggling the toolbar button, it closes all windows. It’s not clear that anything else would work well though, since there wouldn’t be an unambiguous interaction to bring it back.

Fix #2295.

### DIFF
--- a/ksp_plugin_adapter/window_renderer.cs
+++ b/ksp_plugin_adapter/window_renderer.cs
@@ -267,14 +267,15 @@ internal abstract class BaseWindowRenderer : ScalingRenderer, IConfigNode {
   private UnityEngine.GUISkin skin_;
   private bool must_centre_ = true;
   private bool show_ = false;
-  private UnityEngine.Rect rectangle_;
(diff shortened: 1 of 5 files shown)


ksp_plugin_adapter/window_renderer.cs
```diff
@@ -291,7 +292,21 @@ public interface ISupervisor {
...
+            y: 0,
+            width: Width(1),
+            height: Width(1)),
+            "×")) {
+      Hide();
+    }
+    RenderWindowContents(window_id);
+  }
+
+      public void DisposeWindow() {
```
pleroy, line 309: Bad indent.

pleroy (approved):

PR #3398, opened 2022-07-17
Declutter

### LINKED ISSUE #2578: Hide periapsises/apoapsises and ascending/descending nodes
When planning flight, sometimes there are so many PA/AP/AN/DN points on the track, that you can not switch to the body or vessel because instead of clicking them you click this points. It would be useful to have possibility to hide them.

### LINKED ISSUE #3394: Declutter toggle
In #3035, there was a suggestion of an option to hide trajectories; the motivation there was performance, and we have greatly improved that in हरीश चंद्र.
We could improve performance further (we probably shouldn’t update all the meshes all the time), but there is a separate question which is whether we should have an option to hide trajectories; from the conversations on discord at the time, 
> Perhaps a declutter button may be of some use (even without performance concerns, sometimes there is a bit too much noodliness for readability).

For instance in ECEF (which should become increasingly useful even with long histories or predictions once Σκοπός makes ground tracks matter), the sky covered with the ground tracks of the entire solar system, which are useless.

We have other clutter issues; in particular apsides and node markers get out of hand:
- #2578.

In the reference frame selector, it is possible to pin some celestials of interest.
Map markers can be pinned by right-clicking on them.

We could have a declutter mode wherein all trajectories are hidden except those of celestials pinned in the plotting frame selector, and all map markers are hidden except the pinned ones.

The second part is a bit of a mess though, because the markers don’t have an identity; see #3157 (which effectively suggests mitigating that with a mode with _more_ clutter!).
We could probably do a better job of trying to keep track of markers; right now if the total number of apoapsides on a trajectory changes, we change the currently-displayed periapsis, because the index is among all nodes in the pool.

We could probably improve things quite a bit by having a pool per provenance (which would fit nicely with our cap on the number of nodes being per provenance).
This would probably deal with #2196 once and for all.

eggrobin:
On top of #3397.

Fixes #3394.
Fixes #2578.

Adds the following line to the main window:
image

The buttons serves as a general-purpose « my screen is a mess, make it stop » button; it checks both options. In the future as we add more clutter (for instance as part of #3358) it would remove that too.

Demonstration in ECEF with a flight plan from LEO; the prediction is set to 16 000 steps at 10 m, it seems to span about a month—and thus we get a month’s worth of the solar system’s path in the sky—:
| Before | After |
|---|---|
| image | image |

### DIFF
--- a/ksp_plugin_adapter/main_window.cs
+++ b/ksp_plugin_adapter/main_window.cs
@@ -214,21 +214,21 @@ internal class MainWindow : VesselSupervisedWindowRenderer {
       if (adapter_.PluginRunning()) {
         plotting_frame_selector_.RenderButton();
         using (new UnityEngine.GUILayout.HorizontalScope()) {
-        if (UnityEngine.GUILayout.Button(
-                L10N.CacheFormat("#Principia_MainWindow_Declutter"),
-                GUILayoutWidth(5))) {
-          show_only_pinned_markers = true;
-          show_only_pinned_celestials = true;
-        }
-        UnityEngine.GUILayout.Label(
-            L10N.CacheFormat("#Principia_MainWindow_Declutter_ShowOnly"));
-        show_only_pinned_markers = UnityEngine.GUILayout.Toggle(
-            show_only_pinned_markers,
-            L10N.CacheFormat("#Principia_MainWindow_Declutter_PinnedMarkers"));
-        show_only_pinned_celestials = UnityEngine.GUILayout.Toggle(
-            show_only_pinned_celestials,
-            L10N.CacheFormat(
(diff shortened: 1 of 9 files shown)


ksp_plugin_adapter/main_window.cs
```diff
@@ -210,6 +213,23 @@ internal class MainWindow : VesselSupervisedWindowRenderer {
       // interface.
       if (adapter_.PluginRunning()) {
         plotting_frame_selector_.RenderButton();
+        using (new UnityEngine.GUILayout.HorizontalScope()) {
+        if (UnityEngine.GUILayout.Button(
```
pleroy: Bad indent.

PR #3414, opened 2022-08-07
A better test of the Euler force for dynamic frames

pleroy:


physics/dynamic_frame_test.cpp
```diff
@@ -400,6 +401,103 @@ TEST_F(DynamicFrameTest, CentrifugalAcceleration) {
...
+  // No gravity.
+  Vector<Acceleration, Inertial> const gravitational_acceleration;
+  EXPECT_CALL(mock_frame_, GravitationalAcceleration(_, _))
+      .WillRepeatedly(Return(gravitational_acceleration));
+
+  // The test point is on the x axis.
+  DegreesOfFreedom<Rotating> const initial_state_in_rotating_frame = {
+      Rotating::origin + Displacement<Rotating>({100 * Metre,
+                                                 0 * Metre,
+                                                 0 * Metre}),
```
eggrobin, line 450: While our APIs make it very hard to treat x and y differently at this level, z is another story, since it is the axis. In this test and the ones above, use a nonzero speed and displacement in z (preferably with numbers that looks nothing like the ones we are going to encounter, _e.g._, 17 * Metre).

physics/dynamic_frame_test.cpp
```diff
@@ -400,6 +401,103 @@ TEST_F(DynamicFrameTest, CentrifugalAcceleration) {
...
+      Rotating::unmoving};
```
eggrobin, line 451: In this test, where we are initially at rest, and therefore have no Coriolis acceleration, use a general velocity: here x direction is made special by the position, as it is the radial direction.

physics/dynamic_frame_test.cpp
```diff
@@ -400,6 +401,103 @@ TEST_F(DynamicFrameTest, CentrifugalAcceleration) {
...
+                                              0 * Radian / Second / Second,
+                                              ωʹ});
+        Rotation<Inertial, Rotating> const rotation(
+            ωʹ * Pow<2>(t - t0) / 2,
+            angular_acceleration_of_to_frame,
+            DefinesFrame<Rotating>{});
+        RigidTransformation<Inertial, Rotating> const
+            rigid_transformation(
+                /*from_origin=*/Inertial::origin,
+                /*to_origin=*/Rotating::origin,
```
eggrobin, line 425: On second thought, it is probably a bad idea to conflate these origins, since `Rotating::origin` is involved in the computation (it is the axis), and `Inertial::origin` should not be, but could easily be used.
Since this has no effect on the expectations, use a general position for all the tests.

eggrobin (approved):

PR #3423, opened 2022-08-20
Change the flight plan tolerance and steps to be consistent with the prediction

### LINKED ISSUE #3395: The flight plan and prediction tolerance settings are inconsistent
The flight plan tolerance is in powers of 2, whereas the prediction tolerance is in powers of 10.
I suspect this is because the prediction setting was originally persisted in the C# only, so that it was convenient to save an index into an array of powers of 10, but since the flight plan setting started out persisted in the C++ as a `Length`, powers of 2 were easier because of the exact `*=` and `/=`.

pleroy:
Fix #3395.

### DIFF
--- a/ksp_plugin_adapter/flight_planner.cs
+++ b/ksp_plugin_adapter/flight_planner.cs
@@ -265,9 +265,8 @@ class FlightPlanner : VesselSupervisedWindowRenderer {
             UpdateStatus(status, null);
           }
           UnityEngine.GUILayout.TextArea(
-              L10N.CacheFormat("#Principia_PredictionSettings_ToleranceText",
-                               length_integration_tolerances_names_[
-                                   length_integration_tolerance_index_]),
+              length_integration_tolerances_names_[
+                  length_integration_tolerance_index_],
               GUILayoutWidth(3));
           if (length_integration_tolerance_index_ ==
               integration_tolerances_.Length - 1 ||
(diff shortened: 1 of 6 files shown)


ksp_plugin_adapter/flight_planner.cs
```diff
@@ -244,15 +266,19 @@ class FlightPlanner : VesselSupervisedWindowRenderer {
           }
           UnityEngine.GUILayout.TextArea(
               L10N.CacheFormat("#Principia_PredictionSettings_ToleranceText",
-                               parameters.length_integration_tolerance.ToString(
-                                   "0.0e0")),
+                               length_integration_tolerances_names_[
+                                   length_integration_tolerance_index_]),
```
eggrobin, line 270: Just remove the localized string (also from the localization files).

ksp_plugin_adapter/flight_planner.cs
```diff
@@ -684,8 +710,31 @@ class FlightPlanner : VesselSupervisedWindowRenderer {
...
+  private static readonly double[] integration_tolerances_ =
+      {1e-6, 1e-5, 1e-4, 1e-3, 1e-2, 1e-1, 1e0, 1e1, 1e2, 1e3, 1e4, 1e5, 1e6};
+  private static readonly string[] length_integration_tolerances_names_ = {
+      "1 µm", "10 µm", "100 µm", "1 mm", "1 cm", "10 cm", "1 m", "10 m",
+      "100 m", "1 km", "10 km", "100 km", "1000 km"
+  };
+  private static readonly string[] speed_integration_tolerances_names_ = {
+      "1 µm/s", "10 µm/s", "100 µm/s", "1 mm/s", "1 cm/s", "10 cm/s", "1 m/s",
+      "10 m/s", "100 m/s", "1 km/s", "10 km/s", "100 km/s", "1000 km/s"
+  };
```
eggrobin, line 733: Unused, remove.
Also remove this TODO  in the main window :
https://github.com/mockingbirdnest/Principia/blob/5c17d2580bcebc754f9e6e72dae3e71059547ec5/ksp_plugin_adapter/main_window.cs#L430
Let’s not go there.

eggrobin (approved):

PR #3424, opened 2022-08-20
Consistent axis conventions

eggrobin:
This means that if the active vessel is in a sufficiently Keplerian orbit, the osculating elements given by the stock game (and surfaced, e.g., by MechJeb) will roughly be consistent with the elements given by the orbit analyser.

Of course since the universe is tilted, orbits about other bodies will have weird elements; but this should fix issues when using MJ to launch into a specific plane.


pleroy:
retest this please

PR #3453, opened 2022-11-13
/utf-8

eggrobin:
See https://learn.microsoft.com/en-us/cpp/build/reference/utf-8-set-source-and-executable-character-sets-to-utf-8?view=msvc-170.


eggrobin:
retest this please

eggrobin:
retest this please

PR #3466, opened 2022-11-23
Add support for optimization of 1- or 2-D problems

pleroy:
If the box passed to the multi-level single linkage algorithm has one or two of its dimensions which are 0, the algorithm fails because the critical radius becomes 0.

Detect these cases and switch to a 1- or 2-D computation of the critical radius in that case.

The tests that use 2-D functions can now be written without an artificial term in `x₀²`.  This is important because that term is likely to change the size and shape of the regions of attraction, therefore altering the difficulty of the problem.

#3358


numerics/global_optimization.hpp
```diff
@@ -49,7 +51,10 @@ class MultiLevelSingleLinkage {
     Argument centre;
     std::array<Difference<Argument>, 3> vertices;
 
-    Cube<NormType> Measure() const;
+    // If some of the dimensions of the box are zero, we handle the problem as a
+    // 1- or 2-dimensional problem.  This only affects the computation of rₖ.
+    using Measure = std::variant<NormType, Square<NormType>, Cube<NormType>>;
```
eggrobin, line 56: Do we really expect to have cases where we do not know the dimension of the problem at runtime?

pleroy: Done.  That's indeed nicer.

pleroy:
retest this please

eggrobin (approved):

PR #3469, opened 2022-11-27
A global optimization test with a function similar to the gravitational potential

pleroy:
And a fix to limit the gradient descent to a finite radius around the start point.

#3358.


eggrobin (approved):

pleroy:
retest this please

PR #3515, opened 2023-01-25
An explicit linear multistep integrator

pleroy:
Also, Adams-Bashforth methods of order 2 and 3.


integrators/explicit_linear_multistep_integrator.hpp
```diff
@@ -0,0 +1,120 @@
...
+        decltype(std::declval<typename ODE::State>().y);
+
+    // The data for a previous step of the integration.
+    struct Step final {
+      DoublePrecision<IndependentVariable> s;
+      DoubleDependentVariables y;
+      DependentVariableDerivatives yʹ;
+    };
+
+    class Starter : public integrators::Starter<ODE, Step, /*steps=*/order> {
```
eggrobin, line 67: The integrator should not require that the order be equal to the number of steps; that is a property of the method.

integrators/explicit_linear_multistep_integrator_body.hpp
```diff
@@ -0,0 +1,258 @@
...
+  IndependentVariableDifference const& h = step;
+  // Current independent variable.
+  DoublePrecision<IndependentVariable> s = previous_steps.back().s;
+
+  // Current state.  This is a non-const reference whose purpose is to make
+  // the equations more readable.
+  auto& y = current_state.y;
+
+  // Order.
+  int const k = order;
```
eggrobin, line 63: `k` is the number of steps, not the order (in [HW10] the order is r).

eggrobin (changes requested):

pleroy:
retest this please

eggrobin (approved):

Issue #3321, opened 2022-02-06
Renamespacing


eggrobin:
The piles of `using`s throughout the codebase are getting unwieldy, unmaintainable, and uninformative.
Consider `ephemeris_body.hpp`:
https://github.com/mockingbirdnest/Principia/blob/b0bec96cae90105aafa39131f140449fd995936d/physics/ephemeris_body.hpp#L35-L69
What use is it to know that we use `Quotient` and not `Product`? Not everything in that list is actually used anyway: `J2000` is unused, and merely reflects implementation details from 6 years ago.

These `using`s also interact poorly with modules: in a modularized world, either we have to put them in the global namespace, which makes them even more verbose (`using principia::quantities::Length` rather than `using quantities::Length`), or have to put them between the opening of the namespace and the beginning of an `export` block (meaning we cannot use `export namespace`).

There also can sometimes be an ambiguity between these `using`s that exist for convenience and those that are part of the API, like these in `cgs.hpp`:
https://github.com/mockingbirdnest/Principia/blob/b0bec96cae90105aafa39131f140449fd995936d/quantities/cgs.hpp#L18-L19

We want to switch to `using namespace` for those convenience `using`s, with smaller namespaces.

<table>
<thead>
<th>Current</th><th>This issue</th><th>Modularized future</th>
</thead>
<tbody>
<tr>
<td>

```C++
#include "other_project_name/other_file_name.hpp"
// […]
#include "yet_another_project_name/cætera.hpp"

namespace principia {
namespace project_name {
namespace internal_file_name {

using other_project_name::Something;
using other_project_name::SomethingElse;
// […]
using yet_another_project_name::Cætera;

// […]

}  // namespace internal_file_name

using internal_file_name::ExportedEntity;
// […]

}  // namespace project_name
}  // namespace principia
```
</td>
<td>

```C++
#include "other_project_name/other_file_name.hpp"
// […]
#include "yet_another_project_name/cætera.hpp"

namespace principia::project_name::file_name {
namespace internal {

using namespace principia::other_project_name::other_file_name;
// […]
using namespace principia::yet_another_project_name::cætera;

// […]

}  // namespace internal

using internal::ExportedEntity;
// […]

}  // namespace principia::project_name::file_name
```

</td>
<td>

```C++
export module principia.project_name.file_name

import principia.other_project.other_file_name;
// […]
import principia.yet_another_project_name.cætera;

using namespace principia::other_project_name::other_file_name;
// […]
[...]

PR #3531, opened 2023-02-25
Renamespace the base project

pleroy:
#3321.


base/optional_serialization.hpp
```diff
@@ -24,5 +26,13 @@ namespace base {
     }                                                               \
   }()
 
+}  // namespace internal
+
+
```
eggrobin, line 31: Remove these blank lines?

PR #3536, opened 2023-03-01
Consider the orbit of the Sun around the Earth

eggrobin:
We want to be using the same reference frame as for the orbit of the satellite, so we must look at the motion of the Sun that way.


astronomy/orbital_elements_test.cpp
```diff
@@ -452,11 +452,12 @@ TEST_F(OrbitalElementsTest, Years) {
...
   auto const status_or_elements =
-      OrbitalElements::OrbitalElements::ForTrajectories(
-          sun,
+      OrbitalElements::OrbitalElements::ForTrajectory(
           *ephemeris->trajectory(&sun),
+          BodyCentredNonRotatingDynamicFrame<ICRS, GCRS>(ephemeris.get(),
+                                                         &earth),
           earth,
-          *ephemeris->trajectory(&earth));
+          sun);
```
pleroy, line 460: Name the parameters at the call site.

PR #3579, opened 2023-04-01
Remove compatibility using directives in base

eggrobin:
#3321.


base/base64_test.cpp
```diff
@@ -7,6 +7,8 @@
 namespace principia {
 namespace base {
 
+using namespace principia::base::_array;
+
```
pleroy, line 11: Remove empty line.

base/hexadecimal_test.cpp
```diff
@@ -14,6 +14,8 @@ using ::testing::ElementsAre;
 namespace principia {
 namespace base {
 
+using namespace principia::base::_array;
+
```
pleroy, line 18: Same.

benchmarks/encoder.cpp
```diff
@@ -18,6 +18,9 @@ namespace principia {
 namespace base {
 
 using namespace principia::base::_array;
+using namespace principia::base::_hexadecimal;
```
pleroy, line 21: Alphabetize.

PR #3589, opened 2023-04-05
Fix more errors found by Clang

pleroy:


physics/similar_motion.hpp
```diff
@@ -17,6 +17,11 @@
 
 namespace principia {
 namespace physics {
+
+FORWARD_DECLARE_FROM(rigid_motion,
+                     TEMPLATE(typename FromFrame, typename ToFrame) class,
+                     RigidMotion);
```
eggrobin, line 23: Why the forward-declaration? We have it already, see the include l. 14.

PR #3602, opened 2023-04-12
Remove compatibility namespaces from integrators

pleroy:


integrators/embedded_explicit_generalized_runge_kutta_nyström_integrator_body.hpp
```diff
@@ -374,7 +373,7 @@ internal::EmbeddedExplicitGeneralizedRungeKuttaNyströmIntegrator<Method,
                                                                  ODE_> const&
 EmbeddedExplicitGeneralizedRungeKuttaNyströmIntegrator() {
   static_assert(
-      std::is_base_of<methods::EmbeddedExplicitGeneralizedRungeKuttaNyström,
+      std::is_base_of<_methods::EmbeddedExplicitGeneralizedRungeKuttaNyström,
```
eggrobin, line 376: Keep the using directive, drop the `_methods` here.

integrators/embedded_explicit_runge_kutta_integrator_body.hpp
```diff
@@ -367,7 +367,7 @@ template<typename Method, typename ODE_>
 internal::EmbeddedExplicitRungeKuttaIntegrator<Method, ODE_> const&
 EmbeddedExplicitRungeKuttaIntegrator() {
   static_assert(
-      std::is_base_of<methods::EmbeddedExplicitRungeKutta, Method>::value,
+      std::is_base_of<_methods::EmbeddedExplicitRungeKutta, Method>::value,
```
eggrobin, line 370: Same.

integrators/embedded_explicit_runge_kutta_nyström_integrator_body.hpp
```diff
@@ -363,7 +363,7 @@ template<typename Method, typename ODE_>
 internal::EmbeddedExplicitRungeKuttaNyströmIntegrator<Method, ODE_> const&
 EmbeddedExplicitRungeKuttaNyströmIntegrator() {
   static_assert(
-      std::is_base_of<methods::EmbeddedExplicitRungeKuttaNyström,
+      std::is_base_of<_methods::EmbeddedExplicitRungeKuttaNyström,
```
eggrobin, line 366: Same.

integrators/explicit_linear_multistep_integrator_body.hpp
```diff
@@ -214,13 +214,15 @@ template<typename Method, typename ODE_>
 internal::ExplicitLinearMultistepIntegrator<Method, ODE_> const&
 ExplicitLinearMultistepIntegrator() {
   static_assert(
-      std::is_base_of<methods::ExplicitLinearMultistep, Method>::value,
+      std::is_base_of<_methods::ExplicitLinearMultistep, Method>::value,
```
eggrobin, line 217: Same.

integrators/explicit_runge_kutta_integrator_body.hpp
```diff
@@ -219,7 +219,7 @@ WriteToMessage(
 template<typename Method, typename ODE_>
 internal::ExplicitRungeKuttaIntegrator<Method, ODE_> const&
 ExplicitRungeKuttaIntegrator() {
-  static_assert(std::is_base_of<methods::ExplicitRungeKutta, Method>::value,
+  static_assert(std::is_base_of<_methods::ExplicitRungeKutta, Method>::value,
```
eggrobin, line 222: Same.

integrators/symmetric_linear_multistep_integrator_body.hpp
```diff
@@ -358,15 +358,16 @@ template<typename Method, typename ODE_>
 internal::SymmetricLinearMultistepIntegrator<Method, ODE_> const&
 SymmetricLinearMultistepIntegrator() {
   static_assert(
-      std::is_base_of<methods::SymmetricLinearMultistep, Method>::value,
+      std::is_base_of<_methods::SymmetricLinearMultistep, Method>::value,
```
eggrobin, line 361: Same.

eggrobin (changes requested):
Drop the `_methods` from the actual code wherever possible.

integrators/embedded_explicit_generalized_runge_kutta_nyström_integrator_body.hpp
pleroy: No can do, at least not in this PR.

Note that the using directive that I removed was in the `internal` namespace, but here we are no longer in the `internal` namespace, we are in `_embedded_explicit_generalized_runge_kutta_nyström_integrator`.  If I put a using directive of `_methods` here, any client of the integrator will have direct visibility on *all* the methods.  We don't want that.

The proper fix is to move the implementation of this factory method into `internal`, where it would benefit from the using directives without creating leaks.  Unfortunately that's not that simple because the method has the same name as the class.  So it will probably have to go into `internal::factory` or some such to avoid the name clash.

A useful cleanup, but what do we say to the God of Feature Creep?  Not today.

eggrobin (approved):

PR #3609, opened 2023-04-17
A test for the rotating-pulsating reference frame

eggrobin:
And a new integration method for good measure, which did not end up being used.

The errors are far too big, but the expected digits do not change when the precision is refined, nor do they change when replacing the rotating-pulsating frame with the body-centred-body-direction frame, so the code under test is not to blame for that.


eggrobin:
retest this please

PR #3610, opened 2023-04-18
Check for “remove before flight” comments

eggrobin:
It seems that the workflow doesn’t run for the pull request that adds it, so I cannot test the part that checks the pull request title and description.


pleroy:
retest this please

PR #3613, opened 2023-04-20
Update Build Doc

KingParity:
Updated documentation to include creating the folders. I found that otherwise an error happens when trying to build.


eggrobin:
We should just do that in `rebuild_all_solutions` instead of piling on more manual steps. #3614.

PR #3614, opened 2023-04-20
Ensure we have Google and Third Party directories in running rebuild_all_solutions

eggrobin:


pleroy:
Amusingly I ran into this recently when configuring a new laptop.

PR #3617, opened 2023-04-24
A new frame

eggrobin:
Translations will come in a future PR.

The X-Y Lagrange frame can be selected as a plotting frame, but not as a manœuvre frame. Speeds on the navball and at the nodes are displayed in current metres per second, not in constant lunar distances (per second nor per month).

This is the frame where we will want to do something about #3358, which will make it very different from the rigid counterpart; for now, EML is just a much weirder slight variant of MEO.


ksp_plugin/interface.hpp
```diff
@@ -173,6 +173,11 @@ XYZ ToXYZ(Bivector<AngularMomentum, World> const& angular_momentum);
 Instant FromGameTime(Plugin const& plugin, double t);
 double ToGameTime(Plugin const& plugin, Instant const& t);
 
+// A factory for PlottingFrame objects.
+not_null<std::unique_ptr<PlottingFrame>> NewPlottingFrame(
```
pleroy, line 177: Alphabetize, move after `NewNavigationFrame`.

ksp_plugin/plugin.cpp
```diff
@@ -1180,6 +1180,16 @@ Plugin::NewBodySurfaceNavigationFrame(
       ephemeris_.get(), reference_body.body());
 }
 
+not_null<std::unique_ptr<PlottingFrame>>
+Plugin::NewRotatingPulsatingPlottingFrame(Index primary_index,
+                                          Index secondary_index) const {
```
pleroy, line 1185: `const` the parameters.

ksp_plugin/interface_body.hpp
```diff
@@ -631,6 +631,23 @@ inline double ToGameTime(Plugin const& plugin,
   return (t - plugin.GameEpoch()) / Second;
 }
 
+inline not_null<std::unique_ptr<PlottingFrame>> NewPlottingFrame(
```
pleroy, line 634: Same as in the other file, move down after the next method.

ksp_plugin/renderer.cpp
```diff
@@ -147,17 +150,20 @@ Renderer::RenderPlottingTrajectoryInWorld(
       from_plotting_frame_to_world_at_current_time =
           PlottingToWorld(time, sun_world_position, planetarium_rotation);
   for (auto it = begin; it != end; ++it) {
-    auto const& [time, degrees_of_freedom] = *it;
+    auto const& [t, degrees_of_freedom] = *it;
```
pleroy, line 153: Would a range-for loop work here?

ksp_plugin_adapter/burn_editor.cs
```diff
@@ -61,12 +61,23 @@ class BurnEditor : ScalingRenderer {
...
-        adapter_.plotting_frame_selector_.FrameParameters());
+    PlottingFrameParameters plotting_frame_parameters =
+        adapter_.plotting_frame_selector_.FrameParameters();
+    if ((NavigationFrameParameters?)plotting_frame_parameters is
+            NavigationFrameParameters parameters) {
+      reference_frame_selector_.SetFrameParameters(parameters);
+    } else {
+      reference_frame_selector_.SetFrameParameters(
+          new NavigationFrameParameters {
+            extension = (int)FrameType.BODY_CENTRED_PARENT_DIRECTION,
```
pleroy, line 76: Can you comment what's happening here?

ksp_plugin_adapter/ksp_plugin_adapter.cs
```diff
@@ -2398,7 +2401,7 @@ out DisposableIterator
   }
 
   private void UpdatePlottingFrame(
-      NavigationFrameParameters? frame_parameters,
+      PlottingFrameParameters? frame_parameters,
```
pleroy, line 2404: Would fit on the previous line?

ksp_plugin_adapter/reference_frame_selector.cs
```diff
@@ -61,7 +54,7 @@ public enum FrameType {
     });
   }
 
-  public void SetFrameParameters(NavigationFrameParameters parameters) {
+  public void SetFrameParameters(Parameters parameters) {
```
pleroy, line 57: Why is this not `ReferenceFrameParameters`?

ksp_plugin_adapter/reference_frame_selector.cs
```diff
@@ -394,7 +434,10 @@ public enum FrameType {
   public bool FixesBody(CelestialBody celestial) {
     // TODO(egg): When we have the rotating-pulsating frame, this should return
```
pleroy, line 435: Remove the TODO?

ksp_plugin_adapter/localization/en-us.cfg
```diff
@@ -57,11 +57,13 @@ Localization {
     #Principia_ReferenceFrameSelector_Name_BodyCentredNonRotating = <<1>>-Centred Inertial
     #Principia_ReferenceFrameSelector_Name_BodyCentredParentDirection = <<1>>-<<3>>-Orbit
     #Principia_ReferenceFrameSelector_Name_BodySurface = <<1>>-Centred <<1>>-Fixed
+    #Principia_ReferenceFrameSelector_Name_RotatingPulsating = <<3>>-<<1>> Lagrange
 
     #Principia_ReferenceFrameSelector_Abbreviation_Target.en-us = Tgt <<2>>O
     #Principia_ReferenceFrameSelector_Abbreviation_BodyCentredNonRotating.en-us = <<2>>CI
     #Principia_ReferenceFrameSelector_Abbreviation_BodyCentredParentDirection.en-us = <<2>><<4>>O
     #Principia_ReferenceFrameSelector_Abbreviation_BodySurface.en-us = <<2>>C<<2>>F
+    #Principia_ReferenceFrameSelector_Abbreviation_RotatingPulsating.en-us = <<4>><<2>>L
```
pleroy, line 66: Not the same order as BCBD.  Is that intended?

eggrobin: Yes.

ksp_plugin/renderer.cpp
eggrobin: Yes, though we need `base::Range`. At some point we should start using `std::ranges`, and then this would be `std::ranges::subrange` (or rather, this should take a range instead of a pair of iterators).

eggrobin:
retest this please

eggrobin:
retest this please

PR #3624, opened 2023-05-01
Fix a typo in the explanations related to the triple product

pleroy:


documentation/Rotational Motion.tex
```diff
@@ -106,7 +106,7 @@ \section*{Angular velocity}
 \dot{\vb} &= \ga \vf \wedge \vb + \gb \vn \wedge \vb
 \end{dcases}
 \]
-Now multiply (using the scalar product) the first equation by $\vn$, the second by $\vb$ and the third by $\vf$, and use the properties of the triple product to simplify the result using $\scal{(\vn \wedge \vf)}{\vn} = 0$ and $\scal{(\vn \wedge \vf)}{\vb} = 1$ (and similar expressions obtained by circular permutation).  We find:
+Now multiply (using the scalar product) the first equation by $\vn$, the second by $\vb$ and the third by $\vf$, and use the properties of the triple product to simplify the result using $\scal{(\vn \wedge \vf)}{\vn} = 0$ and $\scal{(\vb \wedge \vf)}{\vn} = 1$ (and similar expressions obtained by circular permutation).  We find:
```
eggrobin, line 109: Let’s say (f∧n)⋅b = 1, so it matches what we have when we distinguish the multivectors.

Eventually we fix the rest of the document to be more consistent with the code, but not today.

eggrobin (approved):

documentation/Rotational Motion.tex
pleroy: Nay, this explanation matches what happens for the first equation in the system immediately above.  Let's keep things straightforward, if we ever want to fix the document to be consistent with the code, we might as well rewrite this expression in addition to all the rest.

PR #3625, opened 2023-05-01
Use the jerk in the computation of rotational motion

pleroy:
Also add a function to compute the jerk on a massless body in `Ephemeris`.


eggrobin (approved):

pleroy:
retest this please

PR #3627, opened 2023-05-04
Draw another equipotential to pinpoint L₄/L₅

eggrobin:
#3358.
image


ksp_plugin/geometric_potential_plotter.cpp
```diff
@@ -204,11 +213,14 @@ absl::Status GeometricPotentialPlotter::PlotEquipotentials(
                                 std::make_move_iterator(lines.end()));
   }
   if (parameters.show_l2_level) {
```
pleroy, line 215: Should we rename this parameter?  Should we have a different one for l45?

PR #3639, opened 2023-05-14
Pulsate the camera distance in the rotating-pulsating reference frame

eggrobin:
This means that when the camera targets a fixed point in the reference frame (which can be a bit tricky since the interesting part may be the barycentre of multiple bodies; we should explore camera targeting later), the trajectories are fixed on the screen, and instead the bodies grow and shrink as expected.

### DIFF
--- a/ksp_plugin_adapter/planetarium_camera_adjuster.cs
+++ b/ksp_plugin_adapter/planetarium_camera_adjuster.cs
@@ -21,7 +21,10 @@ public class PlanetariumCameraAdjuster : UnityEngine.MonoBehaviour {
     }
     var reference_rotation =
         (UnityEngine.QuaternionD)adapter.Plugin().CameraReferenceRotation();
+    double scale = adapter.Plugin().CameraScale();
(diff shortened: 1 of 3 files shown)


ksp_plugin_adapter/planetarium_camera_adjuster.cs
```diff
@@ -21,7 +21,10 @@ public class PlanetariumCameraAdjuster : UnityEngine.MonoBehaviour {
     }
     var reference_rotation =
         (UnityEngine.QuaternionD)adapter.Plugin().CameraReferenceRotation();
+    double scale = adapter.Plugin().CameraScale();
     if (should_transfer_camera_coordinates) {
+      // Preserve the position when the plotting frame changes.
```
pleroy, line 26: s/position/scale/

eggrobin: No, the position. We preserve the position by causing the scale correction to be 1.

Issue #3632, opened 2023-05-08
Add support for arrays of structures in interchange messages


eggrobin:
#3631 uses a string to interchange a sequence of integers.

PR #3656, opened 2023-05-20
Support for repeated fields in interchange messages

pleroy:
Also added support for more scalar types, as it would be annoying to have to modify the tool if/when we need them.

First part of #3632 (second part is the TODO that's left in `journal.proto`).


ksp_plugin_adapter/boxed.cs
```diff
@@ -30,6 +30,22 @@ internal class BoxedInt32 : Boxed<int> {
...
 
+internal class BoxedInt64 : Boxed<long> {
+  public static implicit operator BoxedInt64(long all) {
+    return new BoxedInt64(all);
+  }
+
+  protected BoxedInt64(long all) : base(all) {}
+}
+
+internal class BoxedUint32 : Boxed<uint> {
```
eggrobin, line 41: UInt32, like https://learn.microsoft.com/en-us/dotnet/api/system.uint32?view=net-7.0. Everywhere.

eggrobin (approved):

Issue #3483, opened 2022-12-24
Computation failed with error 4 when extending flight plan


Test-Account-KSP:
Error message "Computation failed with error 4" appears when I extend a flight plan similar to the usual TLI maneuver to 10-20 days.
Steps to replicate: 
- enter the VAB, simulate any craft in 200km orbit, 
- create flight plan with max steps 4096, tolerance 10m, plan length 4 days
- add maneuver with t initial = 1h 1m, dV = 3150 prograde, inertially fixed, instant impulse
- extend flight plan from 4 days to 14 days or similar (entered 14 instead of 4 in the field, without moving the slider) 

![screenshot](https://user-images.githubusercontent.com/110846612/209436039-9ed4c327-fd80-4ac6-a70c-cf5ab668467e.png)

[persistent.zip](https://github.com/mockingbirdnest/Principia/files/10298550/persistent.zip)

[INFO.20221224-112749.6464.log](https://github.com/mockingbirdnest/Principia/files/10298551/INFO.20221224-112749.6464.log)

[WARNING.20221224-112749.6464.log](https://github.com/mockingbirdnest/Principia/files/10298552/WARNING.20221224-112749.6464.log)

[ERROR.20221224-112749.6464.log](https://github.com/mockingbirdnest/Principia/files/10298553/ERROR.20221224-112749.6464.log)

Another screenshot (different craft, same issue)

![screenshot19](https://user-images.githubusercontent.com/110846612/209436388-090fa457-1e2a-4570-8868-a96c5abd2b6d.png)

pleroy:
We are running into this limit: https://github.com/mockingbirdnest/Principia/blob/c284c2a9a9c07df3b2a0c3238cee67b124b3b96a/ksp_plugin/flight_plan.hpp#L131
The right fix is probably to make the computation of the flight plan asynchronous, but that's a bit of work.

PR #3658, opened 2023-05-21
Try to emit a message slightly more useful than "error 4"

pleroy:
It is really hard to explain to the user what they should do in that case.  "Just tickle the flight plan length" is not going to be helpful.  I am not proud of the wording, but it's probably an improvement over the current situation.

This will need to be localized in Chinese and Russian.

Not a fix of #3483.

### DIFF
--- a/ksp_plugin_adapter/localization/zh-cn.cfg
+++ b/ksp_plugin_adapter/localization/zh-cn.cfg
@@ -212,6 +212,8 @@ Localization {
     #Principia_FlightPlan_CoastInOrbit = 停泊于 <<1>> 还需等待 <<2>>
     #Principia_FlightPlan_AddManœuvre = 新建轨道机动
     #Principia_FlightPlan_StatusMessage_ChangeFlightPlan = 请更换轨道规划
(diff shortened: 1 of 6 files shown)


eggrobin (approved):

ksp_plugin_adapter/localization/zh-cn.cfg
```diff
@@ -212,6 +212,8 @@ Localization {
     #Principia_FlightPlan_CoastInOrbit = 停泊于 <<1>> 还需等待 <<2>>
     #Principia_FlightPlan_AddManœuvre = 新建轨道机动
     #Principia_FlightPlan_StatusMessage_ChangeFlightPlan = 请更换轨道规划
+    #Principia_FlightPlan_StatusMessage_DeadlineExceeded = flight plan exceeds the period over which celestials were computed
+    #Principia_FlightPlan_StatusMessage_Tickle = repeatedly increasing and shortening the flight plan a bit
```
al2me6, line 216:
```suggestion
    #Principia_FlightPlan_StatusMessage_DeadlineExceeded = 轨道规划长度超过天体计算长度
    #Principia_FlightPlan_StatusMessage_Tickle = 反复延长并缩短规划长度
```

PR #3684, opened 2023-06-17
A tool to implement include-what-you-using and its application to geometry

pleroy:
Note the use of `// 🧙` for includes that must be preserved by the sourcerer.


sourcerer/parser.cs
```diff
@@ -214,16 +216,67 @@ public class Function : Declaration {
...
     }
 
     public override void WriteNode(string indent = "") {
       Console.WriteLine(indent + "Include " + string.Join(", ", path));
     }
 
     public string[] path;
+
+    public bool is_blessed_by_sourcerer =>
+        text.EndsWith("// 🧙");
```
eggrobin, line 251: We should be able to add an explanation. Technically we can do that by having
`// Needed for XYZ // 🧙`,
but maybe EndsWith("🧙") would be enough (alternatively, a regex so it can be at the beginning of the comment.

geometry/conformal_map_body.hpp
```diff
@@ -2,6 +2,7 @@
 
 #include "geometry/conformal_map.hpp"
 
+#include "geometry/homothecy.hpp"  // 🧙
```
eggrobin, line 5: Why? (and the same question in many other places.)

sourcerer/parser.cs
pleroy: Done.

eggrobin (approved):

PR #3692, opened 2023-06-18
Interactive manœuvre markers

al2me6:
> egg: [...] What we don't know is how to mesh draggy.

Draw the Frenet markers as meshes with colliders, which allows user-interaction.

In particular,

- Once created, manœuvre markers may be dragged to move them in time.
- Further interactions, including hover labels and click-to-open-burn-editor, will be implemented in future PRs.

UT sliding is implemented by linearly approximating the trajectory at the starting point of the burn. This works well enough in most simple cases, but has convergence issues if the mouse is too far away from the marker, whereby the marker will oscillate about the 'target' position. Future improvements here are warranted, but for the time being this can be avoided by dragging more slowly.

A demonstration: 

image


ksp_plugin_adapter/manœuvre_marker.cs
```diff
@@ -0,0 +1,210 @@
...
+    var screen_tangent =
+        (ScaledToFlattenedScreenPosition(transform.position + tangent_vector_) -
+         ScreenManœuvrePosition()).normalized;
+
+    var projection = UnityEngine.Vector3.Dot(screen_tangent, drag);
+    var Δut = projection *
+              UnityEngine.Mathf.Exp(-mouse_offset_now.sqrMagnitude * 5e-3f);
+    var burn = get_burn_();
+    burn.initial_time += Δut;
+    modify_burn_(burn);
```
eggrobin, line 132: As mentioned in discussion, this is the part that we should rework to address the stability issues alluded to in the PR description.

My initial thought was that we should be able to minimize the distance to the cursor over the trajectory, and change the time of the burn to the time of that minimum.
Of the local minima found, we need to pick the one closest to the current time of the burn, since there may be multiple revolutions overlapping.

However, we don’t actually have the relevant trajectory: it would be the coast prior to the burn, prolonged past the burn without the burn.

al2me6: FWIW I'm not personally convinced if that level of generality is needed to support the functionality demonstrated above. My thought was that by only allowing interactivity with a preexisting marker, we can get away _without_ searching the entire trajectory, since the problem is reduced to movingly continuously along a continuous curve. Considering the functionality of the kludge implemented above, I'm fairly certain that considering the velocity in line 128 above would result in a nearly indistinguishable user experience compared to full optimization, while being much, much simpler to implement and execute.

But then, Principia isn't exactly known for taking the easy way out, so what am I saying 😆 With optimization, I'd imagine it becomes possible to support hovering along the entirety of the trajectory, stock-style.

eggrobin: > we can get away without searching the entire trajectory, since the problem is reduced to movingly continuously along a continuous curve.

The continuity is an interesting point. What really worries me here is the kludge at ll 118–121, but perhaps we can get rid of it while still benefiting from continuity.

> With optimization, I'd imagine it becomes possible to support hovering along the entirety of the trajectory, stock-style.

This would be a different interaction which would require different UI wizardry; but searching along the flight plan for the nearest point to _create_ a manœuvre would in a sense be easier, because then we can just look at the real flight plan trajectory.

al2me6: > What really worries me here is the kludge at ll. 118–121

TBH I'm not sure if it's necessary. I added it while debugging another issue (due to bad math). In fact, I'm not sure if either of the manipulations I'm doing here (that kludge and the exponential decay) are necessary.

ksp_plugin_adapter/map_node_pool.cs
```diff
@@ -45,11 +45,12 @@ public class SingleProvenancePool {
         new List<KSP.UI.Screens.Mapview.MapNode>(MaxNodesPerProvenance);
   }
 
-  public MapNodePool(Func<bool> show_unpinned) {
+  public MapNodePool(Func<bool> show_unpinned, Func<bool> can_hover) {
```
pleroy, line 48: The addition of a 2nd callback here feels a bit clunky but I am not sure what to suggest instead.  Let's see what egg has to say.

ksp_plugin_adapter/flight_planner.cs
```diff
@@ -406,6 +403,14 @@ class FlightPlanner : VesselSupervisedWindowRenderer {
     }
   }
 
+  internal void ModifyIthBurn(string vessel_guid, int i, Burn burn) {
+    var status = plugin.FlightPlanReplace(vessel_guid, burn, i);
+    UpdateStatus(status, i);
+    if (burn_editors_?.Count > 0) {
```
pleroy, line 409: To be defensive, this test should probably check that there is actually a burn editor at index `i`.

ksp_plugin_adapter/ksp_plugin_adapter.cs
```diff
@@ -2170,59 +2176,73 @@ select PartCentredForceHolder.
...
+                              return plugin_.FlightPlanGetManoeuvre(
+                                  main_vessel_guid,
+                                  manœuvre_index);
+                             },
+                             modify_burn: (burn) => {
+                              flight_planner_.ModifyIthBurn(main_vessel_guid,
+                                                            manœuvre_index,
+                                                            burn);
+                             });
+                  number_of_rendered_manœuvres += 1;
```
pleroy, line 2235: `++number_of_rendered_manœuvres;`

pleroy (commented):
To get the ball rolling, I made a first pass to catch minor stylistic issues.  I don't have a very good understanding of this code, so @eggrobin will need to make another pass.

ksp_plugin_adapter/map_node_pool.cs
al2me6: A single callback returning a 2-tuple/struct?

ksp_plugin_adapter/flight_planner.cs
```diff
@@ -406,6 +403,14 @@ class FlightPlanner : VesselSupervisedWindowRenderer {
     }
   }
 
+  internal void ModifyIthBurn(string vessel_guid, int i, Burn burn) {
```
eggrobin, line 406: The class `FlightPlanner` is associated with a vessel (it is a `VesselSupervisedWindowRenderer`, so we should not pass a vessel—that would be an easy way to mess with the wrong vessel).
Use `predicted_vessel?.id.ToString()` instead.

The name `Ith` is weird unless you use named parameters, which you do not.
Rename to `ReplaceBurn(burn, i)`, so it looks like `FlightPlanReplace` (which it wraps, with additional housekeeping to keep the UI in sync).

ksp_plugin_adapter/manœuvre_marker.cs
```diff
@@ -0,0 +1,225 @@
...
+
+    Disable();
+  }
+
+  // Call on each frame (at or later than |Update|) to set the state of the marker.
+  public void Render(Vector3d world_position,
+                     Vector3d initial_plotted_velocity,
+                     NavigationManoeuvreFrenetTrihedron trihedron,
+                     Func<NavigationManoeuvre> get_manœuvre,
+                     Action<Burn> modify_burn) {
```
eggrobin, line 49: Could we pass the flight planner and the index instead of two delegates?

eggrobin (approved):

PR #3716, opened 2023-07-09
Additional interactions for manœuvre markers

al2me6:
As alluded to in #3692, this PR adds the following functionalities to manœuvre markers:

* A pinnable hover tooltip is displayed when the cursor is hovered over the marker, emulating stock node markers. When manœuvre markers overlap node markers, manœuvre markers will take precedence for hovering. However, the stock node markers have slightly larger colliders, so it remains possible to interact with node markers that are otherwise obscured.
* When a marker is single-clicked, the flight planner will open to the corresponding burn editor.

A new localization string has been introduced, and requires translations for fr and ru.

### DIFF
--- a/ksp_plugin_adapter/manœuvre_marker.cs
+++ b/ksp_plugin_adapter/manœuvre_marker.cs
@@ -11,6 +11,12 @@ internal class ManœuvreMarker : UnityEngine.MonoBehaviour {
   // the cursor may move off the marker while still dragging.
   public bool is_interacting => is_hovered || is_dragged;
 
+  // As mouse events are sent before |Update|, we compute this state there.
+  // We clear it as late as possible, in |WaitForEndOfFrame|.
+  // In particular, the value is meaningful at |BetterLateThanNeverLateUpdate|,
+  // during which it is consumed by node markers.
+  public static bool has_interacting_marker { get; private set; }
+
   // Construct the geometry of the marker when instantiated.
   // 1. Build the mesh of the marker (base 'bulb' & one cylinder for each axis).
   // 2. Attach a spherical collider (centred on the bulb) to the gameobject for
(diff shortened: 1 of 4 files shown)


ksp_plugin_adapter/manœuvre_marker.cs
```diff
@@ -147,14 +213,39 @@ internal class ManœuvreMarker : UnityEngine.MonoBehaviour {
...
     is_hovered = false;
   }
 #endregion Mouse Events
 
+  // Accumulate and reset the interactivity states of all active markers.
+  public void Update() {
+    has_interacting_marker |= is_interacting;
+
+    // Only run one copy of this coroutine.
+    if (index_ == 0) {
```
al2me6, line 239: This will probably break if the first manœuvre in the plan is 'anomalous'. However, I'm not sure what that means and when that might occur.

pleroy: Manœuvres are not anomalous per se.  Segments are the root cause of anomalies.  A segment (which may be a coast or a burn) is anomalous iff we fail to integrate it, for instance because we reached the maximum number of steps without reaching the desired end time (typically this would happen if the segment goes very close to a singularity).  When that happens, all the manœuvres that follow the first anomalous segment are themselves marked as anomalous.  If a manœuvre is later changed by the user so that integration succeeds, segments (and manœuvres) that were marked as anomalous can become normal again.

Long ago, we used to delete the manœuvres that followed the first segment where integration failed.  This was very obnoxious: say that you prepare a flight plan with 3 manœuvres, you tweak the first one and then the following coast goes close to the centre of a celestial for a second and, lo and behold, you've lost the last 2 manœuvres.  Anomalous segments and manœuvres were added to avoid losing these manœuvres.

al2me6: Thanks. In that case, the above snippet should be sound, as if the first manœuvre is anomalous and not drawn, subsequent manœuvres will not be drawn either.

pleroy (commented):
A couple of stylistic nits.  @eggrobin should be able to do a proper review this week.

Note that this PR has merge conflicts, it would be good to resolve them.

eggrobin (approved):

PR #3730, opened 2023-07-19
Don't show manœuvre markers in flight

### LINKED ISSUE #3729: Maneuver marker renders in flight scene
After creating flight plan and maneuver in map scene and switching to flight scene marker is still visible.

al2me6:
Fix #3729.


pleroy:
Thanks for the quick fix.

PR #3746, opened 2023-08-19
Optimize the flight plan optimizer

pleroy:
1. Use a cache to avoid evaluating the flight plan multiple times with the same burn.
2. Use the Gateaux derivative instead of computing the full gradient when possible (i.e., in the line search).  This requires changing a bit the use of the gradient descent to pass `HomogeneousArgument`s instead of `Argument`s.
3. Add a callback executed when a flight plan is evaluated and use it to count the number of evaluations of the flight plan.

Overall, the number of evaluations is reduced by a factor of 2-2.5, although in some cases the use of the Gateaux derivative can increase the cost a bit:
|Test|Manœuvre #|Options|Evaluations|
|---|---|---|---|
|ReachTheMoon|5|Original|164|
|||Cache|102|
|||Cache+Gateaux|88|
||6|Original|130|
|||Cache|89|
|||Cache+Gateaux|57|
||7|Original|127|
|||Cache|101|
|||Cache+Gateaux|47|
|GrazeTheMoon|5|Original|160|
|||Cache|70|
|||Cache+Gateaux|79|
||6|Original|132|
|||Cache|70|
|||Cache+Gateaux|72|
||7|Original|174|
|||Cache|95|
|||Cache+Gateaux|74|

### DIFF
--- a/t.txt
+++ b/t.txt
@@ -1,1 +1,0 @@
-[       OK ] FlightPlanOptimizerTest.DISABLED_GrazeTheMoon (497798 ms)
--- a/t.txt
+++ b/t.txt
@@ -1,0 +1,1 @@
+[       OK ] FlightPlanOptimizerTest.DISABLED_GrazeTheMoon (497798 ms)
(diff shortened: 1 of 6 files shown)


t.txt
eggrobin, line 1: Remove this file.

eggrobin (changes requested):

eggrobin (approved):

Issue #3681, opened 2023-06-13
Optimization in the flight plan


eggrobin:
We used to consider that laughably out of scope, but since @pleroy ended up spending a few months implementing optimization algorithms for #3358 (see https://github.com/mockingbirdnest/Principia/blob/ca2948a74840c8c0416af83efce8f7ebebc6d430/numerics/gradient_descent.hpp, https://github.com/mockingbirdnest/Principia/blob/ca2948a74840c8c0416af83efce8f7ebebc6d430/numerics/global_optimization.hpp), this may now be feasible.

See [discussion](https://discord.com/channels/319857228905447436/480397772248580098/1118084794224476221) with @FioraAeterna and @Test-Account-KSP on Discord:

> [09:49]Fiora: is there an addon or technique or such that makes tweaking interplanetary trajectories less of a nightmare?
coming up with the basic trajectory is easy enough but tweaking them /precisely/ so that you actually get to near orbit feels like it takes 30 minutes of sliding bars around 
[09:51]egg: Note that you can scroll over the digits, which can be nicer than fiddling with the bars when things get laggy.
[09:51]egg: Also you can use the up/down arrows with your cursor in a field.
[09:55]egg: (And you can have your cursor in one field and scroll over another field to more efficiently do your manual gradient descent in 2d. At the end of the day you are doing gradient descent by hand, which is a bit silly. I guess now that we have optimization algorithms in Principia we could try helping with that, though we would have to design some kind of UI to say what you are trying to optimize, and within which constraints…) 
[09:56]test_account: Optimizing impact with a celestial would be a great feature 🙂
[09:56]Fiora: it gets especially tricky when you have to change two directions simultaneously in order to get closer
sometimes gradient descent feels like it bottoms out completely and i have to make a second node later along the path and then try to tweak that, etc
[09:57]Fiora: like, its all doable. its just grindy and annoying
[09:57]egg: Principia: Teaching people that optimization is hard since 2014 :-p
[09:58]test_account: I can imagine it... tell the optimizer to vary only maneuver time and prograde dV, and which celestial to impact, and it finds the trajectory with lowest dV that leads to impact. 
[09:59]Fiora: honestly that alone would be amazing. just an impact optimizer. nothing else. once i have an impact trajectory i can do everything from there
[…]
[10:06]Fiora: one thing that would be amazing about an impact optimizer is that it would also generate automatic course corrections by its very nature

PR #3748, opened 2023-08-26
Add a copy constructor for the FlightPlan and use it in Vessel

pleroy:
#3681.

### DIFF
--- a/ksp_plugin/vessel.cpp
+++ b/ksp_plugin/vessel.cpp
@@ -1165,7 +1165,7 @@ bool Vessel::IsCollapsible() const {
     }
   }
   CHECK_NE(nullptr, containing_pile_up);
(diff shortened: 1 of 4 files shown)


ksp_plugin/vessel.cpp
```diff
@@ -1170,7 +1165,7 @@ bool Vessel::IsCollapsible() const {
     }
   }
   CHECK_NE(nullptr, containing_pile_up);
-  for (const auto part : containing_pile_up->parts()) {
+  for (const auto& part : containing_pile_up->parts()) {
```
eggrobin, line 1168:
```suggestion
  for (auto const& part : containing_pile_up->parts()) {
```

eggrobin (approved):

PR #3749, opened 2023-08-26
Update on sol_initial_state_jd_2433282_500000000.cfg

StarshipSN20:
Uses the most up-to-date data from JPL Horizons.
JPL Horizons settings: Solar System Barycenter, 1950/1/1/00:00-1950/1/1/00:01, steps = 60min. Vector table settings are by default.


eggrobin:
> Vector table settings are by default.

Well, that is wrong, since the default is with respect to the ecliptic, rather than the ICRF axes (and that explains why all the z coordinates are massively changing).

Issue #3789, opened 2023-11-01
Game Crash sometimes when using Optimizer


Ballatik:
Testing the maneuver optimizer button, have attempted 4 lunar transfers, 3 venus transfers, and 3 venus correction burns.  Of these, 3 have caused crashes.  One was likely user error in attempting to optimize a maneuver that had no dV set yet, but the other two were already reasonably close to their destinations.  I did not see any obvious differences between the maneuvers that worked vs. those that caused crashes.
Worth noting, the time saved fiddling with minute corrections in maneuvers is already noticeably more than the time lost to these crashes and restarts, so thank you.
[FATAL.20231030-071205.17436.log](https://github.com/mockingbirdnest/Principia/files/13232091/FATAL.20231030-071205.17436.log)
[FATAL.20231030-210628.7056.log](https://github.com/mockingbirdnest/Principia/files/13232092/FATAL.20231030-210628.7056.log)
[FATAL.20231101-075002.7576.log](https://github.com/mockingbirdnest/Principia/files/13232093/FATAL.20231101-075002.7576.log)
[INFO.20231029-191309.17436.log](https://github.com/mockingbirdnest/Principia/files/13232094/INFO.20231029-191309.17436.log)
[INFO.20231030-204510.7056.log](https://github.com/mockingbirdnest/Principia/files/13232095/INFO.20231030-204510.7056.log)
[INFO.20231101-063056.7576.log](https://github.com/mockingbirdnest/Principia/files/13232096/INFO.20231101-063056.7576.log)

pleroy:
Thanks for testing this version and for reporting issues.

I was able to reproduce and fix two of the crashes (`Check failed: index < number_of_manœuvres() (0 vs. 0)` and `Check failed: 'driver' Must be non NULL`).  To avoid them while the new release is not out, you just need, immediately after adding or removing a manoeuvre, to change a bit the optimization parameters.  Any change, no matter how small, should avoid the crashes.

PR #3790, opened 2023-11-01
The optimizer holds a copy of the flight plan so it must be recreated whenever burns get added/deleted.

pleroy:
That's in addition to being recreated when the optimization parameters change.

Fixes parts of #3789.

### DIFF
--- a/ksp_plugin_adapter/burn_editor.cs
+++ b/ksp_plugin_adapter/burn_editor.cs
@@ -10,12 +10,10 @@ class BurnEditor : ScalingRenderer {
   public BurnEditor(PrincipiaPluginAdapter adapter,
                     Vessel vessel,
                     double initial_time,
-                    int index,
                     Func<int, BurnEditor> get_burn_at_index) {
     adapter_ = adapter;
     vessel_ = vessel;
     initial_time_ = initial_time;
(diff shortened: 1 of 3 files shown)


ksp_plugin_adapter/burn_editor.cs
```diff
@@ -10,12 +10,10 @@ class BurnEditor : ScalingRenderer {
   public BurnEditor(PrincipiaPluginAdapter adapter,
                     Vessel vessel,
                     double initial_time,
-                    int index,
                     Func<int, BurnEditor> get_burn_at_index) {
     adapter_ = adapter;
     vessel_ = vessel;
     initial_time_ = initial_time;
-    this.index = index;
```
eggrobin, line 18: Where is this set now?

eggrobin (approved):

ksp_plugin_adapter/burn_editor.cs
pleroy: As discussed offline, every construction of a `BurnEditor` in the flight planner should be followed by a call to `UpdateAfterBurnEditorInsertionDeletion` which sets the index.  This a more systematic API than setting the index sometimes at construction and sometimes later.

PR #3800, opened 2023-11-09
Do not fix unwinding of classical elements

pleroy:
Do not fix anything, we believe that it's just a singularity.  Add a new `ToMathematica` utility.


eggrobin (changes requested):
The test failures seem to say that this does not put the elements in the ranges we expect.

pleroy:
> The test failures seem to say that this does not put the elements in the ranges we expect.

So these tests are actually useful?  Fixed.

eggrobin (approved):

Issue #3230, opened 2021-12-16
High CPU usage for updating parts in active vessel


StonesmileGit:
High part count vessels have a high performance impact while being loaded.

Unity profiler indicates Principia uses approximately 40% of the CPU cycles on a 265 part vessel.

Journal of 265 part vessel in a 200 km orbit doing nothing: [link to Drive](https://drive.google.com/file/d/1s6xaU-baxIOtD9-eMPCOmwVECQx5pLjL/view?usp=sharing)

pleroy:
I finally replayed this journal, and unfortunately it's not all that interesting and doesn't reveal any C++ path that would have a cost growing with the number of parts.  This journal does mostly three things:
1. It serializes and deserializes the Principia plugin, maybe because of scene changes.
2. It recomputes the past trajectories of celestials after deserialization.
3. It computes the future trajectories of vessels, mostly for the purposes of orbit analysis.

All of this has a cost proportional to the number of vessels/celestials, not to the number of parts.  It is actually quite hard to find anything in the profile that talks about parts (that would be `FreeVesselsAndPartsAndCollectPileUps`) and even so its cost is around 0.3% of the total.

It's conceivable that the performance hit would come from the C# code and would not be visible on the journal, but to analyze that we would need a save obtained with the stock game with instructions to reproduce.  Alternatively, it's possible that there is something wrong with the C++ code, but then we would need a much "cleaner" journal without scene changes, orbit analysis and the like.

I am noticing though that locking/unlocking in `ContinuousTrajectory` is somewhat expensive.  I'll open a separate issue for this.

StonesmileGit:
Short Journal created with stock KSP 1.12.3 and Principia Ἵππαρχος.
Journal is a 1167 parts craft in low Kerbin orbit doing nothing. [link to drive](https://drive.google.com/drive/folders/1ap8XHteVmZ3v2AJyKWIfnrkf-SEZ2Uhe)
The journal box was checked in the tracking station before switching to the vessel.
Craft is https://kerbalx.com/BullShootLatinName/SAAB-J35-Draken
Save file: [link to drive](https://drive.google.com/drive/folders/1ap8XHteVmZ3v2AJyKWIfnrkf-SEZ2Uhe)

pleroy:
I'll need access to these files in Google Drive.

StonesmileGit:
Woops, sorry! Should have access now!

gotmachine:
[...]

PR #3803, opened 2023-11-19
Optimize the code in the adapter

pleroy:
With the vessel in #3230, which has 1167 parts, the time spent in `WaitedForFixedUpdate` goes down from 183 ms to 54 ms.

### DIFF
--- a/ksp_plugin_adapter/ksp_plugin_adapter.cs
+++ b/ksp_plugin_adapter/ksp_plugin_adapter.cs
@@ -2,8 +2,7 @@
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
-using KSP.Localization;
-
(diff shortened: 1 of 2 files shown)


ksp_plugin_adapter/ksp_plugin_adapter.cs
```diff
@@ -2,8 +2,7 @@
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
-using KSP.Localization;
-
+using UnityEngine;
```
eggrobin, line 5: I think we have been trying to avoid that in Principia, fully-qualifying the UnityEngine stuff instead.

eggrobin (changes requested):

eggrobin (approved):

PR #3821, opened 2023-12-14
Do not create a flight plan copy when creating an optimization driver

pleroy:
The copy doesn't have an orbit analyser, we only want one when it was created by the optimizer.


eggrobin (approved):

eggrobin:
> Do not create a flight plan copy when creating an optimization driver

Technically we are still doing that; we are just not returning it until we have done something with it.

PR #3845, opened 2024-01-13
A function to compute the time intervals where a trajectory can collide with a celestial

pleroy:
Related to #2913.

### DIFF
--- a/physics/apsides_body.hpp
+++ b/physics/apsides_body.hpp
@@ -144,7 +144,7 @@ std::vector<Interval<Instant>> ComputeCollisionSegments(
     typename DiscreteTrajectory<Frame>::iterator end,
     DiscreteTrajectory<Frame> const& apoapsides,
     DiscreteTrajectory<Frame> const& periapsides) {
-  std::vector<Intervar<Instant>> segments;
+  std::vector<Interval<Instant>> segments;
 
   auto squared_distance_from_centre = [&reference,
                                        &trajectory](Instant const& time) {
@@ -166,7 +166,7 @@ std::vector<Interval<Instant>> ComputeCollisionSegments(
   for (auto const& [time, _] : apoapsides) {
     apoapsides_times.push_back(time);
   }
-  if (apoapsides_time.empty() ||
+  if (apoapsides_times.empty() ||
       apoapsides_times.front() > periapsides.begin()->time) {
     apoapsides_times.push_front(begin->time);
   }
@@ -174,7 +174,7 @@ std::vector<Interval<Instant>> ComputeCollisionSegments(
     apoapsides_times.push_back(std::prev(end)->time);
   }
 
-  auto ait = apoapsides_time.begin();
+  auto ait = apoapsides_times.begin();
   for (auto pit = periapsides.begin(); pit != periapsides.end();) {
     Instant apoapside_time = *ait;
     Instant periapside_time = pit->time;
@@ -182,7 +182,7 @@ std::vector<Interval<Instant>> ComputeCollisionSegments(
 
     // No collision possible if the periapside is above |max_radius|.
     if (squared_distance_from_centre(periapside_time) < max_radius²) {
-      Interval<Time> segment;
+      Interval<Instant> segment;
       if (squared_distance_from_centre(apoapside_time) > max_radius²) {
         // The periapside is below |max_radius| and the preceding apoapside is
         // above.  Find the intersection point.
@@ -228,10 +228,9 @@ std::vector<Interval<Instant>> ComputeCollisionSegments(
       }
 
       segments.push_back(segment);
-    } else {
-      ++ait;
-      ++pit;
     }
+    ++ait;
+    ++pit;
   }
 
   return segments;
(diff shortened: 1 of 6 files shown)


physics/apsides_body.hpp
```diff
@@ -135,6 +137,115 @@ void ComputeApsides(Trajectory<Frame> const& reference,
...
+        trajectory.EvaluatePosition(time);
+    Displacement<Frame> const displacement =
+        trajectory_position - body_position;
+    return displacement.Norm²();
+  };
+
+  auto const max_radius² = Pow<2>(reference_body.max_radius());
+
+  // NOTE(phl): This function doesn't consider |t_min| and |t_max| as potential
+  // periapsis, only as potential apoapsis.  Unsure if that's right.
```
eggrobin, line 161: That cannot be right.
In particular a trajectory could have neither apoapsis nor periapsis (it could be monotone in altitude from t_min to t_max), and still run into the side of a mountain.

For the purposes of this function, t_min and t_max should always be treated as apsides of the kind opposite the first actual apsis and last actual apsis respectively.

pleroy: Done, PTAL.

physics/apsides_body.hpp
```diff
@@ -135,6 +137,150 @@ void ComputeApsides(Trajectory<Frame> const& reference,
...
+  for (; it != apsides_times.end(); previous_it = it, ++it) {
+    // Here |it| designates a periapsis, and |previous_it| the previous
+    // apoapsis, if any.
+    Instant const periapsis_time = *it;
+
+    // No collision is possible if the periapsis is above |max_radius|.
+    if (squared_distance_from_centre(periapsis_time) < max_radius²) {
+      Interval<Instant> interval;
+      if (previous_it == apsides_times.end()) {
+        // No previous periapsis can only happens the first time through the
```
eggrobin, line 199:
```suggestion
        // No previous periapsis can only happen the first time through the
```

physics/apsides_body.hpp
```diff
@@ -135,6 +137,150 @@ void ComputeApsides(Trajectory<Frame> const& reference,
...
+          // above.  Find the intersection point.
+          interval.min = Brent(
+              [max_radius²,
+               &squared_distance_from_centre](Instant const& time) {
+                return squared_distance_from_centre(time) - max_radius²;
+              },
+              apoapsis_time,
+              periapsis_time);
+        } else {
+          // A periapsis below |max_radius| can only happen the first time
```
eggrobin, line 217:
```suggestion
          // An apoapsis below |max_radius| can only happen the first time
```

physics/apsides_body.hpp
```diff
@@ -135,6 +137,150 @@ void ComputeApsides(Trajectory<Frame> const& reference,
...
+    // apoapsis, if any.
+    Instant const periapsis_time = *it;
+
+    // No collision is possible if the periapsis is above |max_radius|.
+    if (squared_distance_from_centre(periapsis_time) < max_radius²) {
+      Interval<Instant> interval;
+      if (previous_it == apsides_times.end()) {
+        // No previous periapsis can only happens the first time through the
+        // loop.
+        interval.min = periapsis_time;
```
eggrobin, line 201:
```suggestion
        CHECK_EQ(periapsis_time, trajectory.t_min());
        interval.min = periapsis_time;
```

physics/apsides_body.hpp
```diff
@@ -135,6 +137,150 @@ void ComputeApsides(Trajectory<Frame> const& reference,
...
+               &squared_distance_from_centre](Instant const& time) {
+                return squared_distance_from_centre(time) - max_radius²;
+              },
+              apoapsis_time,
+              periapsis_time);
+        } else {
+          // A periapsis below |max_radius| can only happen the first time
+          // through the loop because the nested loop below skips the other
+          // ones.
+          interval.min = periapsis_time;
```
eggrobin, line 220: That should be apoapsis, right?
```suggestion
          CHECK_EQ(apoapsis_time, trajectory.t_min());
          interval.min = apoapsis_time;
```

physics/apsides_test.cpp
```diff
@@ -369,5 +378,364 @@ TEST_F(ApsidesTest, ComputeNodes) {
 
 #endif
 
+// A dedicated fixture for |ComputeCollisionIntervals| because we have many
+// tests for that function.
```
eggrobin, line 382: Do we have tests covering the cases where the endpoints of the trajectory are under max_radius?

I think there are three cases that are treated specially in the implementation, initial time under max_radius ascending, initial time under max_radius descending, and final time under max_radius descending, but it would probably be safest to test all four combinations.

physics/apsides_body.hpp
pleroy: Good catch.

physics/apsides_test.cpp
pleroy: The test that changed has its initial point below `max_radius` and that's a periapsis.

pleroy: Added the other 3 tests.  I also did the TDD thing.

eggrobin (approved):

Issue #3822, opened 2023-12-16
Crash clicking "Optimize"


lpgagnon:
first attempt to use the optimizer in post-hotfix Jordan did not go well. Getting a reproducible crash; journal attached.

context: vessel just reached LEO; plotted maneuver for transfer to Mercury intercept, saved, clicked Optimize to see what it would do. crash to desktop, with principia assertion failure:
```
F1215 20:25:12.957080 3736604672 interface_body.hpp:565] Unexpected extension 992685744
```
restarted game, loaded same save, click Optimize: same result, 100% reproducible, making a journal easy to produce.

[journal & log files, 29MB zip](https://drive.google.com/file/d/1fbZo7EV0IeBUFykH07MEDRdAyQQHpUv6/view?usp=share_link)

pleroy:
On the face of it, it looks like it's accessing some random piece of memory.  I could imagine a confusion regarding passing values vs. references across the C#/C++ boundary, but then it works on Windows.  I wonder if it is specific to macOS.  @rnlahaye Have you ever tried pressing the Optimize button?

rnlahaye:
I have not tried it yet. I will do so and see what happens.

rnlahaye:
Tried it; no crashes.

lpgagnon:
fwiw, it wasn't a one-off; different vessel, different target (earth->moon), same crash.

[new journal, maybe having two to compare will help](https://drive.google.com/file/d/1PIe5gJaLErGceysf57dljbFnW2Tg7AyV/view?usp=share_link)

it does have the look of random data, the "extension" keeps changing. 
`F1216 20:05:32.158290 3166703616 interface_body.hpp:565] Unexpected extension -91866624`
`F1216 19:59:21.077450 3993833472 interface_body.hpp:565] Unexpected extension 465795056`

pleroy:
@rnlahaye This is going to be a strange request: how do you build Principia on macOS?  @eggrobin recently purchased a mac mini 2023 with an M2 for other reasons, and when trying for 10 minutes to build Principia on that machine we have tons of errors related to ARM vs. Intel inconsistencies.  If you have a recipe that works, we'd appreciate if you could share it.  (In case you wonder, for releasing we build on Azure pipelines where the underlying machine is an x64.)

@lpgagnon What machine do you have?  Is it ARM-based?  It doesn't seem impossible that the bug would actually depend on the hardware/ABI.

rnlahaye:
[...]

PR #3881, opened 2024-02-12
Change the generator to produce a marshaler when exchanging a message that has is_class

pleroy:
This requires to change the structure of the marshalers to have access to private fields.

#3822.


eggrobin (approved):

eggrobin:
This appears to fix #3822.

PR #3884, opened 2024-02-14
Update zh-cn.cfg

Peter-JY:
Update new localization contents of Orbit Optimization

### DIFF
--- a/ksp_plugin_adapter/localization/zh-cn.cfg
+++ b/ksp_plugin_adapter/localization/zh-cn.cfg
@@ -204,8 +204,8 @@ Localization {
     #Principia_FlightPlan_TargetAltitude = 目标高度
     #Principia_FlightPlan_AltitudeUnit = m
     #Principia_FlightPlan_TargetInclination = 目标轨道倾角
-    #Principia_FlightPlan_OptimizeInclinationOn = 打开
-    #Principia_FlightPlan_OptimizeInclinationOff = 关闭
+    #Principia_FlightPlan_OptimizeInclinationOn = 启用
+    #Principia_FlightPlan_OptimizeInclinationOff = 禁用
     #Principia_FlightPlan_OptimizeInclinationNoText =
     #Principia_FlightPlan_InclinationUnit = °
     #Principia_FlightPlan_ManœuvreHeader = 第<<1>>次轨道机动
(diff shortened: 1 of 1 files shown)


ksp_plugin_adapter/localization/zh-cn.cfg
```diff
@@ -115,7 +115,7 @@ Localization {
     #Principia_OrbitAnalyser_OrbitDescription_Stationary = 静止
     // See, e.g., ITU-R-HDB-55 https://www.itu.int/pub/R-HDB-55.
     #Principia_OrbitAnalyser_OrbitDescription_Semisynchronous = 半同步
-    #Principia_OrbitAnalyser_OrbitDescription_SunSynchronous = 太陽同步
+    #Principia_OrbitAnalyser_OrbitDescription_SunSynchronous = 太阳同步
```
al2me6, line 118: Huh, thanks for catching that.

ksp_plugin_adapter/localization/zh-cn.cfg
```diff
@@ -200,12 +200,12 @@ Localization {
     #Principia_FlightPlan_Delete = 删除当前计划
     #Principia_FlightPlan_Rebase = 变基
     #Principia_FlightPlan_Duplicate = 重复
-    #Principia_FlightPlan_Optimization = Optimization parameters for <<1>> flyby
-    #Principia_FlightPlan_TargetAltitude = Altitude
+    #Principia_FlightPlan_Optimization = <<1>>飞略轨道参数优化
```
al2me6, line 203: I wonder if `<<1>>飞略优化参数` would be more appropriate here.

Peter-JY: I think `<<1>>飞掠参数优化` will be ok as well, since "parameter" only refers to orbit parameter in this context. But `优化` can't be placed in the middle of "<<1>>飞掠参数", or `优化` will be seen as a verb, and it will mean "<<1>> flyby optimizes parameters".

al2me6: Hm, that is fair. I am admittedly unfamiliar with technical terminology.

Peter-JY: And I'm sorry that I wrongly input "飞掠" (means flyby) as "飞略" : (
I'll fix it instantly.

Issue #3302, opened 2022-01-22
Beef up TransposedView


eggrobin:
> (Note: we may want to have a more general and more powerful `TransposedView` in the future as the result of all `Transpose` (with conversions to the appropriate matrix types if actual transposition of the data is needed), but this is a rabbit hole for another day; for now this draft is good enough.)

_Originally posted by @eggrobin in https://github.com/mockingbirdnest/Principia/pull/3287#discussion_r789228715_

pleroy:
Also unify the operators in `fixed_arrays` and `unbounded_arrays` to avoid having each implementation twice.

PR #3891, opened 2024-03-09
Improve TransposeView by exposing operations for sizing and indexing

pleroy:
Note that this makes heavy use of concepts, so we'll have to see if Clang is happy with it.

Contributes to #3302.

### DIFF
--- a/numerics/concepts.hpp
+++ b/numerics/concepts.hpp
@@ -1,0 +1,34 @@
+#pragma once
+
+#include <concepts>
+#include <type_traits>
+
+namespace principia {
+namespace numerics {
+namespace _concepts {
+namespace internal {
+
+template<typename T>
+concept one_dimensional = requires(T& t, int index) {
+  { t[index] } -> std::same_as<typename T::Scalar&>;
+};
+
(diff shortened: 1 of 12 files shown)


numerics/concepts.hpp
```diff
@@ -0,0 +1,37 @@
...
+};
+
+template<typename T>
+concept two_dimensional = requires(T& t, int row, int column) {
+  { t(row, column) } -> std::same_as<typename T::Scalar&>;
+};
+
+template<typename T>
+concept unbounded = requires(T& t) {
+  { t.size() } -> std::same_as<int>;
```
eggrobin, line 23: I could imagine adding `size` to the fixed types, just like `std::array` has a size.

pleroy: Hmm, actually they `size`, `rows` and `columns`.  So I don't even need the `unbounded` concept, which is good because we want to handle fixed and unbounded similarly if possible.

Note that `size` probably doesn't make much sense for matrices, so I'll remove it ultimately.

eggrobin (approved):

PR #3892, opened 2024-03-10
Use concepts to simplify our code

pleroy:
In this PR I am removing all uses of `std::void_t`, as it is the hardest trait to use.

Incidentally, this is changing `Hilbert` in a way that avoids a bug, so no need to `#ifdef` code anymore.

### DIFF
--- a/geometry/concepts.hpp
+++ b/geometry/concepts.hpp
@@ -6,13 +6,13 @@ namespace _concepts {
 namespace internal {
 
 template<typename T1, typename T2>
-concept has_inner_product = requires(T1 const& t1, T2 const& t2) {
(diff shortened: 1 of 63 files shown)


geometry/concepts.hpp
```diff
@@ -0,0 +1,19 @@
+#pragma once
+
+namespace principia {
+namespace geometry {
+namespace _concepts {
+namespace internal {
+
+template<typename T1, typename T2>
+concept has_inner_product = requires(T1 const& t1, T2 const& t2) {
```
eggrobin, line 9: Rename to hilbert as discussed.

eggrobin (approved):

PR #3912, opened 2024-03-23
Add some algebraic concepts and remove geometry/traits

eggrobin:


quantities/concepts.hpp
```diff
@@ -14,13 +14,50 @@ namespace internal {
 using namespace base::_traits;
 using namespace quantities::_quantities;
 
+template<typename G>
+concept additive_group = requires(G x, G y) {
+  G{};
+  { -x } -> std::same_as<G>;
```
pleroy, line 20: Should we have `+x`?

Also wondering if we should require that `G::zero` exist.  That would sometimes be useful.

quantities/concepts.hpp
```diff
@@ -14,13 +14,50 @@ namespace internal {
...
+};
+
+template<typename V>
+concept real_vector_space = vector_space<V, double>;
+
+template<typename A, typename K>
+concept affine_space = requires(A x, A y) {
+  { x - y } -> vector_space<K>;
+  { y + (x - y) } -> std::same_as<A>;
+  { y += (x - y) } -> std::same_as<A&>;
```
pleroy, line 43: We should probably require an `operator-(A, V)` and similarly for `-=`.

geometry/pair.hpp
```diff
@@ -66,7 +66,7 @@ struct enable_if_vector;
 template<typename T1, typename T2, typename U>
 struct enable_if_vector<
     Pair<T1, T2>, U,
-    std::enable_if_t<std::conjunction_v<is_vector<T1>, is_vector<T2>>>>
+    std::enable_if_t<additive_group<T1> && additive_group<T2>>>
```
pleroy, line 69: I think that this should require vector spaces, otherwise the implementation of barycentre is problematic.

numerics/polynomial_in_monomial_basis.hpp
```diff
@@ -93,10 +93,8 @@ class PolynomialInMonomialBasis : public Polynomial<Value_, Argument_> {
   // powers of (argument - origin).
   constexpr PolynomialInMonomialBasis(Coefficients coefficients,
                                       Argument const& origin);
-  template<typename A = Argument,
-           typename = std::enable_if_t<is_vector_v<A>>>
-  explicit constexpr PolynomialInMonomialBasis(
-      Coefficients coefficients);
+  explicit constexpr PolynomialInMonomialBasis(Coefficients coefficients)
+    requires additive_group<Argument>;
```
pleroy, line 97: This is the weirdest constraint.  How is it sufficient for the argument to have an additive structure?  I mean, we'll compute `arg^N`.

Note that this is showing something problematic with the concepts: you require an addition, and if your argument doesn't have an addition the instantiation will fail.  But if you happen to use a multiplication, the concept will not prevent you from doing that, and you may end up with obscure errors.

numerics/polynomial_in_monomial_basis_body.hpp
```diff
@@ -448,11 +447,10 @@ Primitive() const {
...
 template<typename Value_, typename Argument_, int degree_,
          template<typename, typename, int> typename Evaluator>
-template<typename, typename>
 Primitive<Value_, Argument_>
-PolynomialInMonomialBasis<Value_, Argument_, degree_, Evaluator>::
-Integrate(Argument const& argument1,
-          Argument const& argument2) const {
+PolynomialInMonomialBasis<Value_, Argument_, degree_, Evaluator>::Integrate(
+    Argument const& argument1,
+    Argument const& argument2) const requires additive_group<Value> {
```
pleroy, line 453: Let's revert the formatting change and put the `require` clause on its own line, indented by 2 positions.  Clang is not very good at formatting concepts, we should ignore it.

quantities/concepts.hpp
```diff
@@ -14,13 +14,50 @@ namespace internal {
...
                    is_instance_of_v<Quantity, std::remove_const_t<T>>;
 
 }  // namespace internal
 
+using internal::additive_group;
+using internal::affine_space;
+using internal::real_affine_space;
+using internal::real_vector_space;
 using internal::quantity;
+using internal::vector_space;
```
pleroy, line 60: I think that this is missing a `ring`.

pleroy (commented):
Let's remove the `ksp_plugin_adapter_stub` from this PR.

quantities/concepts.hpp
eggrobin: We’ll see if we need one.

quantities/concepts.hpp
eggrobin: Re zero, that is what G{} is for. double::zero certainly does not exist.

numerics/polynomial_in_monomial_basis.hpp
eggrobin: So far we have just been using concepts to replace enable_if, and we had been using that minimally to disable things that would not compile; `is_vector` was just a statement about the type of the difference, even weaker than additive_group.

We should move to concepts that more closely match the actual requirements, so that we get better errors; but we should do that from the bottom up, not from the middle in Polynomial.

The requirement for a multiplicative structure on Difference<Argument> should just be on Polynomial itself anyway, not here.

geometry/pair.hpp
eggrobin: See my other comment.

PR #3913, opened 2024-03-23
Rename quantity to make way for an actual quantity concept

eggrobin:


quantities/concepts.hpp
```diff
@@ -49,17 +49,22 @@ concept affine_space = requires(A x, A y) {
 template<typename V>
 concept real_affine_space = affine_space<V, double>;
 
+// std::integral || std::floating_point rather than
+// std::convertible_to<double, T> because
+// the former introduces ambiguities on Sign * Vector.
```
pleroy, line 54: Latter?

pleroy (commented):
The phrase `convertible_to_quantity` is not overly elegant, and is what we have been calling "scalar" in the past.  I am a tad worries that it's too much of a mouthful.  For instance, it would be unpleasant to write:
```
template<convertible_to_quantity Scalar>
class ...
```
But let's see how things turn out once we actually have and use `quantity`

eggrobin:
> it would be unpleasant to have to write:
> ```C++
> template<convertible_to_quantity Scalar>
> class ...
> ```
> But let's see how things turn out once we actually have and use quantity

We should never want to write that; our types are templatized on things that are actually `quantity` (that is, `Quantity` or `double`, not `convertible_to_quantity`, which includes `int`—and even `char` or `bool`…)

PR #3920, opened 2024-03-28
Get rid of Barycentre of std::pair and simplify its callers

eggrobin:
Not touching the container one for now, so those callers are still ugly.
Also we probably want a 1-parameter overload for equal weights, but I will do that in a subsequent change.

### DIFF
--- a/geometry/barycentre_calculator_body.hpp
+++ b/geometry/barycentre_calculator_body.hpp
@@ -59,8 +59,7 @@ template<affine Point, homogeneous_field Weight, std::size_t size>
   requires homogeneous_vector_space<Difference<Point>, Weight>
 Point Barycentre(Point const (&points)[size], Weight const (&weights)[size]) {
   static_assert(size != 0);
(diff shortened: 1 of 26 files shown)


geometry/barycentre_calculator_body.hpp
```diff
@@ -55,15 +55,23 @@ template<affine Point, homogeneous_field Weight>
...
-             std::pair<Scalar, Scalar> const& weights) {
-  BarycentreCalculator<T, Scalar> calculator;
-  calculator.Add(ts.first, weights.first);
-  calculator.Add(ts.second, weights.second);
+template<affine Point, homogeneous_field Weight, std::size_t size>
+  requires homogeneous_vector_space<Difference<Point>, Weight>
+Point Barycentre(Point const (&points)[size], Weight const (&weights)[size]) {
+  static_assert(size != 0);
+  BarycentreCalculator<Point, Weight>
+      calculator;
```
pleroy, line 63: Move to the previous line.

PR #3943, opened 2024-03-30
Log CPUID information on startup

eggrobin:


base/cpuid.cpp
```diff
@@ -48,17 +69,29 @@ std::string CPUVendorIdentificationString() {
...
-  auto const leaf_1 = CPUID(1, 0);
-  return static_cast<CPUFeatureFlags>(
-             (ecx_bit * leaf_1.ecx | edx_bit * leaf_1.edx) &
-             static_cast<std::uint64_t>(flags)) == flags;
+std::string CPUFeatures() {
+  std::string result;
+  for (auto const flag : CPUIDFlags()) {
+    if (flag.IsSet()) {
+      if (!result.empty()) {
+        result += " ";
```
pleroy, line 89: Use `StrJoin`.

base/cpuid.cpp
```diff
@@ -9,18 +10,19 @@
...
 namespace {
 
-struct CPUIDResult {
-  std::uint32_t eax;
-  std::uint32_t ebx;
-  std::uint32_t ecx;
-  std::uint32_t edx;
-};
+std::vector<CPUIDFeatureFlag>& CPUIDFlags() {
+  static std::vector<CPUIDFeatureFlag> result;
```
pleroy, line 23: This singleton is weird.  Why is this not a static member variable of `CPUIDFeatureFlag` with a selector?

geometry/grassmann_test.cpp
```diff
@@ -82,7 +82,7 @@ class GrassmannTest : public testing::Test {
 using GrassmannDeathTest = GrassmannTest;
 
 TEST_F(GrassmannTest, VectorFMA) {
-  if (!CanEmitFMAInstructions || !HasCPUFeatures(CPUFeatureFlags::FMA)) {
+  if (!CanEmitFMAInstructions || !cpuid_feature_flags::FPU.IsSet()) {
```
pleroy, line 85: Why the change from `FMA` to `FPU`?

base/cpuid.cpp
```diff
@@ -39,6 +41,25 @@ CPUIDResult CPUID(std::uint32_t const eax, std::uint32_t const ecx) {
...
+  CHECK_LT(bit, 32);
+  CPUIDFlags().push_back(*this);
+}
+
+std::string_view CPUIDFeatureFlag::name() const {
+  return name_;
+}
+
+bool CPUIDFeatureFlag::IsSet() const {
+  return CPUID(leaf_, sub_leaf_).*field_ & (1 << bit_);
```
pleroy, line 60: It would be good if this function was fast.  Why don't we extract the bit at construction instead of computing it on each call?  Also, we should mark this function as inline.

eggrobin: We call this once.

base/cpuid.cpp
```diff
@@ -1,16 +1,19 @@
 #include "base/cpuid.hpp"
 
 #include <string>
+#include <ranges>
```
pleroy, line 4: Albaphetize.

PR #3947, opened 2024-03-31
Our C++ is too modern for clang

eggrobin:
ElementsAre ends up using `std::span::const_iterator` which [is C++23](https://en.cppreference.com/w/cpp/container/span).

### DIFF
--- a/ksp_plugin_test/plugin_compatibility_test.cpp
+++ b/ksp_plugin_test/plugin_compatibility_test.cpp
@@ -262,8 +262,13 @@ TEST_P(PluginCompatibilityTest, Reach) {
   // Saturn flyby from the video; however, the Saturn flyby seems to be missing
   // from this flight plan even if it is extended further.
   EXPECT_THAT(flybys, SizeIs(Ge(10)));
+#if PRINCIPIA_COMPILER_MSVC
   std::span const first_10_flybys(flybys.begin(), 10);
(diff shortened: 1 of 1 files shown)


ksp_plugin_test/plugin_compatibility_test.cpp
```diff
@@ -262,8 +262,13 @@ TEST_P(PluginCompatibilityTest, Reach) {
   // Saturn flyby from the video; however, the Saturn flyby seems to be missing
   // from this flight plan even if it is extended further.
   EXPECT_THAT(flybys, SizeIs(Ge(10)));
+#if PRINCIPIA_COMPILER_MSVC
   std::span const first_10_flybys(flybys.begin(), 10);
   std::span const subsequent_flybys(flybys.begin() + 10, flybys.end());
+#else  // TODO(egg): Remove when clang really has C++23.
```
pleroy, line 268: Let's use the exact version of Clang, like we do with `_MSC_FULL_VER`, this will ensure that this special code will go away ultimately, even if we don't look at the TODOs.  In `macros.hpp` we have a macro `CLANG_VERSION_GE`, we should have a `CLANG_VERSION_EQ` and use it here.

eggrobin: Clang 17 already supports this: https://godbolt.org/z/87nYGx5bY.

PR #3958, opened 2024-04-01
Test ill-formed constructs

eggrobin:
This finds a bug in Multivector::WriteToMessage.
It is easy to add a constraint there, but then it trickles down to other WriteToMessage functions, some of which are virtual. Let’s fix it in a subsequent change.

### DIFF
--- a/testing_utilities/check_well_formedness.hpp
+++ b/testing_utilities/check_well_formedness.hpp
@@ -5,29 +5,26 @@
 #define PRINCIPIA_CONCATENATE_EXPANSION(s1, s2) PRINCIPIA_CONCATENATE(s1, s2)
 
 #define PRINCIPIA_CHECK_WELL_FORMEDNESS(unique_concept_name,                  \
-                                         expected_well_formedness,                    \
-                                         expected_well_formedness_description,        \
-                                         expression, \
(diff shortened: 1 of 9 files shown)


testing_utilities/check_well_formedness.hpp
```diff
@@ -0,0 +1,30 @@
...
+
+#define PRINCIPIA_CONCATENATE(s1, s2) s1##s2
+#define PRINCIPIA_CONCATENATE_EXPANSION(s1, s2) PRINCIPIA_CONCATENATE(s1, s2)
+
+#define PRINCIPIA_CHECK_WELL_FORMEDNESS(unique_concept_name,                  \
+                                        expected_well_formedness,             \
+                                        expected_well_formedness_description, \
+                                        expression,                           \
+                                        ...)                                  \
+  template<template<typename> typename WITH = std::type_identity_t>           \
```
pleroy, line 12: The uppercase `WITH` is rather tacky.  We have the precedent of `with_evaluator`, let's call this one `with_parameter`.

eggrobin: `with_variable`, as discussed.

PR #3967, opened 2024-04-07
A test that uses Boost multiprecision to validate the platform libraries

pleroy:
#1760.


functions/std_accuracy_test.cpp
```diff
@@ -0,0 +1,59 @@
...
+      double const α = angle_distribution(random);
+      max_sin_ulp_distance =
+          std::max(max_sin_ulp_distance,
+                   ULPDistance(std::sin(α), static_cast<double>(Sin(α))));
+      max_cos_ulp_distance =
+          std::max(max_cos_ulp_distance,
+                   ULPDistance(std::cos(α), static_cast<double>(Cos(α))));
+    }
+    EXPECT_EQ(1, max_sin_ulp_distance);
+    EXPECT_EQ(1, max_cos_ulp_distance);
```
eggrobin, line 40: This tells us that it sometimes misrounds, but it does not tell us that it is faithful, nor how much it misrounds. In a subsequent PR it would be nice to compute the distance between `Sin(α)` and `std::sin(α)` in (fractional) ULPs of `static_cast<double>(Sin(α))`.

eggrobin (approved):

PR #3975, opened 2024-04-10
A basic accurate table generator

pleroy:
Many things to be improved, notably:
1. Only looks for accurate values for a single function.
2. Does an exhaustive search.
3. Single-threaded.

#1760.

### DIFF
--- a/functions/accurate_table_generator_test.cpp
+++ b/functions/accurate_table_generator_test.cpp
@@ -1,6 +1,7 @@
 #include "functions/accurate_table_generator.hpp"
 
 #include <iomanip>
+#include <string>
 
 #include "boost/multiprecision/cpp_int.hpp"
 #include "functions/multiprecision.hpp"
--- a/functions/accurate_table_generator_test.cpp
+++ b/functions/accurate_table_generator_test.cpp
@@ -13,11 +13,10 @@
 
 namespace principia {
 namespace functions {
-namespace _multiprecision {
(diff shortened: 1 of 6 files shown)


functions/accurate_table_generator_test.cpp
```diff
@@ -0,0 +1,54 @@
...
+
+class AccurateTableGeneratorTest : public ::testing::Test {};
+
+TEST_F(AccurateTableGeneratorTest, Sin5) {
+  auto const x = ExhaustiveSearch<5>({Sin}, 5.0 / 128.0);
+  EXPECT_EQ(x, cpp_rational(2814749767106647, 72057594037927936));
+  EXPECT_THAT(static_cast<double>(x),
+              RelativeErrorFrom(5.0 / 128.0, IsNear(3.1e-14_(1))));
+
+  // The stupid language doesn't allow printing a float in binary.  So to verify
```
eggrobin, line 34: It might be useful to write a utility that prints a float in binary, or just produces a range of digits; this test is not exactly readable, and we will probably want to do more of that sort of thing.

eggrobin (approved):

functions/accurate_table_generator_test.cpp
pleroy: I'll need `ToMathematica` anyway for these Boost types, so that will be a better basis for checking the output.

PR #3995, opened 2024-05-05
Prerequisites for the implementation of the Stehlé-Zimmermann algorithm

pleroy:
1. Add support for Boost multiprecision types in `quantities`, in particular for elementary functions and tracing.
2. Adjust the concepts so that the Boost multiprecision types are considered to have the appropriate algebraic structure.
3. Fix a bug in the composition of polynomials with nontrivial origins.

#1760.

### DIFF
--- a/quantities/elementary_functions_body.hpp
+++ b/quantities/elementary_functions_body.hpp
@@ -7,7 +7,6 @@
 #include <cmath>
 
 #include "boost/multiprecision/cpp_int.hpp"
-#include "boost/multiprecision/fwd.hpp"
 #include "numerics/cbrt.hpp"
 #include "numerics/fma.hpp"
 #include "numerics/next.hpp"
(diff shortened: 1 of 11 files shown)


quantities/elementary_functions_body.hpp
```diff
@@ -5,30 +5,38 @@
...
 
 template<typename Q1, typename Q2>
 Product<Q1, Q2> FusedMultiplyAdd(Q1 const& x,
                                  Q2 const& y,
                                  Product<Q1, Q2> const& z) {
-  return si::Unit<Product<Q1, Q2>> *
-         numerics::_fma::FusedMultiplyAdd(
-             x / si::Unit<Q1>, y / si::Unit<Q2>, z / si::Unit<Product<Q1, Q2>>);
+  if constexpr (is_number<Q1>::value || is_number<Q2>::value) {
+    return x * y + z;
```
eggrobin, line 33: This condition should also check for `number_category<Q2>::value == number_kind_integer || number_category<Q2>::value == number_kind_rational`; for floating-point we cannot just do that (but I think there is an `fma`).

eggrobin (approved):

PR #4006, opened 2024-05-20
A benchmark to decide on the spacing of accurate values table

pleroy:
* With a spacing of 2/256, we need a degree-2 polynomial for cos.
* With a spacing of 2/1024, we can use a degree-1 polynomial for both sin and cos.

The latter wins:
```
Run on (48 X 3793 MHz CPU s)
CPU Caches:
  L1 Data 32 KiB (x24)
  L1 Instruction 32 KiB (x24)
  L2 Unified 512 KiB (x24)
  L3 Unified 32768 KiB (x4)
-----------------------------------------------------------------------------------
Benchmark                                         Time             CPU   Iterations
-----------------------------------------------------------------------------------
BM_EvaluateSinTableSpacing<2.0 / 256.0>        2.56 ns         2.55 ns    263530000
BM_EvaluateSinTableSpacing<2.0 / 1024.0>       2.26 ns         2.25 ns    320000000
BM_EvaluateCosTableSpacing<2.0 / 256.0>        2.54 ns         2.57 ns    280000000
BM_EvaluateCosTableSpacing<2.0 / 1024.0>       2.23 ns         2.25 ns    298667000
```
#1760.

### DIFF
--- a/benchmarks/elementary_functions_experiments_benchmark.cpp
+++ b/benchmarks/elementary_functions_experiments_benchmark.cpp
@@ -21,6 +21,7 @@ constexpr Argument x_min = π / 12;
 constexpr Argument x_max = π / 6;
 static constexpr std::int64_t number_of_iterations = 1000;
 
+// A helper class to benchmark implementations with various table spacings.
 template<Argument table_spacing>
 class Implementation {
  public:
@@ -30,6 +31,8 @@ class Implementation {
   Value Cos(Argument x);
 
  private:
+  // Despite the name these are not accurate values, but for the purposes of
+  // benchmarking it doesn't matter.
   struct AccurateValues {
     Argument x;
     Value sin_x;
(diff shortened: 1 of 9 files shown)


benchmarks/elementary_functions_experiments_benchmark.cpp
```diff
@@ -0,0 +1,195 @@
...
+  Argument a[number_of_iterations];
+  for (std::int64_t i = 0; i < number_of_iterations; ++i) {
+    a[i] = uniformly_at(random);
+  }
+
+  Value v[number_of_iterations];
+  while (state.KeepRunningBatch(number_of_iterations)) {
+    for (std::int64_t i = 0; i < number_of_iterations; ++i) {
+      using namespace principia::quantities;
+      v[i] = implementation.Cos(a[i]);
```
eggrobin, line 173: Should we also benchmark latency?

eggrobin (approved):

benchmarks/elementary_functions_experiments_benchmark.cpp
pleroy: A latency benchmark is going to quickly converge to a fixed point.  That's not useful, because once all the invocations load from the same table entry, the influence of the table size disappears.  We really want to touch all the table entries.

PR #4014, opened 2024-06-02
Budget the nodes per flight plan segment

### LINKED ISSUE #3837: Node limit improvements
from https://discord.com/channels/319857228905447436/480397772248580098/1192434841069813791

Plotting an Earth-Jupiter-Io flight plan, I run into the issue that no Ap/Pe node is displayed for the actual Io intercept, in either ICI or IJO frames.

This is because there's a limit to the number of nodes shown, starting at the beginning of the flight plan.

possible improvements:
- an option to only show nodes after the last maneuver
- a "start adding nodes from the end of the plan" mode
- make the limit per-segment

eggrobin:
Fix #3837.


ksp_plugin/interface_flight_plan.cpp
```diff
@@ -521,13 +521,21 @@ void __cdecl principia__FlightPlanRenderedApsides(
...
-                                  max_points,
-                                  rendered_apoapsides,
-                                  rendered_periapsides);
+  for (auto const& segment : flight_plan.segments()) {
+    DiscreteTrajectory<World> segment_rendered_apoapsides;
+    DiscreteTrajectory<World> segment_rendered_periapsides;
+    plugin->ComputeAndRenderApsides(
+        celestial_index,
+        flight_plan,
+        segment.begin(),
```
pleroy, line 530: Let's put `begin` and `end` on the same line, as it was originally formatted.  (Same in the other places.)

PR #4016, opened 2024-06-03
Round the polynomial coefficients using Mathematica and take advantage of powers of two

pleroy:
A bit faster:
```
Run on (48 X 3793 MHz CPU s)
CPU Caches:
  L1 Data 32 KiB (x24)
  L1 Instruction 32 KiB (x24)
  L2 Unified 512 KiB (x24)
  L3 Unified 32768 KiB (x4)
----------------------------------------------------------------------
Benchmark                            Time             CPU   Iterations
----------------------------------------------------------------------
BM_ExperimentSinSingleTable       2.89 ns         2.85 ns    235790000
BM_ExperimentCosSingleTable       2.84 ns         2.83 ns    248889000
```
#1760.

### DIFF
--- a/benchmarks/elementary_functions_experiments_benchmark.cpp
+++ b/benchmarks/elementary_functions_experiments_benchmark.cpp
@@ -411,22 +411,19 @@ Value SingleTableImplementation::Cos(Argument const x) {
   }
 }
 
-Value SingleTableImplementation::SinPolynomial(
-    Argument const x) {
+Value SingleTableImplementation::SinPolynomial(Argument const x) {
   // 84 bits.  Works for all binades.
   return -0x1.5555555555555p-3 + 0x1.111110B24ACB5p-7 * x;
 }
 
-Value SingleTableImplementation::CosPolynomial1(
-    Argument const x) {
-    // 72 bits.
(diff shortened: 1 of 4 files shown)


benchmarks/elementary_functions_experiments_benchmark.cpp
```diff
@@ -152,7 +152,9 @@ class SingleTableImplementation {
     Value cos_x;
   };
 
-  static constexpr Value cos_polynomial_0 = -0.499999999999999999999872434553;
+  // If this was ever changed to a value that is not a power of 2, extra care
+  // would be needed in the computations that use it.
```
eggrobin, line 156: For that reason, I would much prefer inlining the -0.5 where this is used, since they need to change anyway if this changes.
This will also make it easier to review the error analyses etc.

eggrobin (approved):

benchmarks/elementary_functions_experiments_benchmark.cpp
pleroy: As discussed, it's useful to give a name to that number and make sure that it has the same value in the two places where it's used.

PR #4023, opened 2024-06-08
Add option RoundingMode to CorrectlyRound

pleroy:
The syntax is something like `RoundingMode->Toward0` and the default is `RoundingMode->Automatic` which uses the global rounding mode.

#1760.

### DIFF
--- a/mathematica/floating_point.wl
+++ b/mathematica/floating_point.wl
@@ -88,9 +88,8 @@ sign(1+\[Mu]/2^(significandBits-1))2^(e-bias)]];
 UlpDistance[x_,y_]:=Abs[Representation[x]-Representation[y]]
 
 
-CorrectlyRound[x_,OptionsPattern[]]:=Block[
-{rounding},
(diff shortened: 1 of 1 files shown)


mathematica/floating_point.wl
```diff
@@ -84,10 +85,19 @@ e=IntegerPart[magnitude/2^(significandBits-1)];
 If[e==0,
 \[Mu]/2^(significandBits-1) 2^(1-bias),
 sign(1+\[Mu]/2^(significandBits-1))2^(e-bias)]];
-CorrectlyRound[x_]:=If[x==\[Infinity]||x==-\[Infinity],x,If[Abs[#]>=2^(bias+1),Sign[x]\[Infinity],#]&@FromRepresentation[correctlyRoundRepresentation[Representation[x]]]];
 UlpDistance[x_,y_]:=Abs[Representation[x]-Representation[y]]
 
 
+CorrectlyRound[x_,OptionsPattern[]]:=Block[
```
eggrobin, line 91: Why not With?

eggrobin (approved):

mathematica/floating_point.wl
pleroy: Done.

PR #4063, opened 2024-08-11
Flush harder in the Mathematica logger

pleroy:
#1760.

### DIFF
--- a/base/file_body.hpp
+++ b/base/file_body.hpp
@@ -36,6 +36,10 @@ inline OFStream::~OFStream() {
   stream_.close();
 }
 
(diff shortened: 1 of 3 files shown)


eggrobin (approved):

base/file_body.hpp
```diff
@@ -36,6 +36,10 @@ inline OFStream::~OFStream() {
   stream_.close();
 }
 
+void OFStream::Flush() {
```
eggrobin, line 39:
```suggestion
inline void OFStream::Flush() {
```

PR #4081, opened 2024-08-25
Add a SUPPORT.md file

pleroy:


SUPPORT.md
```diff
@@ -0,0 +1,75 @@
...
+#### Principia Crashes
+
+These crashes may manifest themselves as a Windows dialog box (on Windows) or a 
+SIGABRT (on Linux).  Or the game may just stop.
+
+Go to `<KSP directory>\glog\Principia`, you should find a `FATAL` file with a
+name that matches the time of the crash, for instance
+`FATAL.20230920-223637.6656.log`. You should also find an `INFO` file created
+shortly before that, for instance `INFO.20230920-223615.6656.log`.  Take these
+two files (one is not enough), upload them on gist. Link the gist in your github
```
eggrobin, line 41: We have been recommending that for a while, but gist isn’t very convenient for that, since these files get far too large to be handled by a tool meant for small source code snippets.
We should recommend attaching the files to the issue instead: https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/attaching-files

eggrobin (changes requested):

SUPPORT.md
eggrobin: It might also be a good idea, here and below, to explicitly say that the entire unredacted unabridged file should be posted, not snippets that you think are interesting.

SUPPORT.md
```diff
@@ -0,0 +1,75 @@
...
+issue.
+
+#### KSP or Unity Crashes
+
```
eggrobin, line 45: We should also ask for the latest INFO file, so we have the Principia version.

PR #4090, opened 2024-09-08
Speculative implementation of StehléZimmermannSimultaneousFullSearch

pleroy:
With this change, a single search around 1167/2048 takes 37 s (elapsed time) instead of 1833 s (that's about 50× faster).

#1760.


functions/accurate_table_generator_body.hpp
```diff
@@ -556,15 +571,111 @@ absl::StatusOr<cpp_rational> StehléZimmermannSimultaneousFullSearch(
...
-      return status;
+      // Some kind of internal error, give up.
+      {
+        absl::MutexLock l(&lock);
+        status_or_solution = status;
+      }
     }
+  };
+
+  // This variable should not be under a mutex at it might cause contention.
```
eggrobin, line 605:
```suggestion
  // This variable should not be under a mutex as it might cause contention.
```

functions/accurate_table_generator.hpp
```diff
@@ -46,13 +48,15 @@ absl::StatusOr<cpp_rational> StehléZimmermannSimultaneousSearch(
 
 // Performs a search around |starting_argument| to find a solution,
 // automatically adjusting the interval over which the search happens.  The
-// argument and function values must be nonzero.
+// argument and function values must be nonzero.  If |search_pool| is notnull,
```
eggrobin, line 51:
```suggestion
// argument and function values must be nonzero.  If |search_pool| is not null,
```

eggrobin (approved):

PR #4098, opened 2024-09-18
Add a test for quadratic and near-quadratic Hermite3, fix SolveQuadraticEquation for determinants near 0

eggrobin:
Thanks to @gotmachine for bringing this eight year old code to our attention.

### DIFF
--- a/numerics/root_finders_body.hpp
+++ b/numerics/root_finders_body.hpp
@@ -356,13 +356,13 @@ BoundedArray<Argument, 2> SolveQuadraticEquation(
   // Use double precision for the discriminant because there can be
   // cancellations.  Higham mentions that it is necessary “to use extended
   // precision (or some trick tantamount to the use of extended precision).”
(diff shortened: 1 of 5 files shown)


numerics/root_finders_body.hpp
```diff
@@ -356,13 +356,13 @@ BoundedArray<Argument, 2> SolveQuadraticEquation(
   // Use double precision for the discriminant because there can be
   // cancellations.  Higham mentions that it is necessary “to use extended
   // precision (or some trick tantamount to the use of extended precision).”
-  DoublePrecision<Discriminant> discriminant = TwoSum(a₁ * a₁, -4.0 * a₀ * a₂);
+  DoublePrecision<Discriminant> discriminant =
+      TwoProduct(a₁, a₁) - TwoProduct(4.0 * a₀, a₂);
 
-  if (discriminant.value == zero && discriminant.error == zero) {
```
pleroy, line 361: If we want to change this, we should document the invariants of `DoublePrecision`.  Your change assumes that the values returned by the functions in `double_precision.h` are normalized.  That's probably true, but it should be tested.

PR #4144, opened 2024-12-30
Fix nanobenchmarks on clang

eggrobin:


include_what_you_using_all_the_things.ps1
```diff
@@ -7,6 +7,7 @@
     --project:geometry `
     --project:integrators `
     --project:journal `
+    --project:nanobenchmarks `
```
pleroy, line 10: Alphabetize.

PR #4146, opened 2024-12-30
Sourcery for functions

eggrobin:


functions/core_math_accuracy_test.cpp
```diff
@@ -8,18 +8,20 @@
...
 namespace principia {
 namespace functions {
-namespace _multiprecision {
 
 using namespace boost::multiprecision;
+using namespace principia::functions::_multiprecision;
 using namespace principia::testing_utilities::_approximate_quantity;
 using namespace principia::testing_utilities::_is_near;
 
+namespace {
```
pleroy, line 23: As discussed, let's not do that.

PR #4194, opened 2025-04-03
Fix integrator plots

eggrobin:
Following some refactoring of `mathematica`, the tool had been generating strings instead of lists of numbers.

Also:
1. provide data on _all_ integrators: both ABA and BAB for the SPRKs as SRKN (bringing the number of varieties of symplectic or conjugate symplectic integrators from 37 to 57), and plot the integrators that are neither symplectic nor conjugate symplectic too;
2. provide information on the types of integrators to make it easier to filter them in Mathematica for specific plots;
3. solve the same problem with the same settings for different final times so that the evolution of the error over time can be examined (in particular, this shows the effect of the (conjugate-)symplectic property).

### DIFF
--- a/mathematica/integrator_plots.cpp
+++ b/mathematica/integrator_plots.cpp
@@ -183,7 +183,7 @@ class WorkErrorGraphGenerator {
                                  int* evaluations)> compute_accelerations,
       ODE::State initial_state,
       std::function<Errors(ODE::State const&)> compute_errors,
-      Instant const& tmax,
+      std::vector<Instant> const& tmax,
       std::string problem_name)
       : methods_(Methods()),
         compute_accelerations_(std::move(compute_accelerations)),
@@ -200,6 +200,12 @@ class WorkErrorGraphGenerator {
       v_errors_[i].resize(integrations_per_integrator_);
       e_errors_[i].resize(integrations_per_integrator_);
       evaluations_[i].resize(integrations_per_integrator_);
+      for (int j = 0; j < integrations_per_integrator_; ++j) {
+        q_errors_[i][j].resize(tmax_.size());
+        v_errors_[i][j].resize(tmax_.size());
+        e_errors_[i][j].resize(tmax_.size());
+        evaluations_[i][j].resize(tmax_.size());
+      }
     }
   }
 
@@ -225,11 +231,11 @@ class WorkErrorGraphGenerator {
     std::vector<std::string> names;
     for (int i = 0; i < methods_.size(); ++i) {
       q_error_data.emplace_back(
-          PlottableDataset(evaluations_[i], q_errors_[i], PreserveUnits));
+          PlottableDataset(evaluations_[i], q_errors_[i], ExpressInSIUnits));
       v_error_data.emplace_back(
-          PlottableDataset(evaluations_[i], v_errors_[i], PreserveUnits));
+          PlottableDataset(evaluations_[i], v_errors_[i], ExpressInSIUnits));
       e_error_data.emplace_back(
-          PlottableDataset(evaluations_[i], e_errors_[i], PreserveUnits));
+          PlottableDataset(evaluations_[i], e_errors_[i], ExpressInSIUnits));
       names.emplace_back(ToMathematica(methods_[i].name));
     }
     std::string result;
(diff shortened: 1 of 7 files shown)


mathematica/integrator_plots.cpp
```diff
@@ -36,28 +41,89 @@
...
-        u8###name, 1                     \
+#define SLMS_INTEGRATOR(name)                                      \
+  {                                                                \
+      static_cast<FixedStepSizeIntegrator<SecondOrderODE> const*>( \
+          &SymmetricLinearMultistepIntegrator<methods::name,       \
+                                              SecondOrderODE>()),  \
+      u8## #name,                                                  \
+      {methods::name::order},                                      \
+      "SLMS",                                                      \
+      1,                                                           \
```
pleroy, line 52: Comment the `1`.

mathematica/integrator_plots.cpp
```diff
@@ -94,64 +163,91 @@ using namespace principia::quantities::_si;
...
 };
 
 // This list should be sorted by:
-// 1. increasing order;
-// 2. SPRKs before SRKNs before symmetric linear multistep integrators.
+// 1. increasing convergence order;
+// 2. method types in the following order:
+//    a. SPRKs,
+//    b. SRKNs,
+//    c. symmetric linear multistep integrators,
```
pleroy, line 197: Capitalize or use the acronym.

mathematica/integrator_plots.cpp
```diff
@@ -177,20 +273,27 @@ class WorkErrorGraphGenerator {
...
         compute_accelerations_(std::move(compute_accelerations)),
         initial_state_(std::move(initial_state)),
         compute_errors_(std::move(compute_errors)),
         tmax_(tmax),
-        problem_name_(std::move(problem_name)) {
+        first_tolerance_(first_tolerance),
+        problem_name_(std::move(problem_name)),
+        step_reduction_(std::exp(std::log(starting_step_size_per_evaluation_ /
+                                          min_step_per_evaluation) /
+                                 integrations_per_integrator_)) {
```
pleroy, line 296: Use `pow`?

mathematica/integrator_plots.cpp
```diff
@@ -200,6 +303,12 @@ class WorkErrorGraphGenerator {
       v_errors_[i].resize(integrations_per_integrator_);
       e_errors_[i].resize(integrations_per_integrator_);
       evaluations_[i].resize(integrations_per_integrator_);
+      for (int j = 0; j < integrations_per_integrator_; ++j) {
```
pleroy, line 306: Could we try to use `std::int64_t` instead of `int`?  (Everywhere in this PR.)

mathematica/integrator_plots.cpp
```diff
@@ -219,137 +328,337 @@ class WorkErrorGraphGenerator {
     }
     CHECK_OK(bundle.Join());
 
-    std::vector<std::string> q_error_data;
-    std::vector<std::string> v_error_data;
-    std::vector<std::string> e_error_data;
+    std::vector<std::vector<std::vector<std::tuple<double, Length>>>>
```
pleroy, line 331: `std::tuple` in this case is just `std::pair`.  But in general I'd prefer to invest in a struct to give names to the components.  Also, naming the vectors would be good because it's really hard to figure out the purpose of each level.

mathematica/integrator_plots.cpp
```diff
@@ -219,137 +328,337 @@ class WorkErrorGraphGenerator {
...
+    std::vector<std::string> types;
     for (int i = 0; i < methods_.size(); ++i) {
-      q_error_data.emplace_back(
-          PlottableDataset(evaluations_[i], q_errors_[i], PreserveUnits));
-      v_error_data.emplace_back(
-          PlottableDataset(evaluations_[i], v_errors_[i], PreserveUnits));
-      e_error_data.emplace_back(
-          PlottableDataset(evaluations_[i], e_errors_[i], PreserveUnits));
-      names.emplace_back(ToMathematica(methods_[i].name));
+      q_error_data.emplace_back();
```
pleroy, line 341: `auto& q_error_data_for_method = q_error_data.emplace_back();`

Same everywhere.  All these `back()` make me throw up in my mouth a little bit.

mathematica/integrator_plots.cpp
```diff
@@ -219,137 +328,337 @@ class WorkErrorGraphGenerator {
...
     Problem problem;
+    InitialValueProblem<FirstOrderODE> first_order_problem;
     int number_of_evaluations = 0;
     problem.equation.compute_acceleration = std::bind(
         compute_accelerations_, _1, _2, _3, &number_of_evaluations);
     problem.initial_state = initial_state_;
+    first_order_problem.equation.compute_derivative =
+        [&number_of_evaluations, this](
+            Instant const& s,
+            std::tuple<Position<World>, Velocity<World>> const& y,
```
pleroy, line 383: `std::pair` (2x).

mathematica/integrator_plots.cpp
```diff
@@ -219,137 +328,337 @@ class WorkErrorGraphGenerator {
...
-    // [BM02] or [BCR01a] tend to use the average error (the normalized L¹ norm)
-    // instead.
-    q_errors_[method_index][time_step_index] = max_q_error;
-    v_errors_[method_index][time_step_index] = max_v_error;
-    e_errors_[method_index][time_step_index] = max_e_error;
-    evaluations_[method_index][time_step_index] = amortized_evaluations;
+    auto append_first_order_state =
+    [this, &max_q_error, &max_v_error, &max_e_error](
+        FirstOrderODE::State const& state) {
+      SecondOrderODE::State second_order_state;
```
pleroy, line 409: Use aggregate initialization to fill this struct.

mathematica/integrator_plots.cpp
```diff
@@ -219,137 +328,337 @@ class WorkErrorGraphGenerator {
...
+  std::vector<std::vector<std::vector<Speed>>> v_errors_;
+  std::vector<std::vector<std::vector<Energy>>> e_errors_;
+  std::vector<std::vector<std::vector<double>>> evaluations_;
+  std::vector<Instant> const tmax_;
+  Length const first_tolerance_;
   std::string const problem_name_;
-  double const step_reduction_ = 1.015;
   Time const starting_step_size_per_evaluation_ = 1 * Second;
   int const integrations_per_integrator_ = 500;
+  double const step_reduction_;
```
pleroy, line 603: Is this still used?

pleroy (commented):
Can we see a graph?

mathematica/integrator_plots.cpp
eggrobin: OK, but then at some point we need to do a general cleanup of the int-by-default indices in this codebase.

mathematica/integrator_plots.cpp
eggrobin: This is pretty much directly what we put in the Mathematica output, so we end up with nasty lists of lists of lists of lists anyway.
And we do not have support for pair in ToMathematica.

mathematica/integrator_plots.cpp
eggrobin: Yes, and in Mathematica I end up with things like `eErrorData[[indices, ;; ;; 1, 3 ;;]]` and `eErrorData[[indices, ;; ;; 1, 2 ;;]]`. But I don’t have any better ideas here.

mathematica/integrator_plots.cpp
eggrobin: No, ExplicitFirstOrderOrdinaryDifferentialEquation requires std::tuple.

mathematica/integrator_plots.cpp
eggrobin: Yes.

mathematica/integrator_plots.cpp
eggrobin: State is not an aggregate type, because it has a user-declared constructor (that lengthens to double precision). Here we have double precision already (from the other state).

```C++
  struct State final {
    // […]
    State(IndependentVariable const& t,
          DependentVariables const& q,
          DependentVariableDerivatives const& v);

    DoublePrecision<IndependentVariable> time;
    std::vector<DoublePrecision<DependentVariable>> positions;
    std::vector<
        DoublePrecision<Derivative<DependentVariable, IndependentVariable>>>
        velocities;
    // […]
  };
```

eggrobin:
Energy error as a function of evaluations for all integrators, coloured by type, on the Kepler problem with the eccentricity of 67P:
all integrators 67P

The same on the Kepler problem with the eccentricity of 1P, with a trail illustrating the evolution of maximal energy error with integration duration. The adaptive stepsize integrators can focus their work on the perihelion, and thus outperform everything else; but the energy drift of the integrators that are not symplectic or conjugate symplectic is visible:

all integrators 1P


The Kepler problem with the eccentricity of 1P again, but this time with position error. The maximum position error drifts for all integrators, symplecticity only helps with the energy error.

TODO(egg): Post once Mathematica has plotted it…

Legend:
integrator types

eggrobin:
All First-Same-As-Last SPRKs with ABA (▲) vs. BAB (▼) compositions on the circular Kepler problem:

Energy error:
FSAL SPRKs circular e


Position error:
FSAL SPRKs circular q


Legend:
FSAL SPRKs legend

PR #4195, opened 2025-04-03
Evaluate the effect of the angular resolution on the cost and quality of PlotMethod3

pleroy:
This code plots the same equipotentials at various angular resolutions and determines (1) their cost, given by the total number of points produced; and (2) their quality, given by the Euclidean distance from the plot at the "correct" resolution (human visual accuracy).

For convenience, the plotted lines are turned into `DiscreteTrajectory` objects, as that makes it possible to evaluate them at any time.

The outcome is that the cost varies approximately like `1/Sqrt[resolution]` and the distance to the "correct" plot like `Sqrt[resolution]`.

Note that at this point there is no telling how the distance corresponds to the "pleasantness" of the graph.  That will need to be examined in game.

### DIFF
--- a/base/status_utilities.hpp
+++ b/base/status_utilities.hpp
@@ -19,8 +19,12 @@ absl::Status const& GetStatus(absl::StatusOr<T> const& s) {
 
 }  // namespace internal
 
-#define CHECK_OK(value) CHECK_EQ((value), ::absl::OkStatus())
-#define DCHECK_OK(value) DCHECK_EQ((value), ::absl::OkStatus())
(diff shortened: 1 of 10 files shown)


eggrobin (approved):

pleroy:
I exposed the angular resolution passed to `Planetarium::Parameters` in the UI in https://github.com/pleroy/Principia/tree/ResolutionManipulate.  In SKL, artifacts appear fleetingly at 3 arc min and become annoying above 10 arc min.  This would correspond to a value of 2-3 cm for the metric computed in this test.

PR #4203, opened 2025-04-19
An algorithm to find all zeroes of a function over an interval

pleroy:
It separates the zeroes by looking for extrema.

### DIFF
--- a/numerics/root_finders.hpp
+++ b/numerics/root_finders.hpp
@@ -4,6 +4,7 @@
 #include <limits>
 #include <vector>
 
+#include "absl/container/btree_set.h"
 #include "base/array.hpp"
 #include "numerics/scale_b.hpp"
 #include "quantities/elementary_functions.hpp"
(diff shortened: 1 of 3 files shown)


numerics/root_finders.hpp
```diff
@@ -36,6 +37,19 @@ Argument Brent(Function f,
                Argument const& lower_bound,
                Argument const& upper_bound);
 
+// Alternatively applies Brent’s procedure `zero`  and Brent’s procedure
```
eggrobin, line 40:
```suggestion
// Alternatively applies Brent’s procedure `zero` and Brent’s procedure
```

numerics/root_finders.hpp
```diff
@@ -36,6 +37,19 @@ Argument Brent(Function f,
...
+// `localmin` (both from [Bre73]) to find all the zeroes over the interval by
+// separating them using the extrema.  `eps` must be significantly larger than
+// the default `eps` passed to `localmin`.  A value of `eps` that's too small
+// will cause the algorithm to loop forever.
```
eggrobin, line 44: Elaborate? How does it end up looping forever? How was 1000 picked?

eggrobin (approved):

PR #4204, opened 2025-04-19
Don't draw the same equipotential multiple times.

pleroy:
Consider the peak P `peaks[i]` and the well W `*expected_delineated_well`.  The code draws a straight line PW and searches a point with energy E.  In the old code, this was done using `Brent` and we were happy with returning a random intersection, say Q.  We would then draw the equipotential Γ going through Q.

The old code used to say "Γ obviously encloses W and therefore W is delineated from P".  But this is not true: there may be multiple points with energy E on the line PW, in which case we have no guarantee whatsoever with respect to the position of Γ with respect to W.  For instance, in the Earth-Moon system, if W is the Moon, there exist energies for which there are three equipotentials: one around the Moon, one around the Earth inside the Moon's orbit, and one around the Earth outside the Moon's orbit.

So we have to (1) compute all the points with energy E on the line P using `DoubleBrent`; (2) properly analyse Γ and check if it actually encloses W; and (3) only retain the equipotentials that create new delineations (that's necessary because multiple the points with energy E on PW may lie on the same equipotential).

This speeds things up a bit, from 24-25 to 26-27 fps in SKL.

The following pictures illustrate the changes: when zooming sufficiently, one sees that the "white" equipotentials (that were brighter because they were drawn multiple times) are now yellow (drawn only once).

At the خیام release:
before 1
With #4202 and #4204:
after 1

### DIFF
--- a/numerics/root_finders_body.hpp
+++ b/numerics/root_finders_body.hpp
@@ -167,8 +167,10 @@ absl::btree_set<Argument> DoubleBrent(Function f,
   Argument const a = lower_bound;
   Argument const b = upper_bound;
 
-  Argument const a_effective = a * (1 + eps);
-  Argument const b_effective = b * (1 - eps);
+  // These computations ensure a reasonable margin even if one of the bounds is
+  // zero.
+  Argument const a_effective = a + eps * std::max(Abs(a), Abs(b));
+  Argument const b_effective = b - eps * std::max(Abs(a), Abs(b));
 
   Value f_a = f(a);
   Value f_b = f(b);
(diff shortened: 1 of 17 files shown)


numerics/root_finders_body.hpp
```diff
@@ -164,8 +164,13 @@ absl::btree_set<Argument> DoubleBrent(Function f,
...
 
-  Argument a = lower_bound;
-  Argument b = upper_bound;
+  Argument const a = lower_bound;
+  Argument const b = upper_bound;
+
+  // These computations ensure a reasonable margin even if one of the bounds is
+  // zero.
+  Argument const a_effective = a + eps * std::max(Abs(a), Abs(b));
+  Argument const b_effective = b - eps * std::max(Abs(a), Abs(b));
```
eggrobin, line 173: Surely this function should work for an affine `Argument`.

1. Add a test with a function Instant -> double;
2. Change this to use eps * Abs(a - b), or if that doesn’t work use subtractions of Argument{} and explain why it needs to be aware of the position of the origin.

eggrobin (approved):

PR #4209, opened 2025-05-03
Handle "resource exhausted" in the flight planner UI

pleroy:
Reported on Discord by "mako":
image

### DIFF
--- a/ksp_plugin_adapter/flight_planner.cs
+++ b/ksp_plugin_adapter/flight_planner.cs
@@ -818,7 +818,7 @@ class FlightPlanner : VesselSupervisedWindowRenderer {
                               actual_final_time -
                               plugin.FlightPlanGetInitialTime(vessel_guid)))
                     : "";
-      if (status_.is_aborted()) {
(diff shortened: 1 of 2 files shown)


eggrobin (approved):

eggrobin:
Something is strange about this screenshot though: note that the error reports a max_steps value of 1000, not the 1024 from the UI.

Is it this 1000? https://github.com/mockingbirdnest/Principia/blob/b4b404a8913358a43452fb1fd87653e9c34bf781/ksp_plugin/integrators.cpp#L61

If so, is it really affected by increasing the setting that is at 1024?

On discord, mako noted that « changing the step count etc didn't change ».

pleroy:
Yes, we do a "fuzzy" search [here](https://github.com/mockingbirdnest/Principia/blob/b4b404a8913358a43452fb1fd87653e9c34bf781/ksp_plugin_adapter/flight_planner.cs#L259) so that we don't have to exactly synchronize the numbers on both sides.

> On discord, mako noted that « changing the step count etc didn't change », but it is unclear what exactly « the step count etc » refers to.

Shrug.  No reproduction steps = no bug.

PR #4211, opened 2025-05-03
A bunch of improvement that were made while looking at plotting

pleroy:
1. Remove an unused parameter in `PlotMethod3`.
2. Add support for `f@x` in *Mathematica*.
3. Speed up the test of `PlotMethod3` by avoiding repeated computations of the equipotentials.
4. Fix the serialization of `ConformalMap` when the scale is not a `double`.
5. Move the computation of angles to `R3Elements` to avoid code duplication.


mathematica/mathematica.hpp
```diff
@@ -122,10 +122,27 @@ constexpr auto ExpressInSIUnits = ExpressIn(si::Unit<Length>,
...
+
+ private:
+  std::string name_;
+};
+
+static_assert(std::is_same_v<int, tail_t<int>>);
+static_assert(std::is_same_v<int, tail_t<char*, int>>);
+
+static_assert(is_instance_of_v<ExpressIn,
+                              ExpressIn<Length,Mass,Time,Current,Temperature>>);
```
eggrobin, line 145: Remove the static_asserts.

eggrobin (approved):

PR #4230, opened 2025-06-15
Interrupt the analysers when recomputing a coast

### LINKED ISSUE #4216: Discrepancies between analysis windows
### Problem

When doing changes to the calculation length of the analysis windows for current orbit and planned orbits, I have noticed a discrepancy.

With the main window orbit analysis feature, changing the duration will immediately trigger a recalculation of the orbit. In the flight plan window, changing duration or even any of the flight plan properties, will ONLY start a new calculation if it wasn't already calculating. 

This is typically not an issue when the time span is short and the calculation time is short, but for slower computers (like mine, hah), when doing calculations for say 100 days, I have to wait for a full calculation before I can adjust the parameters for the maneouvre. I tend to use the mouse wheel for these changes, and if I accidentally make two "clicks" (two mouse wheel up/down events), the second event will not be calculated but the first has to finish.

To trigger a recalculation I have to wait for the planning analysis to finish, then trigger it via either changing the duration, or any other of the parameters.

### Suggested Fix
Make the flight plan analysis window restart calculations immediately whenever any change is detected.

As an alternative, it might be possible to work around if we had an "abort calculation" button in the flight plan analysis window.

pleroy:
Also change the way that we pop segments to make it easier to enforce the invariant that the flight plan starts and ends with a coast.

Fix #4216.

### DIFF
--- a/ksp_plugin/flight_plan.cpp
+++ b/ksp_plugin/flight_plan.cpp
@@ -532,7 +532,7 @@ absl::Status FlightPlan::ComputeSegments(
       if (analysis_is_enabled_) {
         auto const& [first_time, first_degrees_of_freedom] = coast->front();
         auto& analyser = coast_analysers_[it - manœuvres_.begin()];
-        // Interrupt the analyses for the coast that are being recomputed as we
+        // Interrupt the analyses for the coasts that are being recomputed as we
         // won't use their result.
         analyser->Interrupt();
         analyser->RequestAnalysis(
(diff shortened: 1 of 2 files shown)


ksp_plugin/flight_plan.cpp
```diff
@@ -536,7 +531,11 @@ absl::Status FlightPlan::ComputeSegments(
       }
       if (analysis_is_enabled_) {
         auto const& [first_time, first_degrees_of_freedom] = coast->front();
-        coast_analysers_[it - manœuvres_.begin()]->RequestAnalysis(
+        auto& analyser = coast_analysers_[it - manœuvres_.begin()];
+        // Interrupt the analyses for the coast that are being recomputed as we
```
eggrobin:
```suggestion
        // Interrupt the analyses for the coasts that are being recomputed as we
```

PR #4235, opened 2025-06-22
Add a wrapper, Exact, to indicate that the evaluation of an operation doesn't entail an error

pleroy:
We should probably do a better job with 0, powers of 2, and the like, but we'll surely need an explicit escape hatch in some cases.

#1760.


mathematica/ieee754_floating_point_evaluation_test.wl
```diff
@@ -296,6 +312,76 @@ End[]
 (*IEEEEvaluateWithRelativeError*)
 
 
+?IEEEEvaluateWithRelativeError
```
eggrobin, line 318: Could we have tests that use `Exact[`…`]` instead of the private applyOpWithRelativeError?

eggrobin (changes requested):

mathematica/ieee754_floating_point_evaluation_test.wl
pleroy: Done.  Amusingly enough, this uncovered a bug :-/

eggrobin (approved):

PR #4243, opened 2025-07-13
Fixes and improvements to the documentation based on Mathematica error analysis

pleroy:
#1760.


documentation/Sin Cos.tex
```diff
@@ -213,15 +214,62 @@ \subsection*{Argument Reduction Using the Two-Term Approximation}
...
-As long as $2^{\gk_1} \gg_3$ is small enough (less that $1/2$), the rounding cannot cause $n$ to exceed $2^{\gk_1}$.  In practice we choose a relatively small value for $\gk_1$, so this condition is met.
+If $2^{\gk_1} \gg_3$ is small enough (less that $1/2$), the rounding cannot cause $n$ to exceed $2^{\gk_1}$.  In practice we choose a relatively small value for $\gk_1$, so this condition is met.
+
+Now if $x$ is close to an odd multiple of $\frac{\Pi}{4}$ it is possible for misrounding to happen.  In the following analysis we assume that $n > 0$.  The results are symmetrical if $n < 0$.  There are two possible kinds of misrounding, with different bounds.
+
+A misrounding of the first kind occurs if:
+\[
+x < \pa{n - \frac{1}{2}}\frac{\Pi}{2} \quad \text{and} \quad \round{x \round{\frac{2}{\Pi}}} > n - \frac{1}{2}
+\]
+Using equation (\ref{eqnxerror}) we find that this misrounding is possible iff:
```
```suggestion
Using equation (\ref{eqnxerror}) we find that this misrounding is possible only if:
```
« X possible iff Y » is weird (it just means X occurs only if Y), but also we have two inequalities ; the second (looser) bound is not equivalent to the first.

documentation/Sin Cos.tex
```diff
@@ -213,15 +214,62 @@ \subsection*{Argument Reduction Using the Two-Term Approximation}
...
+\[
+x > \frac{\Pi}{2} \pa{n - \frac{1}{2}} \frac{1}{\pa{1 + \gd_2} \pa{1 + \gd_3}} \geq \frac{\Pi}{2} \pa{n - \frac{1}{2}} \frac{1}{\pa{1 + \gg_2}}
```
```suggestion
x > \frac{\Pi}{2} \pa{n - \frac{1}{2}} \frac{1}{\pa{1 + \gd_2} \pa{1 + \gd_3}} \geq \frac{\Pi}{2} \pa{n - \frac{1}{2}} \frac{1}{1 + \gg_2}
```

documentation/Sin Cos.tex
```diff
@@ -213,15 +214,62 @@ \subsection*{Argument Reduction Using the Two-Term Approximation}
...
+\label{eqnxerror}
+\end{equation}
+from which we deduce the upper bound:
 \begin{align*}
-\abs x &\leq 2^{\gk_1} \frac{\Pi}{2} \pa{1 + \gd_1} \\
 \abs n &\leq \iround{2^{\gk_1} \frac{\Pi}{2} \pa{1 + \gd_1} \frac{2}{\Pi} \pa{1 + \gd_2} \pa{1 + \gd_3}} \\
 &\leq \iround{2^{\gk_1} \pa{1 + \gg_3}}
 \end{align*}
-As long as $2^{\gk_1} \gg_3$ is small enough (less that $1/2$), the rounding cannot cause $n$ to exceed $2^{\gk_1}$.  In practice we choose a relatively small value for $\gk_1$, so this condition is met.
+If $2^{\gk_1} \gg_3$ is small enough (less that $1/2$), the rounding cannot cause $n$ to exceed $2^{\gk_1}$.  In practice we choose a relatively small value for $\gk_1$, so this condition is met.
```
```suggestion
If $2^{\gk_1} \gg_3$ is small enough (less than $1/2$), the rounding cannot cause $n$ to exceed $2^{\gk_1}$.  In practice we choose a relatively small value for $\gk_1$, so this condition is met.
```

documentation/Sin Cos.tex
```diff
@@ -236,7 +284,7 @@ \subsection*{Argument Reduction Using the Two-Term Approximation}
 Therefore, as long as $\gk'_1 > 2$, there exist arguments $x$ for which $\abs{\gd y} > \abs y$.
 } using algorithm 4 of \cite{HidaLiBailey2007}.
 
-To compute the overall error on argument reduction, first remember that, from equation (\ref{eqnpitwoterms}), we have:
+To compute the overall error on argument reduction\footnote{Note that this error analysis is correct even in the face of misrounding.  Misrounding can combine with the argument reduction error, though, to cause $\abs{y - {\gd y}}$ to move farther above $\frac{\Pi}{4}$}, first remember that, from equation (\ref{eqnpitwoterms}), we have:
```
```suggestion
To compute the overall error on argument reduction\footnote{Note that this error analysis is correct even in the face of misrounding.  Misrounding can combine with the argument reduction error, though, to cause $\abs{y - {\gd y}}$ to move farther above $\frac{\Pi}{4}$.}, first remember that, from equation (\ref{eqnpitwoterms}), we have:
```

documentation/Sin Cos.tex
pleroy: "only possible if", it's cleaner.

eggrobin (approved):

PR #4256, opened 2025-07-27
Increase the distance after setting focus to a celestial body

paulo-santana:
I would like to propose the following change to the distance set when using the new feature for setting focus to a celestial body in the reference frame list. This would cause the distance after focus to no longer be the smallest possible, but rather give some room to better observe the surroundings of the body in the map view.

This came from personally getting a little bit annoyed by having to zoom out every time after using the focus button. I really appreciate the feature though. It's been a life saver since I run too many simulations and have to recreate plans every time.

This was tested locally, the images show how the bodies would be drawn with the new value. I think the ideal would having it be a setting so the user may set to whatever is preferred

image
image

### DIFF
--- a/ksp_plugin_adapter/reference_frame_selector.cs
+++ b/ksp_plugin_adapter/reference_frame_selector.cs
@@ -641,7 +641,7 @@ internal class
               GUILayoutWidth(1))) {
         PlanetariumCamera.fetch.SetTarget(celestial);
         PlanetariumCamera.fetch.SetDistance(
-            (float)celestial.Radius * ScaledSpace.InverseScaleFactor * 2f);
+            (float)celestial.Radius * ScaledSpace.InverseScaleFactor * 2f * 4);
       }
     }
     if (!celestial.is_leaf(target)) {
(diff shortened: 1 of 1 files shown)


ksp_plugin_adapter/reference_frame_selector.cs
```diff
@@ -641,7 +641,7 @@ private void RenderSubtree(CelestialBody celestial, int depth) {
               GUILayoutWidth(1))) {
         PlanetariumCamera.fetch.SetTarget(celestial);
```
pleroy, line 642:
```suggestion
        PlanetariumCamera.fetch.SetTarget(celestial);
        const double degree = Math.PI / 180;
        const double apparent_size = 15 * degree;
```

ksp_plugin_adapter/reference_frame_selector.cs
```diff
@@ -641,7 +641,7 @@ private void RenderSubtree(CelestialBody celestial, int depth) {
...
         PlanetariumCamera.fetch.SetDistance(
-            (float)celestial.Radius * ScaledSpace.InverseScaleFactor * 2f);
+            (float)celestial.Radius * ScaledSpace.InverseScaleFactor * 2f * 4);
```
pleroy: Thanks for the change, which I agree will improve usability.  I would like it programmed a bit differently, as the factor 4 (or 8, really) looks a bit magic.
```suggestion
            (float)celestial.Radius * ScaledSpace.InverseScaleFactor /
                Math.Tan(apparent_size / 2));
```

pleroy (changes requested):

paulo-santana:
Applied the suggestions. I tested it and it works great.

PR #4269, opened 2025-08-10
Document the polynomials and the error analysis

pleroy:
Comes with a *Mathematica* file that helps with the computations.

#1760.


documentation/Sin Cos.tex
```diff
@@ -128,19 +134,23 @@ \subsubsection*{Two-Term Approximation}
 \abs{\frac{\Pi}{2} - C_1 - \gd C_1} \leq \frac{1}{2} \ULP\of{\frac{\Pi}{2} - C_1} = 2^{\gk'_1 - M - 1} \ULP\of{\frac{\Pi}{2}}
 \label{eqnpitwoterms}
 \end{align}
-and the following upper bound for $\gd C_1$:
+In practice this quantity is smaller than $2^{-103.217}$ (which is much tighter than the theoretical bound) because of the actual bit pattern of $\Pi$.
```
eggrobin: What does this sentence mean? Does « in practice » mean « for the value of κ₁ we actually choose » ? If so, say so and give it. What does « the theoretical bound » mean ? is it the case where κ₁′=κ₁? If so, give κ₁′ too.

Same with all the new « In practice » below.

pleroy: I tried to clarify.  « In practice » means « for the value of κ₁ we actually choose » and « using the actual value of π to do the computation ».  Basically the theoretical upper bound on the error assumes the worst with respect to the bits of π (e.g., plenty of ones after the last mantissa bit) but in reality runs of zeroes do happen in some cases and may yield an actual error significantly smaller than the theoretical bound.

documentation/Sin Cos.tex
```diff
@@ -128,19 +134,23 @@ \subsubsection*{Two-Term Approximation}
 \abs{\frac{\Pi}{2} - C_1 - \gd C_1} \leq \frac{1}{2} \ULP\of{\frac{\Pi}{2} - C_1} = 2^{\gk'_1 - M - 1} \ULP\of{\frac{\Pi}{2}}
 \label{eqnpitwoterms}
 \end{align}
-and the following upper bound for $\gd C_1$:
+With the chosen value $\gk_1 = 8$, the two sides of this inequality turn out to be $2^{-103.217}$ and $2^{-101}$.  The difference comes from the fact that $\abs{\frac{\Pi}{2} - C_1 - \gd C_1}$ has three zeroes after the last bit of its mantissa, but the theoretical computation above assumes the worse, i.e., a run of ones.  In the rest of our discussion of approximations of $\Pi$ we will give the (tighter) error bounds obtained by computing the left-hand side with the actual value of $\Pi$.
```
```suggestion
With the chosen value $\gk_1 = 8$, the two sides of this inequality turn out to be $2^{-103.217}$ and $2^{-101}$.  The difference comes from the fact that $\abs{\frac{\Pi}{2} - C_1 - \gd C_1}$ has three zeroes after the last bit of its mantissa, but the theoretical computation above assumes the worst, i.e., a run of ones.  In the rest of our discussion of approximations of $\Pi$ we will give the (tighter) error bounds obtained by computing the left-hand side with the actual value of $\Pi$.
```

PR #4272, opened 2025-08-13
Rationalize the handling of Boost multiprecision types

pleroy:
Adding concepts to represent the types that need special handling, and removing/simplifying the special-casing.


numerics/elementary_functions_body.hpp
```diff
@@ -2,82 +2,124 @@
...
+#include "quantities/cantor.hpp"
+#include "quantities/concepts.hpp"
 #include "quantities/si.hpp"
 
 namespace principia {
 namespace numerics {
 namespace _elementary_functions {
 namespace internal {
 
+using namespace boost::multiprecision;
```
eggrobin, line 20: What are we using here from boost::multiprecision? It looks like we used to use is_number but we don’t anymore…

quantities/cantor.hpp
```diff
@@ -0,0 +1,57 @@
...
+
+using namespace boost::multiprecision;
+using namespace principia::base::_traits;
+
+// The `cpp_` concepts should be used sparingly, and only in places where the
+// Boost multiprecision API differs from ours or from the standard C++ API.  At
+// any rate, the multiprecision traits should never be used directly.
+
+template<typename T>
+concept cpp_number = (is_number<T>::value || is_number_expression<T>::value) &&
```
eggrobin: The name is weird (and there are no comments stating what this actually is, only comments stating that it should not be used), I thought it meant a standard C++ numeric type.

Rename to boost_cpp_number or boost_multiprecision_number, or boost_mp_number, or something like that.

pleroy: Good idea.  Having concepts named exactly like types was problematic anyway as it was causing clashes.

PR #4275, opened 2025-08-17
Make Pow type-agnostic and move it back to numerics

eggrobin:
This makes the generators in quantities/generators.hpp specific to the definition of operations on Quantity that correspond to predefined operations. When we introduce our own functions that work on Quantity and other things (e.g., `Pow`), we put that in `numerics` (why would `Pow` on a boost multiprecision type live in `quantities`?).

`Exponentiation` works like `Product` and does not know anything about how `Quantity` works (and thus likewise for `Derivatives` etc.).
`Pow` returns an `Exponentiation` and thus does not need to know anything about how `Quantity` works (just like other templatized numerics, e.g., `DoublePrecision`.

The headers that define constants depend on numerics::elementary_functions.

### DIFF
--- a/numerics/elementary_functions_body.hpp
+++ b/numerics/elementary_functions_body.hpp
@@ -117,6 +117,62 @@ constexpr Q NextDown(Q const& x) {
   return si::Unit<Q> * numerics::_next::NextDown(x / si::Unit<Q>);
 }
 
+template<int exponent>
+constexpr double Pow(double x) {
+  // Use the Russian peasant algorithm for small exponents.
+  if constexpr (exponent > 0 && exponent < 32) {
+    // The end of the recursion is handled by the specializations below.
+    auto const y = Pow<exponent / 2>(x);
+    auto const y² = y * y;
+    if constexpr (exponent % 2 == 1) {
+      return y² * x;
+    } else {
+      return y²;
+    }
+  } else if constexpr (exponent < 0 && exponent > -32) {
+    return 1 / Pow<-exponent>(x);
+  } else {
+    return std::pow(x, exponent);
+  }
+}
+
+// Static specializations for frequently-used exponents, so that this gets
+// turned into multiplications at compile time.
+
+template<>
+inline constexpr double Pow<0>(double) {
+  return 1;
+}
+
+template<>
+inline constexpr double Pow<1>(double x) {
+  return x;
+}
+
+template<>
+inline constexpr double Pow<2>(double x) {
+  return x * x;
+}
+
+template<>
+inline constexpr double Pow<3>(double x) {
+  return x * x * x;
+}
+
+template<int exponent, typename Q>
+constexpr Exponentiation<Q, exponent> Pow(Q const& x) {
+  if constexpr (number_category<Q>::value == number_kind_rational) {
+    // It seems that Boost does not define `pow` for `cpp_rational`.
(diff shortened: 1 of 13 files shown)


physics/lagrange_equipotentials_test.cpp
```diff
@@ -129,7 +127,8 @@ TEST_F(LagrangeEquipotentialsTest,
   Position<World> const q_moon = moon_world_dof.position();
   Position<World> const initial_earth_moon_l5 =
       Barycentre({q_earth, q_moon}, {1.0, 1.0}) +
-      (q_earth - q_moon).Norm() * Vector<double, World>({0, Sqrt(3) / 2, 0});
+      (q_earth - q_moon).Norm() *
+          Vector<double, World>({0, quantities::Sqrt(3) / 2, 0});
```
pleroy: I have no idea what is going on here, because I don't see an `Sqrt` defined anywhere in `quantities`.  But anyway we surely should not have to write `quantities` explicitly when all we need is √3.  Basically the old code looked just fine to me.

quantities/named_quantities.hpp
```diff
@@ -23,8 +24,23 @@ template<typename Q>
 using Inverse = Quotient<double, Q>;
 
 template<typename T, int exponent>
-using Exponentiation =
-    typename _generators::ExponentiationGenerator<T, exponent>::Type;
+struct ExponentiationGenerator : not_constructible {
```
pleroy, line 27: Having to define a generator in this header is polluting it and breaking the separation of concern between the files.

But it's easy to fix I think: just move it back to `generators.hpp`, `_body.hpp` and pass `Product` and `Quotient` as two template template parameters.

It makes a lot of sense conceptually too: in order to define exponentiation, you need to tell me how you multiply and divide.

numerics/elementary_functions_body.hpp
```diff
@@ -117,6 +117,62 @@ constexpr Q NextDown(Q const& x) {
...
+  return x * x;
+}
+
+template<>
+inline constexpr double Pow<3>(double x) {
+  return x * x * x;
+}
+
+template<int exponent, typename Q>
+constexpr Exponentiation<Q, exponent> Pow(Q const& x) {
```
pleroy, line 163: FWIW, this function is horrible and I was trying to clean it up in #4272.  A discussion for another day.

pleroy (changes requested):

physics/lagrange_equipotentials_test.cpp
eggrobin: Artefact of the `revert` on a commit that did more than it said, I think.

quantities/named_quantities.hpp
eggrobin: It is not that kind of generator anymore; it is more like DerivativesGenerator.

pleroy: At least, hide it in the body.  We don't put the implementation in the headers.

Also, what is the final goal?  Will this move to another file?  To another project?

eggrobin: > what is the final goal? Will this move to another file? To another project?

Maybe. (Like Product, it is not actually quantities-specific.) But not today. At least it can move now.

eggrobin: As discussed, we cannot hide this in a body without moving it. I will move it in a subsequent pull request.

PR #4276, opened 2025-08-17
A ring requires an identity

pleroy:


geometry/concepts.hpp
```diff
@@ -52,17 +53,25 @@ concept homogeneous_ring =
...
 template<typename A>
-concept ring = homogeneous_ring<A> && requires(A x, A y) {
+concept pseudo_ring = homogeneous_pseudo_ring<A> && requires(A x, A y) {
   { x * y } -> std::same_as<A>;
   { x *= y } -> std::same_as<A&>;
 };
 
+// A pseudo-ring with a multiplicative identity.
+template<typename P>
+concept homogeneous_ring = homogeneous_pseudo_ring<P>;
```
eggrobin: We can’t check or distinguish those, so let’s keep only the pseudo- one.

numerics/fixed_arrays_test.cpp
```diff
@@ -73,6 +73,7 @@ TEST_F(FixedArraysTest, AlgebraConcepts) {
   static_assert(homogeneous_ring<FixedMatrix<Length, 3, 3>>);
 
   static_assert(ring<FixedMatrix<double, 3, 3>>);
+  static_assert(!ring<FixedMatrix<double, 2, 3>>);
```
eggrobin, line 75: Keep them in the same order as above, thus swap lines 75 and 76.

eggrobin (approved):

PR #3931, opened 2024-03-29
Test Reach’s save without FMA

eggrobin:
This brings back two flybys.

PR #4286, opened 2025-08-31
Mark continuous trajectories as WithoutFMA if created or read on a machine without FMA

eggrobin:
Currently, a save created with Grossmann or later on a machine without FMA will change behaviour if transferred to a machine with FMA, potentially wrecking complex flight plans (see, e.g., #3931). With this change, opening a save on a machine without FMA will forever mark the continuous trajectories as not having FMA, thus work done on the machine without FMA should be readable on a machine with FMA.

Note that the other direction is not supported: a flight plan created on a machine with FMA may be broken if you open it on an old machine.

### DIFF
--- a/physics/continuous_trajectory_body.hpp
+++ b/physics/continuous_trajectory_body.hpp
@@ -12,6 +12,7 @@
 
 #include "geometry/interval.hpp"
 #include "glog/stl_logging.h"
+#include "numerics/fma.hpp"
 #include "numerics/newhall.hpp"
 #include "numerics/poisson_series.hpp"
 #include "numerics/polynomial_in_чебышёв_basis.hpp"
@@ -23,8 +24,8 @@ namespace physics {
 namespace _continuous_trajectory {
 namespace internal {
 
-using namespace principia::numerics::_fma;
 using namespace principia::geometry::_interval;
+using namespace principia::numerics::_fma;
 using namespace principia::numerics::_newhall;
 using namespace principia::numerics::_poisson_series;
 using namespace principia::numerics::_polynomial_in_чебышёв_basis;
--- a/physics/continuous_trajectory_body.hpp
+++ b/physics/continuous_trajectory_body.hpp
@@ -23,6 +23,7 @@ namespace physics {
 namespace _continuous_trajectory {
 namespace internal {
 
+using namespace principia::numerics::_fma;
 using namespace principia::geometry::_interval;
 using namespace principia::numerics::_newhall;
 using namespace principia::numerics::_poisson_series;
@@ -51,7 +52,9 @@ ContinuousTrajectory<Frame>::ContinuousTrajectory(Time const& step,
       is_unstable_(false),
       degree_(min_degree),
       degree_age_(0),
-      polynomial_evaluator_policy_(Policy::AlwaysEstrin()) {
+      polynomial_evaluator_policy_(UseHardwareFMA
+                                       ? Policy::AlwaysEstrin()
+                                       : Policy::AlwaysEstrinWithoutFMA()) {
   CHECK_LT(0 * Metre, tolerance_);
 }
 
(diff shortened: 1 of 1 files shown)


physics/continuous_trajectory_body.hpp
```diff
@@ -442,32 +446,66 @@ ContinuousTrajectory<Frame>::ReadFromMessage(
...
+            // that transition, whatever work is done on the old machine will be
+            // faithfully reproduced when it is read again on a new machine.
+            auto const evaluator_kind =
+                polynomial
+                    ->GetExtension(
+                        serialization::PolynomialInMonomialBasis::extension)
+                    .evaluator()
+                    .kind();
+            if (evaluator_kind ==
+                serialization::PolynomialInMonomialBasis::Evaluator::ESTRIN) {
```
pleroy: What about `HORNER`?  We might not use it at the moment, but deserialization must be correct.

physics/continuous_trajectory_body.hpp
```diff
@@ -442,32 +446,66 @@ ContinuousTrajectory<Frame>::ReadFromMessage(
             Polynomial<Position<Frame>, Instant>::template ReadFromMessage<
                 EstrinWithoutFMA>(polynomial));
       } else {
-        serialization::Polynomial const& polynomial = pair.polynomial();
-        if (polynomial.HasExtension(
-                serialization::PolynomialInMonomialBasis::extension) &&
+        not_null<serialization::Polynomial const*> polynomial =
```
pleroy: I find this pointer confusing as I have to wonder if it's going to dangle.  Let's go back to a reference and see my comment below.

physics/continuous_trajectory_body.hpp
```diff
@@ -442,32 +446,66 @@ ContinuousTrajectory<Frame>::ReadFromMessage(
...
-              Polynomial<Position<Frame>, Instant>::ReadFromMessage(
-                  polynomial));
+              UseHardwareFMA ? Polynomial<Position<Frame>, Instant>::
+                                   template ReadFromMessage<Estrin>(*polynomial)
+                             : Polynomial<Position<Frame>, Instant>::
+                                   template ReadFromMessage<EstrinWithoutFMA>(
+                                       *polynomial));
         } else {
-          // The pre-Καραθεοδωρή path.
+          std::optional<serialization::Polynomial> rewritten_polynomial;
```
pleroy: Let's not try to be smart in saving proto assignments.  Just do:
```
auto rewritten_polynomial = polynomial;
```
update it as needed below and then at line 501 just use `rewritten_polynomial`.

physics/continuous_trajectory_body.hpp
```diff
@@ -479,33 +478,42 @@ ContinuousTrajectory<Frame>::ReadFromMessage(
...
+              case Evaluator::ESTRIN:
+                rewritten_evaluator.set_kind(Evaluator::ESTRIN_WITHOUT_FMA);
+                break;
+              case Evaluator::HORNER:
+                rewritten_evaluator.set_kind(Evaluator::HORNER_WITHOUT_FMA);
+                break;
+              case Evaluator::HORNER_WITHOUT_FMA:
+              case Evaluator::ESTRIN_WITHOUT_FMA:
+                break;
+              default:
```
pleroy: Let's remove the `default`, the switch statement is covered, I think.

physics/continuous_trajectory_body.hpp
```diff
@@ -446,26 +446,25 @@ ContinuousTrajectory<Frame>::ReadFromMessage(
...
+                  ? Polynomial<Position<Frame>, Instant>::
+                        template ReadFromMessage<Estrin>(polynomial)
+                  : Polynomial<Position<Frame>, Instant>::
+                        template ReadFromMessage<EstrinWithoutFMA>(polynomial));
         } else {
-          std::optional<serialization::Polynomial> rewritten_polynomial;
-          if (!UseHardwareFMA) {
+          auto rewritten_polynomial = polynomial;
+          if (!CanUseHardwareFMA) {
             // If we are on a machine without FMA, turn Estrin into
```
pleroy: Fix the comment too.

Issue #3019, opened 2021-05-31
Use AVX if available


eggrobin:
Because of different alignment requirements and VEX encoding/VZEROUPPER woes, and because it would be so low-level and pervasive (we cannot gate every operation on R3Element behind a CPUID test) we need a separate DLL for that, with an intermediate layer that chooses which one to use based on CPUID.

We could generate the intermediate layer from journal.proto, and LoadLibrary/GetProcAddress etc.

This blocks FMA usage on Linux & macOS, see https://github.com/mockingbirdnest/Principia/pull/3010#issuecomment-851077178.

Growflavor:
Would this only be for Linux & macOS?  Or might there be the possibility of an implementation as an alternative to FMA3 on Windows 10 machines?  For users like me who are stuck (for a few more years) with Ivy-Bridge processors that support AVX but not FMA3/AVX2:
![IvyBridge](https://user-images.githubusercontent.com/1580559/132951648-781cbd21-ff3d-45d8-bc5e-1e7372309aaa.png)

PR #4308, opened 2025-09-15
Manual DLL loading

eggrobin:
Towards #3019.

Tested in game on Windows and mac.


tools/journal_proto_processor.cpp
```diff
@@ -1200,6 +1205,8 @@ void JournalProtoProcessor::ProcessInOut(
   }
 
   cs_interface_parameters_[descriptor].clear();
+  cs_interface_marshalled_parameters_[descriptor].clear();
```
pleroy: We spell this word "marshaled" because that's what the .Net documentation does (see https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.marshalasattribute?view=net-9.0).

tools/journal_proto_processor.cpp
```diff
@@ -1891,18 +1920,46 @@ void JournalProtoProcessor::ProcessMethodExtension(
...
+  cs_interface_method_declaration_[descriptor] +=
+      Join({"    [UnmanagedFunctionPointer(CallingConvention.Cdecl)]",
+            cs_interface_return_marshal,
+            "public delegate " + cs_interface_return_type + " " + name +
+                "Delegate("},
+           "\n    ");
+  if (!cs_interface_marshalled_parameters.empty()) {
+    cs_interface_method_declaration_[descriptor] +=
+        "\n      " + Join(cs_interface_marshalled_parameters,
+                        /*joiner=*/",\n      ");  // NOLINT
```
pleroy: Bad indent.  Use clang format on the new code.  (But don't mess with the existing code.)

tools/journal_proto_processor.hpp
```diff
@@ -212,6 +212,12 @@ class JournalProtoProcessor final {
            std::function<std::string(std::string const& type)>>
       field_cxx_mode_fn_;
 
+  // For all fields, a lambda that takes a C# parameter type as stored in
+  // `field_cs_type_`, and adds `this` to it if needed.
```
pleroy: Could we have a higher-level description of this method?  It adds "this" because...?

Also, it's probably irrelevant where it gets the C# parameter type from.  It should just say that it take a C# type for a parameter (parameter of what?).  Also document the key of the map.

tools/journal_proto_processor.cpp
```diff
@@ -1891,18 +1920,46 @@ void JournalProtoProcessor::ProcessMethodExtension(
...
+  }
+  cs_interface_method_declaration_[descriptor] += ");\n";
+  cs_interface_method_declaration_[descriptor] +=
+      "    public " + name + "Delegate " + name + " =\n";
+  cs_interface_method_declaration_[descriptor] +=
+      "          Loader.LoadFunction<" + name + "Delegate>(\n";
```
pleroy: Remove two spaces from the indent (in the generated code).

tools/journal_proto_processor.cpp
```diff
@@ -1891,18 +1920,46 @@ void JournalProtoProcessor::ProcessMethodExtension(
...
+  cs_interface_method_declaration_[descriptor] +=
+      "              \"principia__" + name + "\");\n";
```
pleroy: Same.

tools/journal_proto_processor.cpp
```diff
@@ -1891,18 +1920,46 @@ void JournalProtoProcessor::ProcessMethodExtension(
...
   }
-  cs_interface_method_declaration_[descriptor] += ");\n\n";
+  cs_interface_method_declaration_[descriptor] += ") {\n";
+  cs_interface_method_declaration_[descriptor] +=
+      (cs_interface_return_type == "void" ? "    symbols_."
+                                          : "    return symbols_.") +
+      name + "(" +
+      (cs_interface_arguments.empty()
+           ? ""
+           : "\n        " +
```
pleroy: Remove two spaces from the indent of the arguments.

tools/journal_proto_processor.hpp
```diff
@@ -287,6 +293,10 @@ class JournalProtoProcessor final {
   // messages.
   std::map<Descriptor const*, std::vector<std::string>>
       cs_interface_parameters_;
+  std::map<Descriptor const*, std::vector<std::string>>
```
pleroy, line 303: Comment these maps.  You don't get to piggyback on the comment for the preceding map.

tools/journal_proto_processor.cpp
```diff
@@ -1891,18 +1920,46 @@ void JournalProtoProcessor::ProcessMethodExtension(
...
   cxx_functions_implementation_[descriptor] += cxx_run_epilog + "}\n\n";
 
-  cs_interface_method_declaration_[descriptor] = Join(
-      {"  [DllImport(dllName           : dll_path,\n"
-       "             EntryPoint        = \"principia__" + name + "\",\n"
-       "             CallingConvention = CallingConvention.Cdecl)]",
-       cs_interface_return_marshal,
-       "internal static extern " + cs_interface_return_type + " " + name + "("},
-      "\n  ");
+  cs_interface_method_declaration_[descriptor] =
```
pleroy: This needs a new identifier because, believe it or not, it's not a method anymore (hint: the word `class`).  I object to just lumping text in this variable because it's convenient.  Let's keep the data structures moderately clean.  (And yes, some new code will be needed to output this new map.)

I would by the way be fine with `<all the partial classes>` `<all the interface methods>` in the generated code.  I don't think there is value to glueing the two together.

tools/journal_proto_processor.cpp
```diff
@@ -1891,18 +1920,46 @@ void JournalProtoProcessor::ProcessMethodExtension(
...
+  cs_interface_method_declaration_[descriptor] +=
+      "    public " + name + "Delegate " + name + " =\n";
+  cs_interface_method_declaration_[descriptor] +=
+      "          Loader.LoadFunction<" + name + "Delegate>(\n";
+  cs_interface_method_declaration_[descriptor] +=
+      "              \"principia__" + name + "\");\n";
+  cs_interface_method_declaration_[descriptor] +=
+      "  }\n";
+
+  cs_interface_method_declaration_[descriptor] +=
```
pleroy: This one is actually the interface method.  Let's not lump it with the partial class.

tools/journal_proto_processor.hpp
```diff
@@ -212,6 +212,12 @@ class JournalProtoProcessor final {
            std::function<std::string(std::string const& type)>>
       field_cxx_mode_fn_;
 
+  // For all fields, a lambda that takes a C# parameter type as stored in
+  // `field_cs_type_`, and adds `this` to it if needed.
+  std::map<FieldDescriptor const*,
+           std::function<std::string(std::string const& type)>>
+      field_cs_extension_method_fn_;
```
pleroy: This identifier is not great, it makes it sound like this is a method when it is really a parameter name.

tools/journal_proto_processor.hpp
eggrobin: Added wording about extension methods. The rest of the comment is consistent with field_cxx_mode_fn_.

tools/journal_proto_processor.cpp
eggrobin: Well, if they are lumped together, they do not need to be separate partial classes.

eggrobin: Split the maps but kept the interleaving.

tools/journal_proto_processor.hpp
```diff
@@ -213,7 +214,8 @@ class JournalProtoProcessor final {
       field_cxx_mode_fn_;
 
   // For all fields, a lambda that takes a C# parameter type as stored in
-  // `field_cs_type_`, and adds `this` to it if needed.
+  // `field_cs_type_`, and adds `this` to it if `is_subject`; the result can be
+  // used to declare an parameter in an extension method.
```
pleroy:
```suggestion
  // used to declare a parameter in an extension method.
```

PR #4359, opened 2025-12-15
Add an option to generate interfacing code without journaling

pleroy:
This is useful for methods for which journaling would be incorrect or pointless.

Apparently in #797 `InitGoogleLogging` was generated and it was special-cased in #999.

#3019.

### DIFF
--- a/ksp_plugin_adapter/interface.cs
+++ b/ksp_plugin_adapter/interface.cs
@@ -281,8 +281,6 @@ internal static partial class Interface {
     };
   }
 
-  private partial class Symbols {}
-
   private static Symbols symbols_ = null;
 
   internal static void LoadSymbols() {


ksp_plugin_adapter/interface.cs
```diff
@@ -281,31 +281,7 @@ internal static KeplerianElements Elements(this Orbit orbit) {
...
-  }
-
-  internal static void ActivateRecorder(bool activate) {
-    symbols_.ActivateRecorder(activate);
-  }
-
-  internal static void InitGoogleLogging() {
-    symbols_.InitGoogleLogging();
-  }
+  private partial class Symbols {}
```
eggrobin:
```suggestion
```

PR #4360, opened 2025-12-16
Support for building and loading AVX and non-AVX DLLs

### LINKED ISSUE #3019: Use AVX if available

pleroy:
Fix #3019.

### DIFF
--- a/astronomy/orbital_elements_test.cpp
+++ b/astronomy/orbital_elements_test.cpp
@@ -398,7 +398,8 @@ TEST_F(OrbitalElementsTest, RealPerturbation) {
   // expected; the longitude of the ascending node Ω exhibits no obvious
   // precession even if its daily oscillation is filtered out.
   EXPECT_THAT(elements.mean_semimajor_axis_interval().measure(),
-              IsNear(20_(1) * Metre));
+              AnyOf(IsNear(20_(1) * Metre),    // Windows, Ubuntu SSE, macOS.
+                    IsNear(21_(1) * Metre)));  // Ubuntu AVX.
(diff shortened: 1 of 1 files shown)


ksp_plugin_adapter/loader.cs
```diff
@@ -81,33 +64,44 @@ internal static string LoadPrincipiaDllAndInitGoogleLogging() {
...
+      Interface.GetCPUIDFeatureFlags(out bool has_avx, out bool has_fma);
+      if (has_fma && !has_avx) {
+        return "Principia does not run on processors with FMA support but no " +
+               "AVX support.";
+      } else if (has_fma) {
+        // If it turns out that the machine has FMA (and therefore AVX), change
+        // our mind and load the x64_AVX_FMA DLL.  It's necessary to unload the
+        // DLL, otherwise the base address reported by `InitGoogleLogging` might
+        // (incorrectly) be the one of the `x64` DLL.
+        UnloadPrincipiaDll(platform: "x64");
```
eggrobin: Why are we discarding the error message from UnloadPrincipiaDll?

PR #4366, opened 2025-12-19
Move the multiprecision concepts to their own header, use numbers with et_off

eggrobin:
This has no effect on the performance of AccurateTableGeneratorTest, and it makes these types compatible with algebraic concepts.

### DIFF
--- a/mathematica/mathematica_body.hpp
+++ b/mathematica/mathematica_body.hpp
@@ -397,7 +397,7 @@ std::string ToMathematica(cpp_bin_float<digits> const& cpp_bin_float_number,
   using Float = cpp_bin_float<digits>;
   using Int = cpp_int;
   return ToMathematica<Float, Int>(
-      cpp_bin_float,
+      cpp_bin_float_number,
       base,
       [](Float const& x) { return abs(x); },
       [](Float const& x) { return ilogb(x); },


base/base.vcxproj
```diff
@@ -84,7 +85,7 @@
     <ClCompile Include="bits_test.cpp" />
     <ClCompile Include="bundle.cpp" />
     <ClCompile Include="bundle_test.cpp" />
-    <ClCompile Include="concepts_test.cpp" />
+    <ClCompile Include="multiprecision_test.cpp" />
```
pleroy: Alphabetize?

base/base.vcxproj.filters
```diff
@@ -224,6 +224,9 @@
     <ClInclude Include="concepts.hpp">
       <Filter>Header Files</Filter>
     </ClInclude>
+    <ClInclude Include="multiprecision.hpp">
+      <Filter>Header Files</Filter>
+    </ClInclude>
```
pleroy, line 229: Same.

base/base.vcxproj.filters
```diff
@@ -304,7 +307,7 @@
     <ClCompile Include="traits_test.cpp">
       <Filter>Test Files</Filter>
     </ClCompile>
-    <ClCompile Include="concepts_test.cpp">
+    <ClCompile Include="multiprecision_test.cpp">
```
pleroy, line 310: Same.

base/concepts.hpp
```diff
@@ -1,13 +1,8 @@
 #pragma once
 
-#include "boost/multiprecision/number.hpp"
-
 namespace principia {
 namespace base {
 namespace _concepts {
-namespace internal {
```
pleroy, line 8: Let's keep the internal namespace, I'm not sure how the Sourcerer reacts when it is missing.

base/multiprecision.hpp
```diff
@@ -0,0 +1,54 @@
+#pragma once
+
+#include "boost/multiprecision/number.hpp"
+
+namespace principia {
+namespace base {
+namespace _multiprecision {
```
pleroy, line 9: Same: `internal` namespace, using declarations in the non-internal namespace.

base/multiprecision_test.cpp
```diff
@@ -1,4 +1,4 @@
-#include "base/concepts.hpp"
+#include "base/multiprecision.hpp"
 
 #include "boost/multiprecision/cpp_bin_float.hpp"
```
pleroy: I would think that these headers are no longer needed; same comment in many places.  If they are actually needed, it probably tells you that `multiprecision.hpp` is missing includes.

mathematica/mathematica_body.hpp
```diff
@@ -390,13 +390,12 @@ std::string ToMathematica(T const real,
       [](T const& x) { return std::int64_t(x); });
 }
 
-template<unsigned digits,
-         typename OptionalExpressIn>
+template<typename OptionalExpressIn>
 std::string ToMathematica(
-    number<backends::cpp_bin_float<digits>> const& cpp_bin_float,
+    cpp_bin_float_50 const& cpp_bin_float,
```
pleroy: Why don't we have our own `cpp_bin_float` generic?  This change is going to annoy me when I want to output a `cpp_bin_float_100`.

base/base.vcxproj
eggrobin: Done.

base/base.vcxproj.filters
eggrobin: This one is generated from the UI and has never been alphabetized, let’s not start playing with it now.

PR #4368, opened 2025-12-20
Move quantities/arithmetic and geometry/concepts into base/algebra

eggrobin:
No changes to the concepts.

### DIFF
--- a/base/algebra.hpp
+++ b/base/algebra.hpp
@@ -65,8 +65,8 @@ concept additive_group = requires(G x, G y, int n) {
   { x - y } -> std::same_as<G>;
   { x += y } -> std::same_as<G&>;
   { x -= y } -> std::same_as<G&>;
-  // An abelian group is a ℤ-module; we require the corresponding
-  // operations. Note that `std::integral`, not `int`, should be used when
+  // An abelian group is a ℤ-module; we require the corresponding operations.
+  // Note that `std::integral`, not `int`, should be used when
   // implementing these operations to avoid implicit conversions from
   // `double`.
   { n * x } -> std::same_as<G>;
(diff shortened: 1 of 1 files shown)


base/algebra.hpp
```diff
@@ -24,9 +65,10 @@ concept additive_group = requires(G x, G y, int n) {
   { x - y } -> std::same_as<G>;
   { x += y } -> std::same_as<G&>;
   { x -= y } -> std::same_as<G&>;
-  // An abelian group is a ℤ-module; we require the corresponding operations.
-  // Note that `std::integral`, not `int`, should be used when implementing
-  // these operations to avoid implicit conversions from `double`.
+  // An abelian group is a ℤ-module; we require the corresponding
```
pleroy: Why the formatting change?  Same question in multiple places.

base/algebra.hpp
```diff
@@ -122,25 +165,37 @@ template<typename V>
 concept real_affine_space = affine_space<V, double>;
 
 template<typename T1, typename T2>
-concept hilbert = requires(T1 const& t1, T2 const& t2) {
-  InnerProduct(t1, t2);
-};
+concept hilbert =
+    requires(T1 const& t1, T2 const& t2) { InnerProduct(t1, t2); };
```
pleroy: Let's keep the previous formatting, it was easier to read.

base/algebra_body.hpp
```diff
@@ -1,10 +1,10 @@
...
+#include "base/algebra.hpp"
 
 namespace principia {
-namespace quantities {
-namespace _arithmetic {
+namespace base {
+namespace _algebra {
 namespace internal {
 
 template<typename T, int exponent>
```
pleroy, line 10: For the record, it makes me a bit uncomfortable to have generators both in `base` and in `quantities`.  We'll see if it causes trouble.

PR #4381, opened 2025-12-31
Add KSP 1.12.4 to KSP Assemblies HintPaths.

rnlahaye:
There doesn't seem to be a 1.12.5 for macOS.


pleroy:
I presume that it's only a problem for people who want to build Principia, right?  End-users wouldn't be affected?

rnlahaye:
Yes, and in particular only people who want to build the adapter.

Unfortunately it looks from the failing build that this change interferes with the normal build :/

pleroy:
According to ~~The Supreme Authority~~ [StackOverflow](https://stackoverflow.com/questions/15666618/csproj-multiple-hint-paths-for-an-assembly) you need to have conditions that check if the path exists in your `HintPath`.

pleroy:
I'll wait a bit before merging this as I am doing major surgery to the build and everything is broken at the moment.

pleroy:
@rnlahaye Our build is in good shape now, so I am merging.

PR #4396, opened 2026-01-04
Constraints for PolynomialInЧебышёвBasis

eggrobin:


numerics/polynomial_in_чебышёв_basis_body.hpp
```diff
@@ -129,9 +132,10 @@ PolynomialInЧебышёвBasis<Value_, Argument_, std::nullopt>::ReadFromMessag
 
 #undef PRINCIPIA_POLYNOMIAL_IN_ЧЕБЫШЁВ_BASIS_DESERIALIZATION_DEGREE
 
-template<typename Value_, typename Argument_, int degree_>
+template<affine Value_, affine Argument_, int degree_>
+  requires homogeneous_vector_space<Value_, Difference<Argument_>>
 constexpr PolynomialInЧебышёвBasis<Value_, Argument_, degree_>::
-PolynomialInЧебышёвBasis(Coefficients coefficients,
+    PolynomialInЧебышёвBasis(Coefficients coefficients,
```
pleroy: Don't touch the indent.

numerics/polynomial_in_чебышёв_basis.hpp
```diff
@@ -104,7 +105,8 @@ class PolynomialInЧебышёвBasis<Value_, Argument_, std::nullopt>
   virtual absl::btree_set<Argument> RealRootsOrDie(double ε) const = 0;
 };
 
-template<typename Value_, typename Argument_, int degree_>
+template<affine Value_, affine Argument_, int degree_>
+  requires homogeneous_vector_space<Value_, Difference<Argument_>>
```
pleroy: Why do we say `affine` in the template and then immediately "actually, I meant vector space"?  Is this because of inheritance?

eggrobin: We can say additive_group instead of affine. We cannot say vector_space initially because that involves both types.

pleroy (approved):

PR #4412, opened 2026-01-23
Change polynomials to support meaningful operations when the value is not in a vector space

pleroy:
This requires to separate addition and subtraction in `cartesian_product` because the former is not allowed for affine spaces.

### DIFF
--- a/numerics/polynomial_in_monomial_basis.hpp
+++ b/numerics/polynomial_in_monomial_basis.hpp
@@ -385,7 +385,7 @@ operator-(PolynomialInMonomialBasis<Value, Argument, ldegree_,
 // TODO(phl): The operators below work fine when `Value` is a vector space, but
 // they are bogus when it is merely an affine space.  For instance, multiplying
 // by a scalar a polynomial whose `Value` is a `Position` makes no sense.  They
-// do not compile in that case, but the compilation error is inscrutable.  We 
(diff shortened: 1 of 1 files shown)


numerics/polynomial_in_monomial_basis.hpp
```diff
@@ -345,17 +355,40 @@ operator+(PolynomialInMonomialBasis<Value, Argument, ldegree_,
...
+                                    std::max(ldegree_, rdegree_), Evaluator_>
 operator-(PolynomialInMonomialBasis<Value, Argument, ldegree_,
                                     Evaluator_> const& left,
           PolynomialInMonomialBasis<Value, Argument, rdegree_,
                                     Evaluator_> const& right);
 
+// TODO(phl): The operators below work fine when `Value` is a vector space, but
+// they are bogus when it is merely an affine space.  For instance, multiplying
+// by a scalar a polynomial whose `Value` is a `Position` makes no sense.  We
+// need to fix this.
```
eggrobin, line 389:
```suggestion
// they are bogus when it is merely an affine space.  For instance, multiplying
// by a scalar a polynomial whose `Value` is a `Position` makes no sense.  They
// do not compile in that case, but the compilation error is inscrutable.  We 
// need to fix this.
```

Issue #2961, opened 2021-04-24
Downsampling is slow


pleroy:
There has to be a better way to approximate a trajectory by a low-degree polynomial.  The current implementation does *O*(N Ln N) evaluations of the cubic sequentially.

rnlahaye:
I can see two approaches:

1. The error in a cubic approximation can be bounded using the fourth derivative of the underlying function. Using this bound, a semi-optimal cubic spline can be constructed in O(N) time. Of course this bound is not exact and the constructed spline might not be optimal.
2. If you are willing to restrict the segments of the spline to powers of 2 (i.e. the endpoints would be of the form (k*2^i, (k + 1)*2^i)), an optimal spline can be constructed in O(N log log N) time.

I'd be happy to send a PR for either of them.

rnlahaye:
Note for the first approach that there is a nice bound for cubic Hermite splines in particular; see Lemma 1 from [this paper](https://core.ac.uk/download/pdf/81960854.pdf) by Carlson and Hall.

pleroy:
Note that I am not at all sure that a cubic is really what we want, it was just convenient when this code was implemented.  Once you have specified continuity of the position and velocity at the endpoints, a cubic doesn't give you many degrees of freedom (understatement!).  A higher degree polynomial, like we use for continuous trajectory, might be preferable.

If we look at the problem as trying to find a point in a suitable function space that is "close enough" to the trajectory, I have a hunch that we could be sublinear as long as we are not looking for the closest possible match.  Pick a subset of the points, find an approximation, add more points, repeat until the approximation stabilizes.

rnlahaye:
Since vessels are user controllable, their trajectories can have weird kinks in them. I don’t see any way to detect these without checking every point (although of course these checks do not have to be performed at interpolation time; they can be performed when the points are initially added to the trajectory).

That is, I don’t think it is possible to do better than O(N) amortized.

eggrobin:
> although of course these checks do not have to be performed at interpolation time; they can be performed when the points are initially added to the trajectory

This ties into our ideas in https://github.com/mockingbirdnest/Principia/issues/2400#issuecomment-830861942 (the sections of trajectory that do not contain burns should not be serialized, and the burns need to be flagged in some fashion).

[...]

PR #4416, opened 2026-01-27
Documentation for downsampling

pleroy:
#2961.


documentation/Downsampling.tex
```diff
@@ -0,0 +1,218 @@
...
+\KwOut{A stream of intervals $\intclos{j_1}{j_2}$ such that the Hermite spline based on the tuples $\tuple{t_{j_1}, q_{j_1}, p_{j_1}}$ and $\tuple{t_{j_2}, q_{j_2}, p_{j_2}}$ approximates the input tuples $\tuple{t_i, q_i, p_i}$ for $i \in \intclos{j_1}{j_2}$ with a tolerance better that $\gh$.}
+\SetKwData{A}{A}
+\SetKwFunction{BinarySearchDownsampling}{BinarySearchDownsampling}
+\SetKwProg{Fn}{Function}{ is}{end}
+Let $A$ be an array of size $N$ used to store tuples.\;
+\While{not at end of input stream} {
+  \If{\A is full}{
+    Let $k \leftarrow \Floor{N/2}$.\;
+    Emit \SearchInterval{\A, $1$, $k$}.\;
+    Emit \SearchInterval{\A, $k + 1$, $N$}.\;
```
eggrobin: Why not `\SearchInterval{\A, $1$, $N$}`? This forces a subdivision even if a polynomial would fit the whole thing. Do we actually do that? If so, this deserves a comment that $N$ isn’t the maximum size of an interpolated interval, but double that.

documentation/Downsampling.tex
```diff
@@ -0,0 +1,218 @@
...
+}
+\caption{BinarySearchDownsampling.}
+\end{algorithm}
+
+This algorithm has the following properties:
+\begin{itemize}[nosep]
+\item The array $A$ must be stored permanently since it contains tuples that have not been downsampled yet and that will be needed to make a future decision about the intervals to emit.
+\item The choice of $N$ involves a trade-off: $N$ must be large enough that the downsampling is effective and produces long intervals $\intclos{j_1}{j_2}$; but it must not be so large that the size of $A$ affects performance.
+\item The algorithm has a complexity of $\BigO\of{N \log N}$ on average and $\BigO\of{N^2}$ in the worst case.
+\item The algorithm does not produce the longest possible intervals: in the worst case, all the emitted intervals may be too small by a factor of 2.
```
eggrobin, line 138: Where does this 2 come from? They can clearly be much worse than that. If the input stream is a cubic, any interval would work, and so a stream of length $kN$ will result in intervals that are too small by a factor of $2k$. ($k$ in the latest revision.)

documentation/Downsampling.tex
pleroy: Done.

Note that in practice the algorithm doesn't clear A, it retains the tail part that is below the tolerance.  But I didn't want to make the algorithm unnecessarily convoluted, since this is only for descriptive purposes.

documentation/Downsampling.tex
eggrobin: Ah, in the regime where the ideal polynomial spans less than $N$ points, the emitted interval may be too small by a factor of 2.

But there is also the case where the ideal polynomial spans more, and there there is no bound.

pleroy: The factor 2 comes from the splits that can be a bit too zealous, obviously.

I don't think it's interesting to say "it could be infinitely worse than the optimal".  That's understood from the fact that you decide to look at a finite window.  I think it's more useful to point at the trade-off involved in choosing _N_.

Issue #871, opened 2016-02-11
Support for engines of next stage in manœuvre


eggrobin:
So we don't have to activate an engine to plan for it; this is a prerequisite for #870 (since there is no throttling of SRBs, activating means igniting).

Issue #870, opened 2016-02-11
Support for SRBs in manœuvres


eggrobin:
Those should lock the norm of the Δv.

zajc3w:
Sorry for duplicate, didn't search well enough.

JacobB094:
Besides that, can we have the option to manually specify the burn duration? This would enable more accurate insertion, especially with slower burning motors.

PR #4190, opened 2025-03-29
Statically link the UCRT

eggrobin:
And pin it at 10.0.22621 (Windows 11 22H2), which is before before 10.0.26100 (Windows 11 24H2).
(There is no Windows 11 SDK for 10.0.22631, Windows 11 23H2).

Fix #4188.

This also gets rid of the separate glog.dll, libprotobuf.dll, and serialization.dll, which had been introduced in #1686.

PR #4418, opened 2026-02-04
Allow manœuvres to be executed by next stages

deacfb:
This pull requests add a few options that allow the flight planner to plan engines of upper stages, and an option to lock the total Δv for SRBs. It's an attempt to solve #870 and #871 

## Changes
- Added a `Set execution stage` option, when enabled, the burn editor will read the data of engines in the specified stage either from MechJeb2 (done by reflections so no additional build dependency) or the stock delta-v app, and sets the thrust and isp.
- Added a `First manœuvre of the stage` option. When selected, it means dead mass is decoupled before this stage. The initial mass of this manœuvre will be overriden by the specified stage's start mass, instead of last burn's final mass.
- Added a `Lock total Δv` option. When enabled, the total deltav of the manœuvre will be fixed. The increase of one component will cause decrease of other 2 components.
- Fixed wrong SDK version in `principia.props` The readme says we should use version `10.0.26100.0` but the version in the file is `10.0.22621.0`

## Known issues
- The new options are not serialized yet, which means they will get lost after scene changes (after all they are UI-only options so it might not be a big deal)
- Flight plan optimizer not working well when total Δv is locked in the editor (This is expected because I haven't touched the optimizer. It might be a little complex since we are moving from 4d free space to a 3d manifold)
- New strings are not localized yet.

## Demonstration
Below is a video showing the planning of a Mun free-return orbit from Kerbin using only SRB (Sorry about the poor quality because of the 10MB size limit).

https://github.com/user-attachments/assets/46e52fd1-b3f5-402a-8b80-55f16ad5cc35

### DIFF
--- a/principia.props
+++ b/principia.props
@@ -2,7 +2,7 @@
 <Project ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup Label="Globals">
     <Keyword>Win32Proj</Keyword>
-    <WindowsTargetPlatformVersion>10.0.26100.0</WindowsTargetPlatformVersion>
+    <WindowsTargetPlatformVersion>10.0.22621.0</WindowsTargetPlatformVersion>
   </PropertyGroup>
   <!--Define project configurations.  Nothing interesting can happen here.-->
   <ItemGroup Label="ProjectConfigurations">


0xd1cf001e:
Lets not change the version of sdk. Using a newer version causes some tests to fail. The test might need to be fixed.

eggrobin:
> The test might need to be fixed.

No it might not. This is set to an old version on purpose, to preserve compatibility with old saves, see https://github.com/mockingbirdnest/Principia/pull/4190. We care about the preservation of about exact numerical results in some parts of this code, that is why we have spent months working on correctly-rounded elementary functions.

> The new options are not serialized yet, which means they will get lost after scene changes (after all they are UI-only options so it might not be a big deal)

It is very much a big deal, and doing this right will require rewriting it all.

> Flight plan optimizer not working well when total Δv is locked in the editor (This is expected because I haven't touched the optimizer. It might be a little complex since we are moving from 4d free space to a 3d manifold)

Well yes, and you should be working with a chart of that manifold, probably all the way to the UI.

And you should certainly not be projecting things onto that manifold in the C#; we don’t do maths or physics in the C# for good reason.

More broadly, I’m afraid this PR is going in many many wrong directions (depending on MechJeb is also off-limits for many reasons), and neither @pleroy nor I have the energy to steer it back on track. Please talk to us, e.g., on Discord, before starting to work on something complicated like this.

0xd1cf001e:
@eggrobin I messaged you on Discord.

PR #4409, opened 2026-01-16
Чебышёв-Picard Iteration

rnlahaye:
This change adds an `Integrator` subclass which implements Чебышёв-Picard iteration.

Чебышёв-Picard iteration (a combination of Чебышёв approximation with Picard iteration) was first proposed by [Clenshaw and Norton in 1963](https://academic.oup.com/comjnl/article-abstract/6/1/88/400520). It has since been studied by [Shaver](https://dspace.mit.edu/handle/1721.1/130994), [Feagin and Narcozy](https://articles.adsabs.harvard.edu/cgi-bin/nph-iarticle_query?1983CeMec..29..107F&defaultprint=YES&filetype=.pdf), [Fukushima](https://articles.adsabs.harvard.edu/cgi-bin/nph-iarticle_query?1997AJ....113.1909F&defaultprint=YES&filetype=.pdf), and recently by [Junkins et al](https://apps.dtic.mil/sti/pdfs/ADA626629.pdf) (see also [this NASA report](https://ntrs.nasa.gov/api/citations/19690027545/downloads/19690027545.pdf) from 1969 which provides an implementation in FORTRAN).

Given a first order ordinary differential equation $x'=f(x,t)$ with initial condition $x(t_0)=x_0$, it can be used to solve the ODE on an interval $[t_0, t_f]$:
* Start with some initial approximation $x^0$ (typically the constant $x^0(t) = x_0$ is used).
* For each $x^i$:
  * Compute a Чебышёв approximation $g$ to $f(x^i, t)$ over $[t_0, t_f]$ (this requires evaluating $f(x^i, t)$ at Чебышёв nodes). 
  * Compute $x^{i+1}(t) = x_0 + \int_{t_0}^tg(s)\ ds$. Note that this integral can be evaluated analytically and yields another Чебышёв series.
* For sufficiently small intervals (and well-behaved $f$), the sequence $x^i$ converges uniformly.

See the linked papers for more details.

There is [evidence](https://media.proquest.com/media/hms/PFT/1/Y3QNG?_s=HDI7rBPc8F0o6EG3Wmbxt0xesYs%3D) that, together with some of the optimizations below, it has better performance than… some other integrators which Principia does not use.

### Regarding the implementation

The linear algebra code uses the formulation (and notation) from [Macomber's thesis](https://oaktrust.library.tamu.edu/server/api/core/bitstreams/e94bc2a8-7916-4d02-9aef-e7493a07ccb9/content).

[...]

PR #4419, opened 2026-02-06
Don’t escape filenames when listing files changed by include_what_you_using

eggrobin:
Spotted in #4409.


pleroy:
This fails IWYU (which is ironic, I guess).

eggrobin:
Ah yes. The new code only works when there is an IWYU failure.

PR #4425, opened 2026-02-07
Disambiguate the operator+s on PolynomialInMonomialBasis

eggrobin:
It seems subsumption doesn’t work here.


numerics/polynomial_in_monomial_basis.hpp
```diff
@@ -235,11 +235,13 @@ class PolynomialInMonomialBasis : public Polynomial<Value_, Argument_> {
                    PolynomialInMonomialBasis<V, A, r, E> const& right);
   template<affine V, affine A, int l, int r,
            template<typename, typename, int> typename E>
+  requires (!additive_group<V>)
```
pleroy: Indent += 2.  Also NOLINT all the things.

PR #4428, opened 2026-02-09
Allocate matrices on the heap in ЧебышёвPicardIntegrator

pleroy:
Major surgery to the fixed arrays to make them work in this context.

Also a few cleanups:
1. Regenerate the PDF for the bibliography.
2. Add the new files to the MSVC projects.
3. Don't run the test in debug mode.
4. East const.

#4440.

### DIFF
--- a/numerics/fixed_arrays.hpp
+++ b/numerics/fixed_arrays.hpp
@@ -44,8 +44,8 @@ class FixedVector final {
 
   template<bool uh>
   constexpr FixedVector(FixedVector<Scalar, size_, uh> const& other);
-  // A move can only be optimized if the source and destination have the same
-  // heap usage.
+  // A move is only optimized if the source and destination have the same heap
+  // usage.
   constexpr FixedVector(FixedVector&& other);
   // https://cplusplus.com/forum/beginner/269900/
   constexpr explicit FixedVector(FixedVector const volatile&) = delete;


eggrobin (approved):

numerics/fixed_arrays_body.hpp
```diff
@@ -55,8 +55,12 @@ struct Construction<std::unique_ptr<std::array<Scalar, size>>> {
 
   static constexpr std::unique_ptr<std::array<Scalar, size>>
   MakeValueInitialized() {
-    std::array<Scalar, size> zero{};
-    return make_not_null_unique<std::array<Scalar, size>>(std::move(zero));
+    auto a = make_not_null_unique<std::array<Scalar, size>>();
+    // Avoid creating one big array on the stack.
+    for (auto& element : *a) {
+      element = Scalar{};
+    }
```
eggrobin: Isn’t this redundant with the initialization performed by operator new?

pleroy: The default constructor of `std::array` does not zero its elements.  From [cppreference](https://en.cppreference.com/w/cpp/container/array.html): "(note that default initialization may result in indeterminate values for non-class T)".

eggrobin: Yes. But we do not [_default-initialize_](https://eel.is/c++draft/dcl.init#general-7) it.

This is a wrapper around a [_new-expression_](https://eel.is/c++draft/expr.new#24) with a _new-initializer_, so it is [_direct-initialization_](https://eel.is/c++draft/dcl.init#general-15), same as we had for `zero{}` before.

pleroy: TIL.  Fixed.

numerics/fixed_arrays.hpp
```diff
@@ -44,8 +44,11 @@ class FixedVector final {
 
   template<bool uh>
   constexpr FixedVector(FixedVector<Scalar, size_, uh> const& other);
-  template<bool uh>
-  constexpr FixedVector(FixedVector<Scalar, size_, uh>&& other);
+  // A move can only be optimized if the source and destination have the same
+  // heap usage.
+  constexpr FixedVector(FixedVector&& other);
```
eggrobin: For the kinds of things we put in a FixedVector, this is only going to optimize anything in the heap-to-heap case. If we had FixedVectors of things with costly copy constructors, it might make sense to do an element-by-element move instead of copy.

Let’s not go there for now, but maybe change « can only be » to « is only ».

Issue #4424, opened 2026-02-07
Investigate having a FixedMatrix with different types for different columns


pleroy:
This came up in the review of #4409, see [this comment](https://github.com/mockingbirdnest/Principia/pull/4409#discussion_r2771615643).  It might make it possible to avoid the homogeneization/dehomogeneization that occurs in various places.

Note that we have some algebra on tuples in `geometry/cartesian_product.hpp`, we might be able to use some of that.

pleroy:
We won't do this, we will rely on `DirectSum` instead.

PR #4429, opened 2026-02-09
Revert to UnboundMatrix from FixedMatrix in ЧебышёвPicardIntegrator.

rnlahaye:
UnboundMatrix is heap-allocated whereas FixedMatrix is stack-allocated.

One of the tests has $M = 500$ which means there are several $501\times501$ matrices in the constructor, each of which is ~2MB. It turns out having lots of (or possibly any) 2MB variables leads to stack overflows.

Ideally FixedMatrix would be smart-enough to heap-allocate when sufficiently large, but that is currently not the case.

#4440.

### DIFF
--- a/integrators/чебышёв_picard_integrator_body.hpp
+++ b/integrators/чебышёв_picard_integrator_body.hpp
@@ -1,5 +1,7 @@
 #pragma once
 
+#include "integrators/чебышёв_picard_integrator.hpp"
+
 #include <algorithm>
 #include <limits>
 #include <memory>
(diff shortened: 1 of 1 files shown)


rnlahaye:
It looks like I need UnboundedMatrix anyway in order to handle the second order ODEs.

eggrobin:
> It looks like I need UnboundedMatrix anyway in order to handle the second order ODEs.

Right, but those have homogeneous dependent variables, whereas the first order ones allow a tuple of different types. So once we get to #4424, they need to be handled differently anyway.

pleroy:
I don't like that we are bastardizing the data model by reverting to `UnboundedMatrix`: this code really works naturally with `FixedMatrix`.

We'll think of the 2nd order when we get there.

I'll look into adding a template parameter `bool use_heap = false` to `FixedMatrix` and see where it gets me.

pleroy:
Now that #4430 is in, I'll try to fix the new integrator to allocate the `FixedMatrix` objects on the heap.  Stay tuned.

rnlahaye:
Addressed by #4428 instead.

PR #4430, opened 2026-02-10
Support for heap allocation of fixed arrays

pleroy:
The choice of heap vs. stack allocation is done using a template parameter `use_heap`.  I don't expect any performance change when `use_heap` is defaulted (but we should look at the benchmarks).

All the operations that take several arrays work properly even if the arrays have different allocation options.  For the operations that build an array, the result is by default on the heap if all the arguments that are arrays are on the heap.  This can be controlled explicitly using a template parameter.

This PR is mostly big and boring and was largely produced using Copilot with GPT-5.  Any remaining error is the fault of Sam Altman, not mine.

#4440.

### DIFF
--- a/numerics/fixed_arrays_body.hpp
+++ b/numerics/fixed_arrays_body.hpp
@@ -415,7 +415,7 @@ constexpr FixedMatrix<Scalar_, rows_, columns_, use_heap>&
 FixedMatrix<Scalar_, rows_, columns_, use_heap>::operator*=(
     FixedMatrix<double, rows_, columns_, uh> const& right)
   requires(rows_ == columns_) {
-  // TODO(egg): We don't need to copy the whole matrix.
+  // TODO(egg): We don’t need to copy the whole matrix.
   return *this = *this * right;
 }
 


numerics/fixed_arrays.hpp
```diff
@@ -330,175 +456,283 @@ class FixedUpperTriangularMatrix final {
...
+    FixedStrictlyUpperTriangularMatrix<Scalar, columns, ruh> const& right);
+
+template<typename Scalar, std::int64_t columns,
+         bool luh, bool ruh>
+bool operator==(FixedUpperTriangularMatrix<Scalar, columns, luh> const& left,
+                FixedUpperTriangularMatrix<Scalar, columns, ruh> const& right);
+
+template<typename Scalar, std::int64_t size,
+         bool luh, bool ruh>
+bool operator!=(FixedVector<Scalar, size, luh> const& left,
```
eggrobin: Can we rely on C++20 rewritten candidates instead of defining all these `operator!=`s?

See https://en.cppreference.com/w/cpp/language/overload_resolution.html#Call_to_an_overloaded_operator, https://eel.is/c++draft/over.match#oper-3.4.3.

numerics/fixed_arrays_body.hpp
```diff
@@ -197,88 +313,123 @@ constexpr FixedMatrix<Scalar_, rows_, columns_>::FixedMatrix(
...
-    FixedMatrix<double, rows_, columns_> const& right)
+template<typename Scalar_, std::int64_t rows_, std::int64_t columns_,
+         bool use_heap>
+template<bool uh>
+constexpr FixedMatrix<Scalar_, rows_, columns_, use_heap>&
+FixedMatrix<Scalar_, rows_, columns_, use_heap>::operator*=(
+    FixedMatrix<double, rows_, columns_, uh> const& right)
   requires(rows_ == columns_) {
-  // TODO(egg): We don’t need to copy the whole matrix.
+  // TODO(egg): We don't need to copy the whole matrix.
```
eggrobin:
```suggestion
  // TODO(egg): We don’t need to copy the whole matrix.
```
Did the monkey do that?

pleroy: Yes, bad monkey.

PR #4435, opened 2026-02-12
Make DirectSum wrapper struct and use it in ЧебышёвPicardIntegrator.

rnlahaye:
This is one possible solution to #4424. I don't know if it's the best solution or even a desired solution, but it does at least work and the code seems to be cleaner.

The main ingredient here is a `VectorTuple` struct which wraps a tuple and has the various operations defined on it. For some reason the compiler likes this better than using a raw tuple and `using geometry::_cartesian_product::vector_space`.

In order to make this work,
* I made `FixedVector::Norm` (and `Norm²`) conditional (because a tuple cannot be squared)
* I changed the implementation of matrix-vector multiplication (the original implementation had a compile-time overflow for the large M cases).

Let me know if you think this is worth pursuing further.

#4440

### DIFF
--- a/integrators/чебышёв_picard_integrator_test.cpp
+++ b/integrators/чебышёв_picard_integrator_test.cpp
@@ -61,10 +61,7 @@ Displacement<World> displacement() {
 
 Length LInfinityNorm(std::tuple<Displacement<World>> const& v) {
   auto const& coordinates = std::get<0>(v).coordinates();
-
-  // Note that std::max(x, y, z) does not work for some reason.
-  return std::max(std::max(Abs(coordinates.x), Abs(coordinates.y)),
-                  Abs(coordinates.z));
+  return std::max({Abs(coordinates.x), Abs(coordinates.y), Abs(coordinates.z)});
 }
 
 // An initial value problem with a known solution.


pleroy:
> For some reason the compiler likes this better than using a raw tuple and using geometry::_cartesian_product::vector_space.

The reason why it doesn't work with `std::tuple` and it works with a wrapper is that our alias templates `Product`, `Quotient`, etc. in `base/algebra.hpp` look for `operator*`, `operator/`, etc. using ADL.  But the operators in `cartesian_product.hpp` that operate on `std::tuple` are *not* defined in the namespace `std` (that would be UB) so they are not found and everything falls apart.  On the other hand, with a wrapper that defines operators in the same namespace, everything is cool.

@eggrobin and I discussed the possibility of doing something like this, but some thought is required to decide what is the right abstraction.  In particular, a vector space structure is not sufficient if we want to use this for polynomials.

So I am somewhat on the fence when it comes to this CL.  On the one hand it moves in the right direction, on the other hand I suspect that we'd need to rewrite it fairly extensively once we've made up our mind.

(I wonder if we could define `tuple` in the dreaded `principia::std` namespace.  That might work, but I'm not sure if it's a foundation on which we want to build all the rest of Principia.)

numerics/fixed_arrays_body.hpp
```diff
@@ -1381,15 +1385,13 @@ template<typename LScalar, typename RScalar,
...
     FixedVector<RScalar, columns, ruh> const& right) {
-  std::array<Product<LScalar, RScalar>, rows> result;
-  auto const* row = left.data().data();
+  FixedVector<Product<LScalar, RScalar>, rows, uh> result;
   for (std::int64_t i = 0; i < rows; ++i) {
-    result[i] =
-        DotProduct<LScalar, RScalar, std::make_index_sequence<columns>>::
-            Compute(row, right.data());
-    row += columns;
+    for (std::int64_t j = 0; j < columns; ++j) {
```
pleroy: Hmm, I'm not happy with this.  The reason why it was done with `DotProduct` was to force unrolling of the loop for small arrays, and I am pretty sure that we depend on this for performance in some critical places.  Compilers may be smart enough to do this automatically, but then they might not, so we better guide them.

I think it would be sensible to have an `if constexpr (columns > ...)`, keep the current code for small arrays, and use a loop for large ones.

rnlahaye: Added an `if constexpr`.

eggrobin:
> @eggrobin and I discussed the possibility of doing something like this, but some thought is required to decide what is the right abstraction. In particular, a vector space structure is not sufficient if we want to use this for polynomials.
>
> So I am somewhat on the fence when it comes to this CL. On the one hand it moves in the right direction, on the other hand I suspect that we'd need to rewrite it fairly extensively once we've made up our mind.

We discussed this a bit more, and I think we have made up our mind.

The proper abstraction behind the operations in polynomial_ring and pointwise_inner_product is, well, PolynomialInMonomialBasis; so these are probably a distraction.

But what we want isn’t limited to vector spaces; in fact, if you try to use your integrator in integrator_plots.cpp, you will immediately put something in the tuple that is not a vector, to wit `Position<World>`, and that therefore makes the tuple a non-vector either.

Despite the file name, we probably don’t want the full generality of a cartesian product of whatever algebraic structures we start with; we neither need nor want to turn two rings into a direct product of rings with `operator*` being the componentwise product.

We think the right abstraction is thus a class `DirectSum`, which takes structures anywhere between `affine` and `vector_space` and has as much structure as the least powerful of all of its components (_i.e._, a `DirectSum` of affine R-modules should be an affine R-module, a `DirectSum` of additive groups should be an additive group, etc.). This should take a pack (so it should be `DirectSum<T, U, V>`, not `DirectSum<std::tuple<T, U, V>>`), and it should work in structured bindings (so it needs `std::tuple_size` and `std::tuple_element`, see [[dcl.struct.bind] §7](https://eel.is/c++draft/dcl.struct.bind#7) and https://en.cppreference.com/w/cpp/language/structured_binding.html#Case_2:_binding_a_type_implementing_the_tuple_operations, cf. [DiscreteTrajectory::value_type](https://github.com/mockingbirdnest/Principia/blob/326d65098f7a81921379768e921ff32cb312433c/physics/discrete_trajectory_types.hpp#L101-L114)).

Let `S=DirectSum<Args>`. Then `hilbert<S, S>` should hold (so there should be an `InnerProduct` on S, and while we are at it also a `Norm²` and a `Norm`) if `hilbert<T, T>` for all `T` in `Args` and all `Hilbert<T, T>::InnerProductType` are the same.

We thought about whether for `S1=DirectSum<Args1>`, `S2=DirectSum<Args2>` where `Hilbert<T1, T2>::InnerProductType` are the same for all corresponding `T1` and `T2` in `Args1` and `Args2`, we should have `hilbert<S1, S2>`, but that felt a bit like linear form territory and we didn’t want to have to think about what it means to transpose a `DirectSum`, so let’s hold off for now. If it comes up naturally we can add it.

Eventually we would want it to replace our existing usage of operators on std::tuple, which seems to be limited to `PolynomialInMonomialBasis`. `DirectSum` seems like the right abstraction there (a direct sum of monomials), and the additional structure currently in polynomial_ring and pointwise_inner_product would probably move to PolynomialInMonomialBasis. But let’s not touch the polynomials in this PR, that way madness lies.

rnlahaye:
`DirectSum` sounds good to me; I will make the changes.

pleroy:
While shaving this morning I thought that we could probably replace our `Pair` (which @eggrobin never liked) with `DirectSum` at some point in the future (not in this PR of course).

geometry/cartesian_product.hpp
```diff
@@ -64,14 +69,124 @@ namespace pointwise_inner_product {
...
+using namespace principia::base::_algebra;
+using namespace principia::quantities::_tuples;
+
+// The direct sum of a pack of affine types.
+//
+// Has as much structure as the least powerful of its components (a `DirectSum`
+// of affine R-modules is an affine R-module, a `DirectSum` of additive groups
+// is an additive group, etc.).
+template<affine... T>
+struct DirectSum {
```
pleroy: Let's move it to its own file, `geometry/direct_sum.hpp` etc. and make it a class.  I don't think that clients should access the `tuple` directly.

geometry/cartesian_product.hpp
```diff
@@ -64,14 +69,124 @@ namespace pointwise_inner_product {
...
+  constexpr DirectSum() = default;
+
+  // Constructor from elements.
+  constexpr DirectSum(T&&... args);  // NOLINT(runtime/explicit)
+
+  // Constructor from tuple.
+  constexpr explicit DirectSum(std::tuple<T...>&& tuple);
```
pleroy: Do we actually need this constructor?

geometry/cartesian_product.hpp
```diff
@@ -64,14 +69,124 @@ namespace pointwise_inner_product {
...
+
+  template<std::size_t I>
+  constexpr auto const& get() const;
```
pleroy: Can we "deduce this" instead of having multiple `get`?  See [this example](https://github.com/mockingbirdnest/Principia/blob/326d65098f7a81921379768e921ff32cb312433c/physics/discrete_trajectory_types.hpp#L64).

geometry/cartesian_product.hpp
```diff
@@ -64,14 +69,124 @@ namespace pointwise_inner_product {
...
+  template<std::size_t I>
+  constexpr auto& get();
+
+  constexpr auto Norm() const
+    requires hilbert<DirectSum<T...>, DirectSum<T...>>;
+  constexpr auto Norm²() const
+    requires hilbert<DirectSum<T...>, DirectSum<T...>>;
+
+  friend auto operator<=>(DirectSum const& left,
```
pleroy: Note that we don't have ordering for affine space.  I am wondering what this declaration will do in that case.  Maybe worth a test.

geometry/cartesian_product.hpp
```diff
@@ -64,14 +69,124 @@ namespace pointwise_inner_product {
...
+                          DirectSum const& right) = default;
+
+  template<affine... U>
+  DirectSum<T...>& operator+=(DirectSum<U...> const& right);
```
pleroy: This should be something like:
```
DirectSum<T...>& operator+=(Difference<DirectSum<U...>> const& right);
```
because if one of the `U` is affine the addition makes no sense.  This tell us btw that `Difference` should work for `DirectSum`, but I have no idea how hard/easy it will be.  In `pair.hpp`, which is the closest example I can think of, we have the dedicated trait `vector_of_t`, but that may go back to ancient history.

geometry/cartesian_product.hpp
```diff
@@ -64,14 +69,124 @@ namespace pointwise_inner_product {
...
+  template<affine... U>
+  DirectSum<T...>& operator-=(DirectSum<U...> const& right);
```
pleroy: Same comment.

geometry/cartesian_product.hpp
```diff
@@ -64,14 +69,124 @@ namespace pointwise_inner_product {
...
+
+  template<typename Scalar>
+  DirectSum<T...>& operator*=(Scalar const& right);
+  template<typename Scalar>
+  DirectSum<T...>& operator/=(Scalar const& right);
+
+  std::tuple<T...> tuple;
```
pleroy: Let's make this a private member.

geometry/cartesian_product.hpp
```diff
@@ -64,14 +69,124 @@ namespace pointwise_inner_product {
...
+};
+
+template<affine... T>
```
pleroy: This should be `additive_group` I think.  We don't have unary `+` for affine (it would make some kind of sense but who uses unary `+` anyway?) and we certainly don't have unary `-` (it makes no sense).

geometry/cartesian_product.hpp
```diff
@@ -64,14 +69,124 @@ namespace pointwise_inner_product {
...
+constexpr auto operator+(DirectSum<T...> const& right);
+
+template<affine... T>
+constexpr auto operator-(DirectSum<T...> const& right);
+
+template<affine... L, affine... R>
+constexpr auto operator+(DirectSum<L...> const& left,
+                         DirectSum<R...> const& right);
```
pleroy: It's more complicated than that.  We need three operators:
```
template<additive_group... L, additive_group... R>
constexpr auto operator+(DirectSum<L...> const& left,
                         DirectSum<R...> const& right);
template<additive_group... L, affine... R>
constexpr auto operator+(DirectSum<L...> const& left,
                         Difference<DirectSum<R...>> const& right);
template<affine... L, additive_group... R>
constexpr auto operator+(Difference<DirectSum<L...>> const& left,
                         DirectSum<R...> const& right);
```
Look at [polynomials](https://github.com/mockingbirdnest/Principia/blob/326d65098f7a81921379768e921ff32cb312433c/numerics/polynomial_in_monomial_basis.hpp#L352-L379) for inspiration.

geometry/cartesian_product.hpp
```diff
@@ -64,14 +69,124 @@ namespace pointwise_inner_product {
...
+
+template<affine... L, affine... R>
+constexpr auto operator-(DirectSum<L...> const& left,
+                         DirectSum<R...> const& right);
+
+template<typename L, affine... R>
```
pleroy: `vector_space... R`

Same below.

geometry/cartesian_product_body.hpp
```diff
@@ -399,6 +401,128 @@ constexpr auto CartesianProductPointwiseMultiplicativeSpace<
...
+}
+
+template<affine... T>
+constexpr auto operator+(DirectSum<T...> const& right) {
+  return right;
+}
+
+template<affine... T>
+constexpr auto operator-(DirectSum<T...> const& right) {
+  std::tuple<> zero;
```
pleroy: I *think* that this should be:
```
std::tuple<> zero{};
```
to force zero-initialization (e.g., if one of the component types is `double`).

geometry/cartesian_product_test.cpp
```diff
@@ -0,0 +1,82 @@
...
+  static_assert(
+      homogeneous_module<DirectSum<std::chrono::seconds, double>, int>);
+
+  static_assert(real_affine_space<DirectSum<Point<Length>>>);
+  static_assert(real_affine_space<DirectSum<Point<Length>, Speed>>);
+
+  static_assert(real_vector_space<DirectSum<double, double>>);
+  static_assert(real_vector_space<DirectSum<Length, Mass, Time>>);
+
+  using R² = DirectSum<double, double>;
```
pleroy: I think we would write `ℝ²` 🙂

geometry/cartesian_product_test.cpp
```diff
@@ -0,0 +1,82 @@
+#include "geometry/cartesian_product.hpp"
```
pleroy: I'd suggest to test something similar to `DegreesOfFreedom`, i.e., `DirectSum<Position<...>, Velocity<...>>` and check that all the operators that we need exist *and no others* (you should not be able to add two of these things, it's important to catch errors early when doing complicated physics).

geometry/cartesian_product.hpp
```diff
@@ -64,14 +69,124 @@ namespace pointwise_inner_product {
...
+using namespace principia::quantities::_tuples;
+
+// The direct sum of a pack of affine types.
+//
+// Has as much structure as the least powerful of its components (a `DirectSum`
+// of affine R-modules is an affine R-module, a `DirectSum` of additive groups
+// is an additive group, etc.).
+template<affine... T>
+struct DirectSum {
+  constexpr DirectSum() = default;
```
pleroy: Would it make sense to have a constructor taking an `uninitialized_t`?  We use that a lot for fixed arrays.

pleroy (commented):
I only looked at DirectSum.  Let's get that one in good shape first.

geometry/cartesian_product.hpp
rnlahaye: Moved to a separate file.

geometry/cartesian_product.hpp
rnlahaye: Done.

geometry/cartesian_product.hpp
rnlahaye: I am using this to construct `DirectSum`s from `ODE::DependentVariables` and such in `ЧебышёвPicardIntegrator`.

Of course, an alternate solution would be to use `DirectSum` in the various ODEs… but that would touch a bunch of things and probably shouldn't go in this PR.

I also tried `std::apply` and couldn't get it to work.

geometry/cartesian_product.hpp
rnlahaye: Done… until I made `get` a free function (to work with `for_all_of`) and I couldn't figure out how to deduce this in that case (is it even possible?).

geometry/cartesian_product.hpp
rnlahaye: Added a test, which failed, so I removed this and replaced it with `==` and `!=`.

geometry/cartesian_product.hpp
rnlahaye: It turns out that `DirectSum<Difference<T...>>` works fine. Done.

geometry/cartesian_product.hpp
rnlahaye: Done.

geometry/cartesian_product.hpp
rnlahaye: I still need a way to get the tuple somehow in order to convert it back to an ODE::DependentVariables.

geometry/cartesian_product.hpp
rnlahaye: Done.

geometry/cartesian_product.hpp
rnlahaye: I did something along these lines but with only two overloads—in the additive group case both overloads are identical so I just turn one of them off (also I did the same thing for subtraction). Is this sufficient?

geometry/cartesian_product.hpp
rnlahaye: Multiplication should only require module, right? Also we need the homogeneous versions in order to allow quantities.

Changed to homogeneous_module for multiplication and homogeneous_vector_space for division.

geometry/cartesian_product_body.hpp
rnlahaye: Done.

geometry/cartesian_product_test.cpp
rnlahaye: Done.

geometry/cartesian_product_test.cpp
rnlahaye: Done.

geometry/cartesian_product.hpp
pleroy: Sounds good.  If this class works well we might want to use it in more places instead of `std::tuple`, but not now.

geometry/cartesian_product.hpp
pleroy: I see, it cannot be a member function for `get` to work.  And you cannot deduce this when there is no this...  (Silly that you cannot extend `std::get`.)

geometry/direct_sum.hpp
```diff
@@ -0,0 +1,137 @@
...
+  constexpr DirectSum() = default;
+  constexpr explicit DirectSum(uninitialized_t);
+
+  // Constructor from elements.
+  constexpr DirectSum(T&&... args);  // NOLINT(runtime/explicit)
+
+  // Constructor from tuple.
+  constexpr explicit DirectSum(std::tuple<T...>&& tuple);
+
+  template<std::size_t i, affine... U>
```
pleroy: Do we need these friend declarations?  This is a `struct` and the field `tuple` is public, and you have `get` non-member functions anyway.

But then I really want this to be turned into a class, so you'll need the friend declarations (in the private part).

geometry/cartesian_product.hpp
pleroy: You only need `==`, `!=` is a *rewritten candidate* in C++20, see https://en.cppreference.com/w/cpp/language/overload_resolution.html.

geometry/cartesian_product.hpp
pleroy: Sure, but let's have a getter for that (and we can deduce this):
```
template<typename Self>
auto&& tuple(this Self&& self);
```

geometry/direct_sum.hpp
```diff
@@ -0,0 +1,137 @@
...
+template<additive_group... T>
+constexpr DirectSum<T...> operator-(DirectSum<T...> const& right);
+
+template<affine... T>
+constexpr DirectSum<T...> operator+(DirectSum<T...> const& left,
+                                    DirectSum<Difference<T>...> const& right);
+template<affine... T>
+constexpr DirectSum<T...> operator+(DirectSum<Difference<T>...> const& left,
+                                    DirectSum<T...> const& right)
+  requires(!std::is_same<DirectSum<T...>, DirectSum<Difference<T>...>>::value);
```
pleroy: This is technically correct, but I find it rather convoluted.  Basically this `requires` clause knows a lot about the difference between `additive_group` and `affine`.  I think we would prefer to stick to just using concepts without depending on their definition, but @eggrobin is the master of concepts, I'll leave it to him.

geometry/cartesian_product.hpp
pleroy: Yes you're right.

geometry/direct_sum_body.hpp
```diff
@@ -0,0 +1,182 @@
...
+template<affine... T>
+constexpr DirectSum<T...>::DirectSum(uninitialized_t) {}
+
+template<affine... T>
+constexpr DirectSum<T...>::DirectSum(T&&... args) : tuple(args...) {}
+
+template<affine... T>
+constexpr DirectSum<T...>::DirectSum(std::tuple<T...>&& tuple) : tuple(tuple) {}
+
+template<std::size_t i, affine... T>
```
pleroy: Let's put the definitions in the `_body.hpp` in the same order as the declarations in the `.hpp`, so move the `get` down.

geometry/direct_sum_body.hpp
```diff
@@ -0,0 +1,182 @@
...
+}
+template<std::size_t i, affine... T>
+constexpr auto& get(DirectSum<T...>& self) {
+  return std::get<i>(self.tuple);
+}
+
+template<affine... T>
+constexpr auto DirectSum<T...>::Norm() const
+  requires hilbert<DirectSum<T...>, DirectSum<T...>>
+{
```
pleroy, line 41: Is that the formatting produced by `clang-format`?  I would have expected the `{` at the end of the previous line, but I am not completely sure what it does.

rnlahaye: Yes.

geometry/direct_sum_body.hpp
rnlahaye: Done.

geometry/direct_sum.hpp
rnlahaye: Removed friend and moved out of the class body.

geometry/cartesian_product.hpp
rnlahaye: Done.

geometry/cartesian_product.hpp
rnlahaye: Done.

geometry/direct_sum.hpp
```diff
@@ -0,0 +1,141 @@
...
+using namespace principia::base::_tags;
+using namespace principia::quantities::_tuples;
+
+// The direct sum of a pack of affine types.
+//
+// Has as much structure as the least powerful of its components (a `DirectSum`
+// of affine R-modules is an affine R-module, a `DirectSum` of additive groups
+// is an additive group, etc.).
+template<affine... T>
+struct DirectSum {
```
pleroy: `class`

geometry/direct_sum_body.hpp
```diff
@@ -0,0 +1,190 @@
...
+namespace geometry {
+namespace _direct_sum {
+namespace internal {
+
+using namespace principia::base::_for_all_of;
+using namespace principia::geometry::_hilbert;
+using namespace principia::numerics::_elementary_functions;
+
+template<affine... T>
+constexpr DirectSum<T...>::DirectSum(uninitialized_t) {}
```
pleroy, line 23: Technically this should forward to the constructor of each `T` that takes an `uninitialized_t`, if there is one (think of a `DirectSum` with a component that's a `FixedMatrix`).  But that's probably a bit tricky.  Can you leave a `TODO(phl)` to look into this?

geometry/direct_sum_body.hpp
```diff
@@ -0,0 +1,190 @@
...
+  return zero - right;
+}
+
+template<affine... T>
+constexpr DirectSum<T...> operator+(DirectSum<T...> const& left,
+                                    DirectSum<Difference<T>...> const& right) {
+  DirectSum<T...> sum(uninitialized);
+  for_all_of(left, right, sum)
+      .loop([](auto const& left, auto const& right, auto& sum) {
+        sum = left + right;
```
pleroy, line 110: This code is obviously much simpler than the drudgery in `cartesian_product_body.hpp`.  Part of it is that `cartesian_product_body.hpp` is rather old and was probably written for C++11.  But part of it is that its performance is critical for polynomial operations, and that drudgery results in code that's fully inline and unrolled.

So I wonder what the performance of the calls to `for_all_of` will be.  We don't need to answer this question right now, but we will before we can use `DirectSum` in more places.  Inconveniently we don't have a benchmark for `cartesian_product`, we benchmark the polynomials, so a comparison is not easy.

geometry/direct_sum_test.cpp
```diff
@@ -0,0 +1,190 @@
...
+concept less_than_or_equal = requires(T a, T b) {
+  { a <= b };
+};
+template<typename T>
+concept greater_than_or_equal = requires(T a, T b) {
+  { a >= b };
+};
+
+TEST(DirectSumTest, Unordered) {
+  static_assert(!less_than<ℝ²>);
```
pleroy, line 74: Why only test one concept?  You declared 4 of them above.

numerics/fixed_arrays.hpp
```diff
@@ -74,8 +74,8 @@ class FixedVector final {
   constexpr FixedVector& operator*=(double right);
   constexpr FixedVector& operator/=(double right);
 
-  Scalar Norm() const;
-  Square<Scalar> Norm²() const;
+  auto Norm() const requires homogeneous_ring<Scalar>;
+  auto Norm²() const requires homogeneous_ring<Scalar>;
```
pleroy, line 78: "Ring" is too strong, we just need `Scalar` to be a Hilbert space.  This being said, our concepts around Hilbert spaces are probably not well organized, so I'm not sure what to advise.  @eggrobin for guidance.

integrators/чебышёв_picard_integrator_body.hpp
```diff
@@ -23,70 +26,46 @@ namespace internal {
...
+// Wrap a tuple in DoublePrecision.
+template<typename... T>
+std::tuple<DoublePrecision<T>...> WrapInDoublePrecision(
+    std::tuple<T...> const& in) {
+  std::tuple<DoublePrecision<T>...> out;
+  for_all_of(in, out).loop([](auto const& inᵢ, auto& outᵢ) { outᵢ = inᵢ; });
+  return out;
 }
 
 // Returns max|aᵢⱼ|.
```
pleroy: ... expressed in SI units.

Note that there is an implicit normalization factor here (e.g., 1 second ≈ 1 metre) and I am somewhat uncomfortable that it is hard-wired deep in the guts of the integrator here and cannot be injected as a parameter.

WDYT?

rnlahaye: Done.

numerics/fixed_arrays.hpp
rnlahaye: Yes, I don't think `double` is `hilbert` currently.

geometry/direct_sum_test.cpp
rnlahaye: Whoops! Fixed.

geometry/direct_sum_body.hpp
rnlahaye: It seems to do something vaguely reasonable on Godbolt at least: https://godbolt.org/z/7hab6dxTh.

geometry/direct_sum_body.hpp
rnlahaye: Done.

geometry/direct_sum.hpp
rnlahaye: Done.

integrators/чебышёв_picard_integrator_body.hpp
eggrobin: > Note that there is an implicit normalization factor here (e.g., 1 second ≈ 1 metre) and I am somewhat uncomfortable that it is hard-wired deep in the guts of the integrator here and cannot be injected as a parameter.
> 
> WDYT?

There is a similar problem with the adaptive stepsize methods, and the way we have dealt with that is that instead of passing a tolerance and computing an artificial norm of the error, we let the caller do the messy thing in a problem-specific way by passing a ToleranceToErrorRatio.

So here we should probably pass a similar function (although we do not need the ratio, only a boolean telling us whether the error is tolerable).

eggrobin: Note for instance that as currently written, the integrator still wouldn’t compile for an equation involving `Position` or `Displacement`, as neither si::Unit nor std::abs in this function would be applicable.

In particular, this means you can’t use it on the existing equations in the integrator plots.

numerics/fixed_arrays.hpp
eggrobin: `hilbert` and its friend `Hilbert` are a mess. I’ll deal with this later, I don’t think making these functions work for non-homogeneous_ring Scalar is needed to make progress on the new integrator.

geometry/direct_sum_body.hpp
pleroy: To be honest we mostly care about MSVC (the majority of our users are on Windows) but the generated code is similar, https://godbolt.org/z/jrndjjdh6.

geometry/direct_sum.hpp
eggrobin: I think this constraint could be reformulated as `requires (!(additive_group<T> && ...))`? (Or `requires (!additive_group<T> || ...)` depending on how we feel about De Morgan today.)

rnlahaye: Done.

integrators/чебышёв_picard_integrator_body.hpp
rnlahaye: Changed to use a function.

integrators/чебышёв_picard_integrator_test.cpp
eggrobin: Change this to use `Position`, with a test `Frame` as in
https://github.com/mockingbirdnest/Principia/blob/6b50fc81a01217425f9174d2c669c461194ea9b3/mathematica/integrator_plots.cpp#L166-L169

It is fine if the actual problems in the test stay one-dimensional along one of the axes, but this will check that we are no longer requiring `si::Unit` or a multiplicative structure or anything else that would get in the way of using it in the integrator plots and the ephemeris.

rnlahaye: Done.

integrators/чебышёв_picard_integrator_test.cpp
```diff
@@ -39,11 +42,31 @@ using namespace principia::testing_utilities::_almost_equals;
...
+//
+// Used to embed one-dimensional problems in the World.
+Displacement<World> displacement() {
+  return Displacement<World>({1 * Metre, 0 * Metre, 0 * Metre});
+}
+
+Length LInfinityNorm(std::tuple<Displacement<World>> const& v) {
+  auto const& coordinates = std::get<0>(v).coordinates();
+
+  // Note that std::max(x, y, z) does not work for some reason.
```
pleroy: I think it must be written `std::max({x, y, z})`.  Not very important anyway.

rnlahaye: Done.

numerics/fixed_arrays.hpp
eggrobin: See #4444.

PR #4444, opened 2026-02-18
Get rid of Hilbert, make hilbert usable

eggrobin:
Hilbert has long been a thorn in our side and a stress test for compilers; and despite that it only works for selected types (double, quantities, and multivectors).

Instead this adds free functions `InnerProduct`, `Norm`, and `Norm²`.

These works for any ordered field (dealing with the `double` problem, and the more general problem that requiring Quantity to have a dedicated `.Norm()` and `InnerProduct` overload seems tasteless) and for anything that provides `.Norm()`, `.Norm²()`, and `InnerProduct` (`InnerProduct` is found by ADL preferentially to the one in `_hilbert`, and the free `Norm` and `Norm²` call the members in the non-ordered field case).

### DIFF
--- a/numerics/fixed_arrays_test.cpp
+++ b/numerics/fixed_arrays_test.cpp
@@ -3,11 +3,11 @@
 #include <utility>
 
 #include "base/algebra.hpp"
+#include "geometry/complexification.hpp"
 #include "geometry/hilbert.hpp"
 #include "gmock/gmock.h"
 #include "gtest/gtest.h"
 #include "numerics/elementary_functions.hpp"
-#include "geometry/complexification.hpp"
 #include "numerics/transposed_view.hpp"
 #include "quantities/quantities.hpp"
 


base/algebra.hpp
```diff
@@ -238,6 +246,8 @@ using internal::Product;
 using internal::Quotient;
 using internal::Square;
 using internal::Sum;
+using internal::dimension;
+using internal::real_dimension;
```
pleroy: Alphabetize.

geometry/complexification_body.hpp
```diff
@@ -57,10 +57,16 @@ Complexification<Vector> Complexification<Vector>::Conjugate() const {
...
-  return Hilbert<Vector>::Norm²(real_part_) +
-         Hilbert<Vector>::Norm²(imaginary_part_);
+  using geometry::_hilbert::Norm²;
+  return Norm²(real_part_) + Norm²(imaginary_part_);
+}
+
+template<typename Vector>
+typename NormType<Vector> Complexification<Vector>::Norm() const {
+  using geometry::_hilbert::Norm;
+  return Norm(real_part_) + Norm(imaginary_part_);
```
pleroy: Wat?

geometry/complexification_body.hpp
```diff
@@ -181,18 +187,14 @@ Complexification<Quotient<LVector, RVector>> operator/(
...
-          Hilbert<LVector, RVector>::InnerProduct(left.real_part(),
-                                                  right.real_part()));
+Complexification<InnerProductType<LVector, RVector>> InnerProduct(
+    Complexification<LVector> const& left,
+    Complexification<RVector> const& right) {
+  return Complexification<InnerProductType<LVector, RVector>>(
+      InnerProduct(left.real_part(), right.real_part()) +
+          InnerProduct(left.imaginary_part(), right.imaginary_part()),
+      InnerProduct(left.imaginary_part(), right.real_part()) -
+          InnerProduct(left.real_part(), right.real_part()));
```
pleroy: I believe that this computation is not well-conditioned, but that's not a change...

geometry/complexification.hpp
```diff
@@ -42,7 +42,8 @@ class Complexification {
 
   Complexification Conjugate() const;
 
-  typename Hilbert<Vector>::Norm²Type Norm²() const;
+  typename Norm²Type<Vector> Norm²() const;
```
pleroy: I don't think you need `typename` here (2x).

geometry/grassmann.hpp
```diff
@@ -351,4 +351,10 @@ using internal::Wedge;
 }  // namespace geometry
 }  // namespace principia
 
+template<typename Scalar, typename Frame, int rank>
+constexpr int principia::base::_algebra::dimension<
+    principia::geometry::_grassmann::Multivector<Scalar, Frame, rank>,
+    double> = principia::geometry::_grassmann::
+    Multivector<Scalar, Frame, rank>::dimension;
```
pleroy: Would it make sense to put this in the `_grassmann` namespace (given that `dimension` is fully qualified)?

geometry/hilbert.hpp
```diff
@@ -1,187 +1,87 @@
...
-
-  using InnerProductType = Product<T1, T2>;
-  static InnerProductType InnerProduct(T1 const& t1, T2 const& t2);
+
+template<homogeneous_field T, homogeneous_field U>
+  requires std::totally_ordered<Product<T, U>>
+constexpr Product<T, U> InnerProduct(T const& left, U const& right);
+
+// We go through a generator instead of defining the return types of the two
+// overloads of Norm² as auto for the one that does .Norm²() and Square<T> for
```
pleroy: Backticks around the identifiers.

geometry/hilbert.hpp
```diff
@@ -1,187 +1,87 @@
...
-    _MSC_FULL_VER == 194'435'221 || \
-    _MSC_FULL_VER == 194'435'222
-  {  // NOLINT
-    return _grassmann::internal::InnerProduct(t1, t2);
-  }
-#else
-  ;  // NOLINT
-#endif
-};
+template<typename T, typename U = T>
```
pleroy: Move this up after `InnerProduct`, to keep the functions and types together.

geometry/hilbert.hpp
```diff
@@ -1,187 +1,87 @@
...
-  ;  // NOLINT
-#endif
-
-  using NormalizedType = Quotient<T, NormType>;
+using NormalizedType = Quotient<T, NormType<T>>;
+
+template<typename V>
+concept hilbert = requires(V u, V v) {
+  { Norm(u) } -> homogeneous_field;
+  { InnerProduct(u, v) } -> homogeneous_field;
```
pleroy, line 64: Move one nudge up, `InnerProduct` before `Norm`, as you did for the functions above.

geometry/r3_element.hpp
```diff
@@ -256,4 +256,8 @@ using internal::SphericalCoordinates;
 }  // namespace geometry
 }  // namespace principia
 
+template<typename Scalar>
+constexpr int principia::base::_algebra::
+    dimension<principia::geometry::_r3_element::R3Element<Scalar>, double> = 3;
```
pleroy: Same as above, preference for having it in the `_r3_element` namespace.

numerics/hermite3_body.hpp
```diff
@@ -206,9 +206,8 @@ auto Hermite3<Value_, Argument_>::LInfinityL₂Error(
         get_value) const -> NormType {
   NormType result{};
   for (const auto& sample : samples) {
-    result = std::max(result,
-                      Hilbert<Difference<Value>>::Norm(
-                          (*this)(get_argument(sample)) - get_value(sample)));
+    result =
+        std::max(result, Norm((*this)(get_argument(sample))-get_value(sample)));
```
pleroy: Fits on the previous line?  Also, spaces around `-`.  In other words, format this line.

numerics/nearest_neighbour.hpp
```diff
@@ -63,11 +63,11 @@ class PrincipalComponentPartitioningTree {
   using Displacement = Difference<Value>;
 
   // The type of the norm (and its square) of `Displacement`.
-  using Norm = typename Hilbert<Displacement>::NormType;
-  using Norm² = typename Hilbert<Displacement>::Norm²Type;
+  using Norm = NormType<Displacement>;
```
pleroy: It is perhaps sad that this is not named `NormType`.  Same for the next declaration.  Change that?

quantities/quantities.hpp
```diff
@@ -191,4 +193,8 @@ using internal::ToM256D;
 }  // namespace quantities
 }  // namespace principia
 
+template<typename D>
+constexpr int principia::base::_algebra::dimension<
+    principia::quantities::_quantities::Quantity<D>, double> = 1;
```
pleroy: In `_quantities`?

geometry/grassmann.hpp
eggrobin: It would not be legal.

geometry/r3_element.hpp
eggrobin: Same as above, not possible.

quantities/quantities.hpp
eggrobin: No (same as above).

numerics/hermite3_body.hpp
eggrobin: It looks like clang-format is supremely confused by this line, it removed the spaces…

geometry/complexification_body.hpp
eggrobin: The header has
```C++
// TODO(egg): Numerical analysis.
template<typename LVector, typename RVector>
Complexification<Product<LVector, RVector>> operator*(
    Complexification<LVector> const& left,
    Complexification<RVector> const& right);
```

So I think we are aware of that.

geometry/grassmann.hpp
pleroy: Ok but I have no idea what the rules are for where a partial specialization can be declared (I am reading 13.7.6.1(1, 6) and it's clear as mud).  In particular, our two compilers disagree as to whether it could be declared in `principia` or `principia::base`, https://godbolt.org/z/q6b1GE1j6.

I would prefer to explicitly reopen the namespace:
```
namespace principia {
namespace base {
namespace _algebra {

template<typename Scalar, typename Frame, int rank>
constexpr int dimension<
    geometry::_grassmann::Multivector<Scalar, Frame, rank>,
    double> = geometry::_grassmann::Multivector<Scalar, Frame, rank>::dimension;

}  // namespace _algebra
}  // namespace base
}  // namespace principia
```
Same comment in the other places.

geometry/hilbert_test.cpp
```diff
@@ -2,132 +2,63 @@
...
-TEST(HilbertTest, BadParameters) {
-#if 0
-  using H1 = Hilbert<void*, Length>;
-  using T1 = H1::InnerProductType;
-
-  using H2 = Hilbert<Bivector<double, World>, Vector<Length, World>>;
-  using T2 = H2::InnerProductType;
-#endif
-}
+static_assert(hilbert<double>);
```
pleroy: I'd prefer to wrap these assertions in a `TEST` just to have a name documenting what we are testing (see #4445 for an example).

base/algebra.hpp
```diff
@@ -204,19 +204,28 @@ concept homogeneous_affine_module = affine<A> && homogeneous_ring<R> &&
...
-concept hilbert = requires(T1 const& t1, T2 const& t2) {
-  InnerProduct(t1, t2);
-};
+template<typename V, ring K>
+  requires homogeneous_affine_module<V, K>
+int dimension;
+
+template<affine A, ring K>
+  requires(!additive_group<A>)
+int dimension<A, K> = dimension<Difference<A>, K>;
```
pleroy, line 213: Clang doesn't like this, https://godbolt.org/z/d698844rW.

quantities/quantities.hpp
```diff
@@ -191,10 +191,14 @@ using internal::ToM256D;
 
 }  // namespace _quantities
 }  // namespace quantities
-}  // namespace principia
+namespace base {
```
pleroy, line 195: Blank line before.

Issue #4440, opened 2026-02-16
Tracking issue for Чебышёв-Picard iteration


pleroy:
This work is starting to be sizeable, so creating a tracking issue to group all the changes that relate to it (we should probably have done this earlier).

PR #4445, opened 2026-02-19
Make {Fixed,Unbounded}{Vector,Matrix} satisfy affine.

rnlahaye:
This allows them to go in a `DirectSum` or similar.

#4440

### DIFF
--- a/numerics/fixed_arrays_test.cpp
+++ b/numerics/fixed_arrays_test.cpp
@@ -3,10 +3,10 @@
 #include <utility>
 
 #include "base/algebra.hpp"
-#include "geometry/frame.hpp"
-#include "geometry/space.hpp"
 #include "geometry/complexification.hpp"
+#include "geometry/frame.hpp"
 #include "geometry/hilbert.hpp"
+#include "geometry/space.hpp"
 #include "gmock/gmock.h"
 #include "gtest/gtest.h"
 #include "numerics/elementary_functions.hpp"
@@ -17,10 +17,10 @@ namespace principia {
 namespace numerics {
 
 using namespace principia::base::_algebra;
-using namespace principia::geometry::_frame;
-using namespace principia::geometry::_space;
 using namespace principia::geometry::_complexification;
+using namespace principia::geometry::_frame;
 using namespace principia::geometry::_hilbert;
+using namespace principia::geometry::_space;
 using namespace principia::numerics::_elementary_functions;
 using namespace principia::numerics::_fixed_arrays;
 using namespace principia::numerics::_transposed_view;


numerics/fixed_arrays.hpp
```diff
@@ -69,8 +69,12 @@ class FixedVector final {
 
   constexpr FixedVector& operator=(Scalar const (&right)[size_]);
 
-  constexpr FixedVector& operator+=(FixedVector const& right);
-  constexpr FixedVector& operator-=(FixedVector const& right);
+  template<bool uh>
+  constexpr FixedVector& operator+=(
+      FixedVector<Difference<Scalar>, size_, uh> const& right);
```
pleroy, line 76: The usage of `Difference` here makes me think that we should require `Scalar` to be `affine`.  I don't imagine that we would ever want a fixed array of something that's not at least affine.  (There are other places that should probably use concepts, e.g., `*=` wants `Scalar` to be a module, but let's not go there for now.)

Same comment for `FixedMatrix`.

numerics/unbounded_arrays.hpp
```diff
@@ -72,8 +72,8 @@ class UnboundedVector final {
 
   UnboundedVector& operator=(std::initializer_list<Scalar> right);
 
-  UnboundedVector& operator+=(UnboundedVector const& right);
-  UnboundedVector& operator-=(UnboundedVector const& right);
+  UnboundedVector& operator+=(UnboundedVector<Difference<Scalar>> const& right);
+  UnboundedVector& operator-=(UnboundedVector<Difference<Scalar>> const& right);
```
pleroy, line 78: Same comment, if we can change `Scalar` to be `affine` without too much hassle, let's do it.

numerics/unbounded_arrays.hpp
```diff
@@ -83,8 +83,8 @@ class UnboundedVector final {
 
   void EraseToEnd(std::int64_t begin_index);
 
-  Scalar Norm() const;
-  Square<Scalar> Norm²() const;
+  auto Norm() const requires homogeneous_ring<Scalar>;
+  auto Norm²() const requires homogeneous_ring<Scalar>;
```
pleroy: @eggrobin for a heads-up as he was changing this, too.

numerics/fixed_arrays.hpp
rnlahaye: Done.

numerics/unbounded_arrays.hpp
rnlahaye: Done.

numerics/unbounded_arrays.hpp
eggrobin: Yes. This is actually incorrect (consider `Scalar`=`Complexification<double>`), you want either `homogeneous_field<Scalar> && std::totally_ordered<Scalar>` if you keep the existing implementation, or `hilbert` using the operations from `_hilbert`.

rnlahaye: Changed to `hilbert`.

PR #4516, opened 2026-04-12
Fix the ordering of dependencies in rebuild_all_solutions

eggrobin:


eggrobin:
https://github.com/mockingbirdnest/protobuf/commit/65e9f6a70b408a9feb803f598060b20fb610436b instead.

PR #4519, opened 2026-04-14
Add .Times(AtLeast(1)) to log expectations for old version notices

eggrobin:
This fixes some of the issues found by the diurnal build (there is something strange happening in discrete_trajectory_test, where the same trajectory as in the Lpg test here evaluates differently; I am still trying to figure that out).

### DIFF
--- a/ksp_plugin_test/plugin_compatibility_test.cpp
+++ b/ksp_plugin_test/plugin_compatibility_test.cpp
@@ -123,18 +123,14 @@ TEST_F(PluginCompatibilityTest, PreCohen) {
   EXPECT_CALL(log,
               Log(absl::LogSeverity::kWarning,
                   _,
-                  HasSubstr("pre-Cohen ContinuousTrajectory"))).
-      Times(AtLeast(1));
+                  HasSubstr("pre-Cohen ContinuousTrajectory")))
+      .Times(AtLeast(1));
   // The save is even older.
   EXPECT_CALL(log,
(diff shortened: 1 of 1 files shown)


ksp_plugin_test/plugin_compatibility_test.cpp
```diff
@@ -123,18 +123,14 @@ TEST_F(PluginCompatibilityTest, PreCohen) {
   EXPECT_CALL(log,
               Log(absl::LogSeverity::kWarning,
                   _,
-                  HasSubstr("pre-Cohen ContinuousTrajectory"))).
-      Times(AtLeast(1));
```
pleroy, line 127: The expectations were carefully formatted by hand to all look the same.  This is gone now, because clang-format decides to use a different style for each and every expectation.  Could we leave it as it was?

eggrobin: I found the old style particularly hard to read, compared to the old expectations on the StringLogSink.
Most of this is noise, the important parts are the HasSubstr which is in the middle and the contents of the Times.

Is there a way to make this more compact? Maybe a shorter alias for kWarning? Or we could just put no matcher on the severity, it is not really relevant.
If we manage to make it more compact the autoformatting should be more consistent.

pleroy: Let's go with what you have.  Trying to shorten the expectations is generally a dubious idea since it makes them harder to understand.

PR #4521, opened 2026-04-17
Fix a dereference of an invalid iterator that was introduced in the spring 2026 migration

pleroy:
This is expected to fix the diurnal build.

### DIFF
--- a/physics/discrete_trajectory_segment_body.hpp
+++ b/physics/discrete_trajectory_segment_body.hpp
@@ -166,7 +166,6 @@ DiscreteTrajectorySegment<Frame>::EvaluateDegreesOfFreedom(
     return it->degrees_of_freedom;
   }
   auto const& interpolation = get_hermite3(it);
-  LOG(ERROR)<<"Hermite3: "<<interpolation;
   return {interpolation(t), interpolation.EvaluateDerivative(t)};
 }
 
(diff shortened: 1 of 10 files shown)


physics/discrete_trajectory_segment_body.hpp
```diff
@@ -586,7 +586,12 @@ void DiscreteTrajectorySegment<Frame>::UpdateInterpolation(
   CHECK(upper != timeline_.cbegin());
   auto const lower = std::prev(upper);
   auto const& [lower_time, lower_degrees_of_freedom] = *lower;
-  auto const& [upper_time, upper_degrees_of_freedom] = *upper;
+
+  // This needs to be copied because the element designated by `upper` is going
+  // to be extracted from the set, which invalidates `upper`.
+  auto const upper_time = upper->time;
+  auto const upper_degrees_of_freedom = upper->degrees_of_freedom;
```
eggrobin, line 593: Couldn’t this simply be
```
  auto const [upper_time, upper_degrees_of_freedom] = *upper;
```
(Copying into a value_type)?

pleroy: No because `*upper` contains another field (a `unique_ptr`) which is not accessible by structured bindings, but still causes the compiler to complain with "deleted assignment".

eggrobin (approved):

Issue #3697, opened 2023-06-19
Service Module and enclosed fairings will start to rotate faster and faster when free falling in atmosphere


pleroy:
This was reported by "Wilds" on the [forum](https://forum.kerbalspaceprogram.com/topic/162200-wip181-191-1101-1110%E2%80%932-1122%E2%80%935-principia%E2%80%94version-%E4%BC%8A%E8%97%A4-released-2023-06-18%E2%80%94n-body-and-extended-body-gravitation/?do=findComment&comment=4294038):

> Certain parts like SM-6A Service Module and enclosed fairings will start to rotate faster and faster when free falling in atmosphere. Eventually their rotating increases so much that the craft krakens; and the camera shakes violently.
>
> This behavior is only happening in atmospheres. Floating down to Venusian surface below 10m/s in a simple SM-6A service bay with 1.25m heat shield shield will see the craft start to rotate/spin without any reaction wheels. Once it starts to rotate, its rotation rate accelerates dramatically. Attaching any kind of wings or certain aerodynamic parts will counteract this spinning issue.
>
> Removing Principia will cause these parts to behave normally again.
>
> It certainly "feels" like a phantom force is being applied to one side of the craft.
>
> edit: The fix for me is to just not use those parts for re-entry capsules. A 1.25m service bay will never have this spinning issue, same for the pods that can contain Kerbals.

> I enabled journal, changed scenes, and launched a suborbital rocket with a 4 part capsule known for spinning. I reach around 550km apoapsis. At around 17mins into the flight, the capsule re-enters atmosphere. Spinning gets worse and worse, parachute is ripped off from high G spinning, and it krakens itself

Journal is [here](https://www.mediafire.com/file/rkwgtdinquz85f5/JOURNAL.7z/file).

This is reminiscent of the so-called "scimas problem" with #2519.

pleroy:
We were finally able to reproduce this problem in a very simple situation: take a single part, a `Mk1-3 Command Pod`, cheat it to an altitude of 100 km with a pitch of 90°, and let it fall.  When it starts entering the atmosphere the drag slowly orients it in the "right" direction (that is, with the bottom, where the heat shield should be, facing down).  As the atmosphere becomes denser the oscillations increase in frequency but remain of moderate amplitude.  However, around an altitude of 5 km, the oscillations start increasing in amplitude and exceed 90° when it touches the ground.

[...]

PR #4541, opened 2026-04-25
Change the computation of the drag torque to have the same effect as in PhysX

### LINKED ISSUE #4166: Craft Jitters at high speed and low altitude
Only using Restock, Restock +  and Principia, on a fresh sandbox save. 

When going really fast at low altitude (speed needed depends on altitude and plane design, using a simple rapier https://imgur.com/a/mZsxA8N it starts at around 800m/s bellow 2000m and 1000m/s at 5000m) the plane and camera shakes side to side, increasing in severity with higher speed until the ship gets separated and broken. The aircraft is still a bit controllable. Video: https://imgur.com/a/rKuZTdX (warning, sudden motion)

This also happens on reentry with a rocket.

Of course if i remove principia from GameData this stops happening.

### LINKED ISSUE #4531: Possible Principia issue during reentry / vessel split: NaN velocities and vessel instability
I’m not fully sure this is a Principia-only bug, but after a lot of testing it looks like Principia is on the critical path of the failure.

My setup is KSP 1.12.5 on Windows x64, using Principia 2026041712-Lehmer. The main repro is in an RSS-based modded install, and it happens even at x1 warp.

What happens is this: during reentry, usually around vessel split or parachute deployment, the craft suddenly becomes unstable, its velocity jumps to absurd values (sometimes millions of metres per second), it gets thrown away in an apparently random direction, and then KSP starts reporting NaN collider / transform errors, sometimes recalculates the orbit to the Sun, and then freezes or crashes.

The reason I’m reporting this here is that Principia logs consistently show the failure happening around unpack / split transitions: unpacked parts appearing between `BetterLateThanNever` and `WaitForFixedUpdate`, vessel split / extraction, and repeated internal collision reports. In some reduced repros, Principia also removes the vessel as `dead, vessel is completely unfaithful`.

I’m not claiming this is definitely a Principia bug, but removing Principia makes the problem disappear on my side, while keeping it installed brings it back.

UPD: The problem was analyzed and Codex tried to solve it. I don't understand code myself, but I admit that the neural network can talk complete nonsense, but the visual description of the problem is correct.

Image

[KSP.log](https://github.com/user-attachments/files/26910638/KSP.log)
[stderr.log](https://github.com/user-attachments/files/26910639/stderr.log)
[INFO.20260420-201341.39976.log](https://github.com/user-attachments/files/26910642/INFO.20260420-201341.39976.log)
[WARNING.20260420-201341.39976.log](https://github.com/user-attachments/files/26910641/WARNING.20260420-201341.39976.log)
[ERROR.20260420-201341.39976.log](https://github.com/user-attachments/files/26910640/ERROR.20260420-201341.39976.log)

pleroy:
The computation of the torque is done on the C++ side in `Part` and the test is moved to `part_test.cpp`.

I tested the save in #4166 as well as the ones in #2519 and #3697.

Fix #4166.
Fix #4531.


ksp_plugin/interface_part.cpp
```diff
@@ -1,27 +1,82 @@
...
+  InertiaTensor<PartPrincipalAxes> const inertia_tensor_in_principal_axes(
+      R3x3Matrix<MomentOfInertia>({moments_of_inertia.x, zero, zero},
+                                  {zero, moments_of_inertia.y, zero},
+                                  {zero, zero, moments_of_inertia.z}));
+
+  Rotation<PartPrincipalAxes, RigidPart> const principal_axes_to_part(
+      FromWXYZ(principal_axes_rotation));
+
+  return principal_axes_to_part(inertia_tensor_in_principal_axes);
+}
```
eggrobin: Use this in `principia__InsertOrKeepLoadedPart` too (consider moving it to `interface_body` with the other conversions as needed).

ksp_plugin/part.hpp
```diff
@@ -167,6 +169,16 @@ class Part final {
   // Returns "part name (part ID)".
   std::string ShortDebugString() const;
 
+  // Computes the torque that implements the bogus PhysX exponential decay of
+  // the angular velocity.  It is here for ease of testing.  It cannot be a
+  // method of `Part` because this is computed before the `Part` has been
+  // constructed/updated.
```
eggrobin:
```suggestion
  // the angular velocity.  It is here for ease of testing.  It must be static
  // because this is computed before the `Part` has been constructed/updated.
```
This is not C#, Java, or Python we don’t call member functions methods (and those languages call their equivalents of static member functions methods too anyway).

eggrobin (approved):

ksp_plugin/part.hpp
pleroy: Done.

ksp_plugin/interface_part.cpp
pleroy: Done.

Issue #4555, opened 2026-05-04
Revise the multithreading of reanimation


pleroy:
`RecurringThread` is probably not the right abstraction.  It was designed for best-effort activities like prolongation, not for activities upon which clients depend for correctness.

* Reanimation is not really a periodic activity, if someone wants to reanimate the past, there is no reason to wait for 20 ms before starting.  If the point is to yield the processor for other activities, this could/should be done in the callback.
* The `RecurringThread` doesn't manage a queue (well, it does, but it's a 1-element queue) so different callers may step on each others' toes.  In particular, `Put` stomps on the previous data in the queue.
* The only virtue of the `RecurringThread` is that it can be cancelled/restarted because it uses `jthread`.
* Callers that need guarantees about their request being ultimately processed (because they block) need to announce themselves using the awkward `Client`/`Clientele` API.  This is only correct if all the places that may ask for a restart or call `Put` properly look at the `Clientele`.

We need some kind of a `ThreadPool` which is based on `jthread`, and whose queue is time-ordered, probably with handles to allow individual cancellation of requests.

Issue #4490, opened 2026-03-30
Absurdly large per-vessel message size


periodically-makes-puns:
[INFO.20260330-153303.51640.log](https://github.com/user-attachments/files/26359686/INFO.20260330-153303.51640.log)
[WARNING.20260330-153303.51640.log](https://github.com/user-attachments/files/26359684/WARNING.20260330-153303.51640.log)
[ERROR.20260330-153303.51640.log](https://github.com/user-attachments/files/26359685/ERROR.20260330-153303.51640.log)

Link to the relevant savefile: https://drive.google.com/file/d/1YkF2KOLpE91D1q-Ej4oeC8WZDTyHRftp/view?usp=drive_link

[KSP.zip](https://github.com/user-attachments/files/26359698/KSP.zip)

No actions taken except loading in, and immediately quitting to main menu. Unsure how to reproduce from a fresh save, since I pretty much assumed this was normal behaviour for Principia + Skopos. 

In the provided INFO glog, each vessel has an absurd message->ByteSize() and message->SpaceUsed(). 

I know for certain I messed up the update procedure (replaced Principia folder instead of deleting/re-adding) from Лефшец to Lebesgue, which may have played into this? However vessel serialisation appears normal even after said update:
[INFO.20260320-094520.81472.log](https://github.com/user-attachments/files/26359930/INFO.20260320-094520.81472.log)
[WARNING.20260320-094520.81472.log](https://github.com/user-attachments/files/26359931/WARNING.20260320-094520.81472.log)
[ERROR.20260320-094520.81472.log](https://github.com/user-attachments/files/26359932/ERROR.20260320-094520.81472.log)
[no corresponding save or KSP.log since I wiped my saves to save space, unfortunately]

periodically-makes-puns:
I've been able to replicate what I think is the underlying cause on a fresh save, with Legendre properly installed this time. 

The save file size appears to increase by 1.3MB or thereabouts every time I perform a stationkeeping pass on my Skopos satellites. One such stationkeeping pass is attached in the logs below. I believe that the issue lies in the high number of checkpoints made per vessel during such a stationkeeping pass.

During a stationkeeping pass I perform the following procedure on each vessel:
- Set MechJeb to point the probe in SUN -> UP/DOWN direction as needed for the solar panels to realign.
- Open the orbit analyser and set it to 187d. Manually adjust the orbit using translational RCS until variance on the longitudes of equatorial crossings is acceptably low (<3-4 degrees).

[...]

PR #4556, opened 2026-05-04
Improve the handling of multithreading in reanimation

pleroy:
1. `Vessel` was missing locking in `RequestReanimation` (I have to believe that it was lost at some point, there was a block where locking should have happened).
2. There was a race where `last_desired_t_min_` was read without holding the lock in `Ephemeris`.
3. The check that reanimation has already reached the desired time was missing from `Ephemeris` (probably because of a misguided attempt at making some of this code lock-free).
4. That check should happen before restarting the reanimator anyway: no point in restarting if we have nothing to do.

This makes `Vessel` and `Ephemeris` more similar, but the entire multithreading scheme is rather ugly, see #4555.

Related to #4490.

### DIFF
--- a/ksp_plugin/vessel.cpp
+++ b/ksp_plugin/vessel.cpp
@@ -395,7 +395,10 @@ void Vessel::RequestReanimation(Instant const& desired_t_min,
   // No locking here because vessel reanimation is only invoked from the main
   // thread.
   bool must_restart;
+  Instant allowable_desired_t_min;
   {
+    absl::MutexLock l(&lock_);
+
     // If the reanimator is asked to do significantly less work (in terms of
     // checkpoints to reanimate) than it is currently doing, interrupt it.  Note
     // that this is fundamentally racy: for instance the reanimator may not have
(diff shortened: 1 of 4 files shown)


ksp_plugin/vessel.hpp
```diff
@@ -445,7 +445,7 @@ class Vessel {
   Clientele<Instant> reanimator_clientele_;
 
   // Parameter passed to the last call to `RequestReanimation`, if any.
-  std::optional<Instant> last_desired_t_min_;
+  std::optional<Instant> last_desired_t_min_ ABSL_GUARDED_BY(lock_);
```
eggrobin, line 448: How is this accessed from multiple threads?

ksp_plugin/vessel.cpp
```diff
@@ -395,15 +395,18 @@ void Vessel::RequestReanimation(Instant const& desired_t_min,
   // No locking here because vessel reanimation is only invoked from the main
   // thread.
   bool must_restart;
+  Instant allowable_desired_t_min;
   {
+    absl::MutexLock l(&lock_);
```
eggrobin: I think the comment explains why there was no locking. Now we are doing reanimation from a thread pool, but I don’t think any vessel can be reanimated concurrently from multiple threads (the main thread is waiting for the pool to be done, and vessel is reanimated by only one thread in the pool).

pleroy: I think it's brittle to depend on what we know about the callers today.  The contract of reanimation is that it should be called from any place that needs to know the past, without assuming that the clients will synchronize.  We can certainly reanimate the `Ephemeris` from multiple threads, it would be surprising if the `Vessel` worked differently.

If the lock is not contended it is going to cost us a few cycles.  If it ever is contended, well, we'll be happy to have a lock.

I am removing this comment.

PR #4557, opened 2026-05-05
A manual test of timezones

pleroy:


base/file_log_sink_test.cpp
```diff
@@ -44,7 +44,8 @@ class FileLogSinkTest : public ::testing::Test {
     }
   }
 
-  static std::filesystem::path UniqueLogFile(absl::LogSeverity const severity) {
+  static std::filesystem::path UniqueLogFile(absl::LogSeverity const severity,
+                                             bool check_first_line = true) {
```
eggrobin: It’s not the first line, it’s the filename. `check_file_name`?

pleroy: Silly me.  Done.

PR #4578, opened 2026-05-22
Change SinCos.tex to remove characters that cause TeXworks 0.6.11 to die

pleroy:
This also regenerates the PDFs.

Diffs: [diff.txt](https://github.com/user-attachments/files/28149664/diff.txt)


documentation/Sin Cos.tex
```diff
@@ -32,7 +32,7 @@
 \newcommand{\roundTowardZero}[1]{\doubleSquareBrackets{#1}_0}
 \newcommand{\roundTowardPositive}[1]{\doubleSquareBrackets{#1}_+}
 \newcommand{\roundTowardNegative}[1]{\doubleSquareBrackets{#1}_-}
-\newcommand{\roundAll}[1]{\doubleSquareBrackets{#1}_⋯}
+\newcommand{\roundAll}[1]{\doubleSquareBrackets{#1}_\textellipsis}
```
eggrobin: Wrong dots: this is  `\unicodecdots` in unicode-math (not sure how this differs from `\cdots`, but it certainly is not a baseline ellipsis).

pleroy: Done.

pleroy:
Downgrading TeXworks to 0.6.9 fixes the problem, without needing to mess with our source files.

PR #4593, opened 2026-06-05
C++ graphs

eggrobin:


astronomy/orbital_elements.hpp
```diff
@@ -70,6 +72,24 @@ class OrbitalElements {
   // variations, i.e., variations whose period is the orbital period.
   std::vector<ClassicalElements> const& mean_elements() const;
 
+  // Returns a graph of the given `element` over `mean_elements()`.
+  // `element` must not be `&ClassicalElements::time` nor
+  // `&ClassicalElements::mean_anomaly`.
+  template<typename T>
+  Graph<Instant, T> PlotTimeSeries(T ClassicalElements::* element,
+                                   int width,
```
pleroy: `std::int64_t` (many places in this PR).

astronomy/orbital_elements_body.hpp
```diff
@@ -87,11 +87,89 @@ absl::StatusOr<OrbitalElements> OrbitalElements::ForTrajectory(
...
+    RGB24 line_colour) const {
+  Interval<T> ordinate_range;
+  if constexpr (std::is_same_v<T, Length>) {
+    ordinate_range = {mean_periapsis_distance_interval().min,
+                      mean_apoapsis_distance_interval().max};
+  } else if constexpr (std::is_same_v<T, double>) {
+    CHECK_EQ(element, &ClassicalElements::eccentricity);
+    ordinate_range = mean_eccentricity_interval();
+  } else {
+    auto element_interval =
```
pleroy: `const`

Also, let's spell out the type, it's not obvious.

astronomy/orbital_elements_body.hpp
```diff
@@ -87,11 +87,89 @@ absl::StatusOr<OrbitalElements> OrbitalElements::ForTrajectory(
...
 
+inline Graph<double, double> OrbitalElements::PlotEccentricityVector(
+    int width,
+    int height,
+    RGBA32 background,
+    RGB24 axis_colour,
+    RGB24 line_colour) const {
+  double const aspect_ratio =
+      static_cast<double>(width) / static_cast<double>(height);
+  std::vector<std::pair<double, double>> eccentricity_vector;
```
pleroy, line 144: I presume that this `pair` represents 2-dimensional coordinates.  Could we have an `R2Element` class for this?  For one thing, this would give us a central place for optimizations (e.g., use SSE) if we want.

astronomy/orbital_elements_body.hpp
```diff
@@ -87,11 +87,89 @@ absl::StatusOr<OrbitalElements> OrbitalElements::ForTrajectory(
...
+  Interval<double> e_sin_ω_range;
+  for (ClassicalElements const& elements : mean_classical_elements_) {
+    auto const [sin_ω, cos_ω] = SinCos(elements.argument_of_periapsis);
+    auto const& e = elements.eccentricity;
+    eccentricity_vector.emplace_back(e * cos_ω, e * sin_ω);
+    e_cos_ω_range.Include(eccentricity_vector.back().first);
+    e_sin_ω_range.Include(eccentricity_vector.back().second);
+  }
+  if (e_cos_ω_range.measure() / width > e_sin_ω_range.measure() / height) {
+    double midpoint = e_sin_ω_range.midpoint();
```
pleroy: `const`

astronomy/orbital_elements_body.hpp
```diff
@@ -87,11 +87,89 @@ absl::StatusOr<OrbitalElements> OrbitalElements::ForTrajectory(
...
+    e_sin_ω_range.min = midpoint - e_cos_ω_range.measure() / aspect_ratio / 2;
+    e_sin_ω_range.max = midpoint + e_cos_ω_range.measure() / aspect_ratio / 2;
+  } else {
+    double midpoint = e_cos_ω_range.midpoint();
```
pleroy: `const`

astronomy/лидов_古在_test.cpp
```diff
@@ -178,6 +181,29 @@ TEST_F(Лидов古在Test, MercuryOrbiter) {
   EXPECT_THAT(elements.mean_argument_of_periapsis_interval().max,
               IsNear(129_(1) * Degree));
 
+  EXPECT_GOLDEN_GRAPH(
+      elements.PlotTimeSeries(&OrbitalElements::ClassicalElements::eccentricity,
+                              200,
```
pleroy: Comment the parameters (multiple places).

geometry/interval.hpp
```diff
@@ -31,6 +31,9 @@ struct Interval {
 
   // Extends this interval so that it contains x.
   void Include(T const& x);
+
+  // The intersection of `*this` with `other`, which is also an interval.
+  Interval IntersectedWith(Interval const& other) const;
```
pleroy: I don't like that this is a member function: it breaks the symmetry of the operation.  Let's have a self-standing function outside of the struct.

graphics/colours.hpp
```diff
@@ -0,0 +1,984 @@
...
+  std::uint8_t green;
+  std::uint8_t blue;
+};
+
+struct RGBA32 {
+  RGB24 colour;
+  std::uint8_t alpha;
+};
+
+constexpr RGBA32 Opaque(RGB24 colour) {
```
pleroy: `const` the parameter.

graphics/colours.hpp
```diff
@@ -0,0 +1,984 @@
...
+}
+
+namespace xkcd {
+// From https://xkcd.com/color/rgb.txt, with the following substitutions:
+// " " ↦ "_";
+// "'" ↦ "";
+// "/" ↦ "_slash_".
+// See also https://xkcd.com/color/rgb/.
+// License: https://creativecommons.org/publicdomain/zero/1.0/
+constexpr RGB24 cloudy_blue{0xac, 0xc2, 0xd9};
```
pleroy: I would alphabetize the declarations in this list.  It will make it easier to find the exact spelling when you have decided on a colour.

graphics/graph.hpp
```diff
@@ -0,0 +1,71 @@
...
+using namespace principia::base::_algebra;
+using namespace principia::geometry::_interval;
+using namespace principia::graphics::_colours;
+
+template<affine Abscissa, affine Ordinate>
+class Graph {
+ public:
+  Graph(int width,
+        int height,
+        Interval<Abscissa> x_range,
```
pleroy: Pass the intervals by `const&`.

graphics/graph.hpp
```diff
@@ -0,0 +1,71 @@
...
+namespace internal {
+
+using namespace principia::base::_algebra;
+using namespace principia::geometry::_interval;
+using namespace principia::graphics::_colours;
+
+template<affine Abscissa, affine Ordinate>
+class Graph {
+ public:
+  Graph(int width,
```
pleroy: Would it make sense for `width` and `height` to be template parameters?  I don't imagine them being dynamic.  If you did that you could use an `std::array` instead of an `std::vector` and that would be more efficient.

graphics/graph.hpp
```diff
@@ -0,0 +1,71 @@
...
+  void PlotHorizontalLine(Ordinate y, RGB24 colour);
+
+  std::vector<RGBA32> const& pixels() const;
+  int width() const;
+  int height() const;
+
+ private:
+  int AbscissaToPixel(Abscissa x) const;
+  Interval<Abscissa> PixelToAbscissa(int i) const;
+  int OrdinateToPixel(Ordinate y) const;
```
pleroy: These functions should be in `snake_case` I think.

graphics/graph_body.hpp
```diff
@@ -0,0 +1,124 @@
...
+    double f_max;
+    if (f_x_min <= f_x_max) {
+      f_min = f_x_min;
+      f_max = f_x_max;
+    } else {
+      f_min = f_x_max;
+      f_max = f_x_min;
+    }
+    for (int j = OrdinateToPixel(f_max); j <= OrdinateToPixel(f_min); ++j) {
+      SetPixel(i, j, {.colour = colour, .alpha = 255});
```
pleroy: Call `Opaque`.

graphics/graph_body.hpp
```diff
@@ -0,0 +1,124 @@
...
+  }
+}
+
+template<affine Abscissa, affine Ordinate>
+void Graph<Abscissa, Ordinate>::Plot(std::function<Ordinate(Abscissa)> const& f,
+                                     Interval<Abscissa> range,
+                                     RGB24 colour) {
+  for (int i = AbscissaToPixel(range.min); i <= AbscissaToPixel(range.max);
+       ++i) {
+    Interval pixel_range = PixelToAbscissa(i).IntersectedWith(range);
```
pleroy: I'd declare this variable outside of the loop.

graphics/graph_body.hpp
```diff
@@ -0,0 +1,124 @@
...
+    : x_range_(x_range),
+      y_range_(y_range),
+      width_(width),
+      height_(height),
+      pixels_(width * height, background) {}
+
+template<affine Abscissa, affine Ordinate>
+template<std::ranges::range Points>
+void Graph<Abscissa, Ordinate>::ListPointPlot(Points const& points,
+                                             RGB24 colour) {
```
pleroy: `const` the parameters, everywhere in this class.

graphics/graph_body.hpp
```diff
@@ -0,0 +1,124 @@
...
+  for (auto const& [x, y] : points) {
```
pleroy, line 30: So there is this implicit assumption (nowhere formalized) that a point decomposes as `[x, y]`.  We really need an `R2Element`.

graphics/graph_body.hpp
```diff
@@ -0,0 +1,124 @@
...
+    }
+  }
+}
+
+template<affine Abscissa, affine Ordinate>
+void Graph<Abscissa, Ordinate>::PlotVerticalLine(
+    Abscissa x,
+    RGB24 colour,
+    std::optional<Interval<Ordinate>> y_range) {
+  for (int j = y_range.has_value() ? OrdinateToPixel(y_range->max) : 0;
```
pleroy: I don't get why this is `max` and the next line is `min`.  Shouldn't it be the other way around?

graphics/graph_body.hpp
```diff
@@ -0,0 +1,124 @@
...
+       ++j) {
+    SetPixel(AbscissaToPixel(x), j, {.colour = colour, .alpha = 255});
+  }
+}
+
+template<affine Abscissa, affine Ordinate>
+void Graph<Abscissa, Ordinate>::PlotHorizontalLine(Ordinate y,
+                                                          RGB24 colour) {
+  for (int i = 0; i < width_; ++i) {
+    SetPixel(i, OrdinateToPixel(y), {.colour = colour, .alpha = 255});
```
pleroy: `Opaque`.

graphics/graph_body.hpp
```diff
@@ -0,0 +1,124 @@
...
+Interval<Abscissa> Graph<Abscissa, Ordinate>::PixelToAbscissa(int i) const {
+  return {i * x_range_.measure() / width_ + x_range_.min,
+          (i + 1) * x_range_.measure() / width_ + x_range_.min};
+}
+
+template<affine Abscissa, affine Ordinate>
+int Graph<Abscissa, Ordinate>::OrdinateToPixel(Ordinate y) const {
+  return static_cast<int>(height_ * (y_range_.max - y) /
+                          y_range_.measure());
+}
```
pleroy, line 117: Given that we are going to call these functions **all.the.time** and we don't care too much about precision, I would invest in constants such as `one_over_width`, etc. to avoid gazillions of divisions.

graphics/graph_body.hpp
```diff
@@ -0,0 +1,124 @@
...
+
+template<affine Abscissa, affine Ordinate>
+void Graph<Abscissa, Ordinate>::SetPixel(int const column,
+                                         int const row,
+                                         RGBA32 const pixel) {
+  if (row < 0 || column < 0 || row >= height_ || column >= width_) {
```
pleroy, line 123: This looks like it should be a `CHECK`, setting a pixel outside of the bitmap is likely to be a bug.

graphics/graph_body.hpp
```diff
@@ -0,0 +1,124 @@
+#pragma once
```
pleroy, line 1: Blank line after.

graphics/graph_test.cpp
```diff
@@ -0,0 +1,48 @@
...
+using namespace principia::geometry::_interval;
+using namespace principia::graphics::_colours;
+using namespace principia::numerics::_elementary_functions;
+using namespace principia::quantities::_quantities;
+using namespace principia::quantities::_si;
+
+class GraphTest : public ::testing::Test {};
+
+TEST_F(GraphTest, Sinusoid) {
+  Graph<Angle, double> graph(162,
```
pleroy: Comment the parameters.  Also, why 162?

testing_utilities/golden_graphs.hpp
```diff
@@ -0,0 +1,60 @@
...
+      (graph), FILESYSTEM_STRING(suffix), FILESYSTEM_STRING(__FILE__)))
+namespace principia {
+namespace testing_utilities {
+namespace _golden_graphs {
+namespace internal {
+
+using namespace principia::graphics::_graph;
+using ::testing::Eq;
+
+template<typename Abscissa, typename Ordinate, typename Character>
```
pleroy, line 32: Comment why we template on the character type, this is unusual.

testing_utilities/golden_graphs.hpp
```diff
@@ -0,0 +1,60 @@
...
+void ExpectGoldenGraph(Graph<Abscissa, Ordinate> const& graph,
+                       Character const* const suffix,
```
pleroy: The year is 2026, would `std::string_view` work?  Or maybe `std::filesystem::path` if these are file name elements.

testing_utilities/golden_graphs.hpp
```diff
@@ -0,0 +1,60 @@
...
+                              .replace_extension()
+                              .concat("_")
+                              .concat(suffix)
+                              .replace_extension(".png");
+  std::vector<std::uint8_t> golden;
+  std::ifstream in(image_path, std::ios::binary | std::ios::in);
+  while (in.good()) {
+    in.read(reinterpret_cast<char*>(&golden.emplace_back()), 1);
+    if (in.eof()) {
+      golden.pop_back();
```
pleroy: I find it a bit distasteful to emplace and then pop.  Could we read in a local `uint8_t` and only emplace if `!eof()`?

testing_utilities/golden_graphs.hpp
```diff
@@ -0,0 +1,60 @@
...
+      break;
+    }
+  }
+  std::vector<std::uint8_t> actual;
+  lodepng::encode(actual,
+                  reinterpret_cast<std::uint8_t const*>(graph.pixels().data()),
+                  graph.width(),
+                  graph.height());
+  EXPECT_THAT(actual, Eq(golden)) << image_path << " has changed";
```
pleroy, line 56: I believe that `ContainerEq` would make the message more explicit in case of failure.

install_deps.sh
```diff
@@ -9,7 +9,7 @@ echo "Required runtime dependencies: libc++1"
 mkdir -p deps
 pushd deps
 
-for repo in abseil-cpp benchmark config gipfeli googletest multiprecision protobuf re2 zfp core-math; do
+for repo in abseil-cpp benchmark config gipfeli googletest multiprecision protobuf re2 zfp core-math lodepng; do
```
pleroy: Let's alphabetize this list.

astronomy/orbital_elements_body.hpp
eggrobin: They need not be homogeneous. I suppose we could use DirectSum, but we don’t actually use the algebraic structure…

graphics/graph.hpp
eggrobin: They are obviously dynamic, we don’t know the user’s UI size at compile time.

(And it might even make sense to have resizable graphs to be able to show larger maps etc.…)

graphics/graph_body.hpp
eggrobin: Consted instead, and since its value depends on the loop variable it is not moving out.

graphics/graph_body.hpp
eggrobin: The point is that this can plot _anything_ that decomposes that way. It could plot an std::map.

We might want to pass various things at the call sites, but we really should _not_ require some kind of R2Element here.

graphics/graph_body.hpp
eggrobin: No, the rows are top-to-bottom.

testing_utilities/golden_graphs.hpp
eggrobin: I really do not want to see a dump of the PNG bytes.

testing_utilities/golden_graphs.hpp
eggrobin: No it would not, obviously I started here. We pass literals of whose type is platform dependent, and we cannot deduce the type of the character through the implicit conversion to `basic_string_view<T>`.

See https://github.com/mockingbirdnest/Principia/pull/4593/commits/7ec5c31fa034cb5d3833c4d8397982bd5e377c0d.

graphics/graph_test.cpp
eggrobin: 100 φ.

graphics/graph_body.hpp
eggrobin: No, it just means that you chose a range that does not include all of the data.

This check would fail with the real graphs that we draw today, for instance we have axes out of bounds in the eccentricity vector plots.

graphics/graph_body.hpp
pleroy: Ok but this needs a comment explaining that it works that way because `OrdinateToPixel` is a *decreasing* function of its argument.

graphics/graph_test.cpp
pleroy: Ok so please write `static_cast<std::int64_t>(std::ceil(100 φ))`.  Just say no to magic numbers.

I am actually wondering if this should somehow be the default: pass height or width and we'll calculate the other dimension based on the golden ratio.

testing_utilities/golden_graphs.hpp
pleroy: The left part of that commit is wrong, but that doesn't mean that it cannot be made to work, see https://godbolt.org/z/KjWxb7jW9.

testing_utilities/golden_graphs.hpp
pleroy: When you have an unexpected difference you'll want to see the bytes.  It's very possible that `ContainerEq` is not the right tool, but then we will probably end up writing our own matcher, because "has changed" is hardly better than the `ed` error message `?`.

astronomy/orbital_elements_body.hpp
pleroy: Ok, then could we have a struct `Point<Abscissa, Ordinate>` either directly in `graph.hpp` or (probably better) nested in class `Graph`.

I am going to strenuously object to using `pair` as an "abstraction".

graphics/graph_body.hpp
pleroy: I disagree.  I want to have some kind of dedicated struct for a point.  I am not going to let you pass a random pair extracted from a random map.

astronomy/orbital_elements_body.hpp
```diff
@@ -90,8 +90,8 @@ absl::StatusOr<OrbitalElements> OrbitalElements::ForTrajectory(
 template<typename T>
 Graph<Instant, T> OrbitalElements::PlotTimeSeries(
     T OrbitalElements::ClassicalElements::* element,
-    int width,
-    int height,
+    std::int64_t width,
```
pleroy: `const` the parameters please.  I should not have to ask this on every single function body.

geometry/interval.hpp
```diff
@@ -31,17 +31,19 @@ struct Interval {
...
   // Extends this interval so that it contains x.
   void Include(T const& x);
-
-  // The intersection of `*this` with `other`, which is also an interval.
-  Interval IntersectedWith(Interval const& other) const;
 };
 
+// The intersection of `left` and `right`, which is also an interval.
+template<typename T>
+Interval<T> Intersection(Interval<T> const& left, Interval<T> const& right);
```
pleroy, line 38: Could we also add `Union` while we are at it?  We try to make abstractions reasonable complete.

eggrobin: No, the union of two intervals is not an interval.

testing_utilities/golden_graphs.hpp
eggrobin: No, I don’t want to see the bytes, I want to see the diff on GitHub. In practice, this means I need to commit whatever just changed by running the test, which makes the test pass without having to look at the output, and then I (and the reviewer) can look at the changes to the image.

graphics/graph_body.hpp
eggrobin: As discussed, consider the eminently graphable
https://github.com/mockingbirdnest/Principia/blob/65cdb02837f422e103e0cf13fdb2338fc0e5a402/numerics/fast_fourier_transform.hpp#L69

graphics/graph_test.cpp
eggrobin: This seems useful for tests but not in the game (our graphs don’t have that shape in game).
For tests, maybe we should even have a default size.

testing_utilities/golden_graphs.hpp
eggrobin: Might as well use ""sv literals then.

eggrobin: Ah no, that requires a using.

testing_utilities/golden_graphs.hpp
pleroy: What I had in mind was this: imagine that you have minor numeric changes.  You may have pixels changing in subtle ways that may not even be visible on the image diff.  This will be especially likely once we do antialiasing and one pixel changes from dark grey to not-so-dark grey and the next one from not-so-dark grey to dark grey.

We may need a dedicated matcher to help us with this.  But we'll see how things work in practice.

PR #4607, opened 2026-06-16
Fix typo

Clayell:
Made these commits back in november and forgot to make the PR for them, apparently


pleroy:
The fix looks fine but then I am wondering why we don't have a localization file for Russian.

@eggrobin, do you know?  Should we add a file for Russian?

(Sorry, I pinged the wrong person.)

eggrobin:
We must have forgotten about it in #3133…

pleroy:
Ok, I am going to merge this PR and I'll add Russian separately.

PR #4630, opened 2026-08-01
Only rewrite generated files if they changed

eggrobin:
Hopefully this will cause the compiler to recompile less.


tools/generate_profiles.cpp
```diff
@@ -21,6 +21,36 @@ const char warning[] = "// Warning!  This file was generated by running a "
                        "// If you change it, the changes will be lost the next "
                        "time the generator is\n"
                        "// run.  You should change the generator instead.\n\n";
+
+class DiffingFileStream {
+ public:
+  explicit DiffingFileStream(std::filesystem::path path) : path_(path) {}
```
pleroy: `const&` or `move`.

PR #4652, opened 2026-08-17
Golden diffing: don’t use compact-summary, don’t quote paths

eggrobin:


principia_make.sh
```diff
@@ -40,18 +40,20 @@ set -e
...
-  #  2 files changed, 2 insertions(+), 22 deletions(-)
-  git diff --compact-summary HEAD |
-      awk '/\.png/ { if ($2 == "(gone)") { print("(deleted) " $1) } else { system("sha1sum -b " $1) } }' \
+  # git diff --name-status --no-renames outputs something like
+  # A       path/to/new.file
+  # D       path/to/deleted.file
+  # M       path/to/modified.file
+  git config core.quotePath false
+  git diff --name-status --no-renames HEAD |
+      awk '/\.png/ { if ($1 == "A") { print("(deleted) " $2) } else { system("shasum -b " $2) } }' \
```
pleroy, line 51: Should be `"D"` I think.

principia_make.sh
```diff
@@ -40,18 +40,20 @@ set -e
...
       > golden_hashes.txt
-  HASH=$(sha1sum golden_hashes.txt | awk '{print($1)}')
+  cat golden_hashes.txt
```
pleroy, line 53: Remove?

eggrobin: No, I would like to be able to see what is going on when looking at the logs.

PR #4659, opened 2026-08-18
Clean up the principia_make.sh script

pleroy:
The changes attempt to conform to https://google.github.io/styleguide/shellguide.html.

### DIFF
--- a/principia_make.sh
+++ b/principia_make.sh
@@ -1,6 +1,6 @@
 #! /bin/bash
 
-readonly NEW_EXTENSION='.new'
+readonly SQUIRREL_EXTENSION='.squirrel'
 
 if [[ "${PRINCIPIA_PLATFORM?}" != "x64" &&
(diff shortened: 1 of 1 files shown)


principia_make.sh
```diff
@@ -1,126 +1,143 @@
...
+    echo "Branch ${branch_name} already exists."
     echo "Merge the following PR to update these goldens:"
-    gh pr list --head ${BRANCH_NAME}
+    gh pr list --head ${branch_name}
   fi
 fi
 
-if [[ "${MAKE_RESULT}" != 0 ]]; then
-  exit "${MAKE_RESULT}"
+if (( "${make_exit_status}" != 0 )); then
```
eggrobin:
```suggestion
if (( ${make_exit_status} != 0 )); then
```

eggrobin (approved):

principia_make.sh
pleroy: Done (also removing parameter expansion).

Issue #4624, opened 2026-07-28
Add an ability to draw only the parts of the flight plan we're interested in


AntonKuzin:
As it's the drawing that causes the most lag, it seems useful to have the ability to not render the whole trajectory(especially for the very long flight plans).
I see 2 ways to do it:

1. Use the global rendering window - something like "rendering window offset" - the offset from the flight plan start, and "rendering window duration".
2. Do the same per manoeuvre. Usually we do manoeuvres to get somewhere, and we probably only interested in our trajectory near the target. Doing it that way would make multiple fly-by plans much more manageable.

To keep the new feature inline with the existing behavior, the default offset and duration values could be chosen so the whole trajectory gets rendered. Maybe, an override checkbox would be needed to not to mess up the user defined values.

pleroy:
Discussed with @eggrobin today:
1. Add two text boxes + sliders on top of the plotting frame selection window to select the beginning and end of the time range to plot.
2. The time range is expressed in mission elapsed time, because managing absolute dates is messy (we don't want to figure out the duration of the year of whatever is replacing Kerbin).
3. The boxes have buttons `-∞` and `+∞`, respectively, to reset to displaying the complete trajectories.
4. These settings are persisted by frame (it probably makes sense to show different time ranges for different frames) and by vessel.
5. They affect all the trajectories: history and psychohistory, prediction, and flight plan.
6. At some point we might look into heuristics to fill these boxes with "nice" defaults, but anyway the user must be able to edit the defaults.

eggrobin:
> They affect all the trajectories: history and psychohistory, prediction, and flight plan

And, importantly, the trajectories of celestials as well.

AntonKuzin:
I wanna note that the beginning and the end of a rendering window are less convenient to use than the offset from the start of the mission and the window duration. With your approach one has to adjust 2 sliders to move the window, but with mine only adjusting the offset would do.

eggrobin:
It’s not completely clear that you would want the window to remain the same size when you move it (timescales change massively depending on which « interesting thing » you are looking at), but even if you need to adjust the size of the window, it is probably more convenient to have begin & size than begin & end, otherwise when moving a long way forward you would either move the end first, and end up having to move the end so far that you would run out of points and not see where the end is, or move the beginning first and end up with a segment of size zero (at which point you wouldn’t see anything).

eggrobin:
[...]

PR #4661, opened 2026-08-19
Improve rendering of the manœuvre markers

pleroy:
Each time we wanted to display the manœuvre markers, we would transform the *entire* flight plan from `Barycentric` to the plotting frame, and then use only the first point of each burn.  This was very inefficient as the coordinate transforms can be fairly expensive.

I am changing the code so that only the first point of each burn is transformed.  On my (lame) laptop, rendering a 100-d6 flight plan with a single manœuvre goes from 11 fps to 16 fps with this PR.  YMMV.

With this change, some methods in `Renderer` becomes unused and I am removing them.  I am also killing entirely the so-called plugin integration test, which was the last client of these methods, and, since it had not been properly maintained, was exercising the logic as it was about 9 years ago.

Noticed when working on #4624.


serialization/journal.proto
```diff
@@ -1175,6 +1175,27 @@ message FlightPlanRenderedClosestApproaches {
   optional Out out = 2;
 }
 
+message FlightPlanRenderedManoeuvre {
```
eggrobin: Explain what this function does.

ksp_plugin/interface_flight_plan.cpp
```diff
@@ -621,6 +624,49 @@ void __cdecl principia__FlightPlanRenderedClosestApproaches(
...
+        manœuvre.end(),
+        FromXYZ<Position<World>>(sun_world_position),
+        plugin->PlanetariumRotation());
+
+    if (!rendered_manœuvre.empty() &&
+        rendered_manœuvre.begin()->first != segment->front().time) {
+      // TODO(egg): this is ugly; we should centralize rendering.
+      // If this is a burn and we cannot render the beginning of the burn, we
+      // render none of it, otherwise we try to render the Frenet trihedron at
+      // the start and we fail.
```
eggrobin: This comment no longer makes sense when we render at most one point (which I think is the beginning?).

serialization/journal.proto
eggrobin: And maybe call it ManoeuvreInitialState or something, it does not render the entire manœuvre.

pleroy: > Explain what this function does.

Done.  Since we are talking about comments, I'd love to know what `FlightPlanGetManoeuvreInitialPlottedVelocity` does and how it differs from this new function.  (I compared the code, it's plotting instead of `World`.)

> And maybe call it ManoeuvreInitialState or something, it does not render the entire manœuvre.

`FlightPlanRenderedManoeuvreInitialDegreesOfFreedom`

ksp_plugin/interface_flight_plan.cpp
pleroy: ~~The "if" needs to go, but the rest of the comment is still valid I think.~~

pleroy: Nevermind.

eggrobin (approved):

PR #4663, opened 2026-08-21
A new class, DiscreteTrajectoryView, to represent a read-only view of a range of a discrete trajectory

pleroy:
#4624.


physics/discrete_trajectory_view_test.cpp
eggrobin, line 1: Test that std::ranges::begin(DiscreteTrajectoryView(&some_trajectory)) compiles; I think it won’t. If it doesn’t, make it compile using enable_borrowed_range.

physics/discrete_trajectory_view_test.cpp
```diff
@@ -0,0 +1,262 @@
...
+                              0 * Metre / Second});
+    return MakeTrajectory(t0_, DegreesOfFreedom<World>(World::origin, v1));
+  }
+
+  Instant const t0_;
+};
+
+TEST_F(DiscreteTrajectoryViewTest, ViewOfEntireNonemptyTrajectory) {
+  auto const trajectory = MakeTrajectory();
+  DiscreteTrajectoryView<World> view(&trajectory);
```
eggrobin: Does deduction work (auto view = DiscreteTrajectory(&trajectory))? If so, test it at least once for each constructor for which it works. If not, consider adding deduction guides.

Does the view need to be non-const? Same question throughout.

physics/discrete_trajectory_view_test.cpp
pleroy: Done.

physics/discrete_trajectory_view_test.cpp
pleroy: I don't imagine how `auto` would work.  I have added deduction guidelines so that the frame can be omitted.

Also added `const` to all the tests.

Changing the base class in the same manner.

physics/discrete_trajectory_view.hpp
```diff
@@ -0,0 +1,137 @@
...
+using internal::DiscreteTrajectoryView;
+
+}  // namespace _discrete_trajectory_view
+}  // namespace physics
+}  // namespace principia
+
+namespace std {
+namespace ranges {
+template<typename Frame>
+inline constexpr bool enable_borrowed_range<
```
eggrobin, line 131: I think the `inline` is redundant (but we do that in many other places—though not all——, so maybe we should clean it up separately).

eggrobin (approved):

PR #4674, opened 2026-08-27
Use trajectory views in the interface, the Plugin and the Renderer

pleroy:
#4624.

### DIFF
--- a/ksp_plugin_test/renderer_test.cpp
+++ b/ksp_plugin_test/renderer_test.cpp
@@ -15,6 +15,7 @@
 #include "ksp_plugin_test/mock_vessel.hpp"
 #include "physics/degrees_of_freedom.hpp"
 #include "physics/discrete_trajectory.hpp"
+#include "physics/discrete_trajectory_view.hpp"
 #include "physics/mock_continuous_trajectory.hpp"
 #include "physics/mock_ephemeris.hpp"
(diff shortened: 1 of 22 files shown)


ksp_plugin_test/renderer_test.cpp
```diff
@@ -124,8 +126,7 @@ TEST_F(RendererTest, RenderBarycentricTrajectoryInPlottingWithoutTargetVessel) {
 
   auto const rendered_trajectory =
       renderer_.RenderBarycentricTrajectoryInPlotting(
-          trajectory_to_render.begin(),
-          trajectory_to_render.end());
+          DiscreteTrajectoryView(&trajectory_to_render));
```
eggrobin, line 129: I wonder if we shouldn’t make this constructor implicit, like string_view.

eggrobin (approved):

ksp_plugin_test/renderer_test.cpp
pleroy: I'd rather make it explicit that we are building a view, in particular because of lifetime reasons.  For instance, views should be captured by copy, not by reference.  Trajectories can be capture by reference.  See [this code in this PR](https://github.com/mockingbirdnest/Principia/pull/4674/changes#diff-b6d0a2ef60e69a4d4bcc11210339bfbab2990e0e7e7e0094e49fab973adc4307R47) for example.

PR #4676, opened 2026-08-29
Persistence of the vessel payloads

pleroy:
Somehow I forgot this when writing the C++ code for payloads.

#4624.

### DIFF
--- a/ksp_plugin/vessel.cpp
+++ b/ksp_plugin/vessel.cpp
@@ -713,6 +713,17 @@ void Vessel::WriteToMessage(not_null<serialization::Vessel*> const message,
       LOG(FATAL) << "Unexpected flight plan variant " << flight_plan.index();
     }
   }
+
+  // This a hash map so the order is unspecified.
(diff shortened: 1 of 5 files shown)


ksp_plugin/vessel.cpp
```diff
@@ -713,6 +713,17 @@ void Vessel::WriteToMessage(not_null<serialization::Vessel*> const message,
       LOG(FATAL) << "Unexpected flight plan variant " << flight_plan.index();
     }
   }
+
+  // This a hash map so the order is unspecified.
```
eggrobin, line 717: Won’t that be a problem for serialization round-trip testing?

pleroy: It doesn't fail the serialization round-trip test if you don't test serialization round-trip 😀

I am adding a test, which fails nondeterministically if you do the naïve comparison, and works (100 runs) if you ignore the order of the payloads.  (Apologies for the convoluted use of gMock, but the nice protobuf matchers are not publicly available because of course.)

eggrobin (approved):

PR #4678, opened 2026-08-31
Modernize and clarify LunarOrbitTest

eggrobin:
Also:
* make EXPECT_GOLDEN_GRAPH work in parameterized tests;
* make ListPointPlot work with anything that works with structured bindings, not just tuple-like types;
* fix a bug in PlotEccentricityVector when the locus of the eccentricity vector is taller than it is wide.


astronomy/lunar_orbit_test.cpp
```diff
@@ -77,58 +85,171 @@ using namespace principia::physics::_massless_body;
...
+  Length const y₀;
+  Length const z₀ = 0 * Metre;
+  Speed const  u₀;
+  Speed const v₀;
+  Speed const w₀;
+  Length const a₀ = +1.861791339407e+03 * Kilo(Metre);
+  double const e₀ = +2.110475283361e-02;
+  Angle const i₀ = +9.298309294740e+01 * Degree;
+  Angle const ω₀ = -7.839337618501e+01 * Degree;
+  Angle const Ω₀ = -1.589469097527e+02 * Degree;
```
pleroy: Why are these fields defaulted, when the functions below give them explicit values?

astronomy/lunar_orbit_test.cpp
```diff
@@ -77,58 +85,171 @@ using namespace principia::physics::_massless_body;
...
+  // The duration for the above expectation.
+  Time long_term;
+
+  // The expected status of the integration over `long_term`.
+  absl::StatusCode long_term_status = absl::StatusCode::kOk;
+
+  // Override the range of the eccentricity vector plot for the first month.
+  // This is used for orbit A, where the osculating eccentricity vector varies
+  // much more than the mean one.
+  std::optional<double> first_month_e_cos_ω_plot_range = std::nullopt;
```
pleroy: Remove the default value.

astronomy/lunar_orbit_test.cpp
```diff
@@ -77,58 +85,171 @@ using namespace principia::physics::_massless_body;
...
-struct GeopotentialTruncation {
+struct RL06Orbit {
+  // Initial conditions and elements from table 2 of [RL06].
+  int const revolutions_per_month;
+  Length const x₀;
+  Length const y₀;
+  Length const z₀ = 0 * Metre;
+  Speed const  u₀;
+  Speed const v₀;
+  Speed const w₀;
```
pleroy, line 114: I find it strange that these are not `R3Element`s.  That would make the initial dof near line 459 more natural.  We very seldom write `{x, y, z}`.

astronomy/lunar_orbit_test.cpp
```diff
@@ -336,15 +474,19 @@ TEST_P(LunarOrbitTest, NearCircularRepeatGroundTrackOrbit) {
     // error on our LU with respect to the one in the paper: the semimajor axis
     // has the same value in LU.
     EXPECT_THAT(*initial_osculating.semimajor_axis,
-                RelativeErrorFrom(a0, IsNear(9.0e-4_(1))));
+                RelativeErrorFrom(orbit.a₀, IsNear(9.0e-4_(1))));
```
pleroy: Why is this `9.0e-4` part of orbit?

astronomy/lunar_orbit_test.cpp
```diff
@@ -357,188 +499,168 @@ TEST_P(LunarOrbitTest, NearCircularRepeatGroundTrackOrbit) {
...
 
   // To find the nodes, we need to convert the trajectory to a reference frame
   // whose xy plane is the Moon's equator.
-  DiscreteTrajectory<LunarSurface> surface_trajectory;
+  DiscreteTrajectory<Selenocentric> first_month_selenocentric_trajectory;
+  DiscreteTrajectory<Selenocentric> selenocentric_trajectory;
   for (auto const& [time, degrees_of_freedom] : trajectory) {
-    EXPECT_OK(surface_trajectory.Append(
-        time, lunar_frame_.ToThisFrameAtTime(time)(degrees_of_freedom)));
+    if (time - J2000 <= month) {
```
pleroy: `time <= J2000 + month` makes the upper bound more explicit.

astronomy/lunar_orbit_test.cpp
```diff
@@ -357,188 +499,168 @@ TEST_P(LunarOrbitTest, NearCircularRepeatGroundTrackOrbit) {
...
+      auto const [sin_ω, cos_ω] = SinCos(ω);
+      e_sin_ω = e * sin_ω;
+      e_cos_ω = e * cos_ω;
+    }
+    double e_cos_ω;
+    double e_sin_ω;
+  };
+
+  std::vector<EccentricityVector> first_month_osculating_eccentricity_vector;
+  for (Instant t = J2000; t <= J2000 + month; t += month / 50'000) {
```
pleroy: Should we invest in a constant for 50000?  How was that number chosen?

astronomy/lunar_orbit_test.cpp
```diff
@@ -357,188 +499,168 @@ TEST_P(LunarOrbitTest, NearCircularRepeatGroundTrackOrbit) {
...
-                  surface_trajectory.EvaluatePosition(t) - LunarSurface::origin,
-                  ExpressIn(Metre));
+            *moon_,
+            satellite_,
+            selenocentric_trajectory.EvaluateDegreesOfFreedom(t) -
+                selenocentre_,
+            t).elements_at_epoch();
+
+    Angle const& ω = *elements.argument_of_periapsis;
+    double const& e = *elements.eccentricity;
```
pleroy, line 539: I don't think we normally take references to `double`.

astronomy/lunar_orbit_test.cpp
```diff
@@ -357,188 +499,168 @@ TEST_P(LunarOrbitTest, NearCircularRepeatGroundTrackOrbit) {
...
+    EXPECT_THAT(first_month_e_sin_ω.min,
+                IsNear(GetParam().first_month_descending_nodes.min_e_sin_ω));
+    EXPECT_THAT(first_month_e_sin_ω.max,
+                IsNear(GetParam().first_month_descending_nodes.max_e_sin_ω));
   }
 
+  auto const long_term_mean_elements = OrbitalElements::ForTrajectory(
+      selenocentric_trajectory, *moon_, MasslessBody{});
+  ASSERT_OK(long_term_mean_elements);
+  auto long_term_eccentricity_vector_graph =
```
pleroy, line 616: Could we build the graph right before the place where we expect it?

graphics/graph.hpp
```diff
@@ -18,10 +18,20 @@ using namespace principia::geometry::_interval;
...
 
 template<typename Point, typename Abscissa, typename Ordinate>
-concept graph_point =
-    (std::tuple_size_v<Point> == 2) &&
-    std::convertible_to<std::tuple_element_t<0, Point>, Abscissa> &&
-    std::convertible_to<std::tuple_element_t<1, Point>, Ordinate>;
+concept graph_point = requires(Point const& p) {
+  {
+    [&p] {
+      auto const [x, y] = p;
```
pleroy: `_` instead of `y`.

graphics/graph.hpp
```diff
@@ -18,10 +18,20 @@ using namespace principia::geometry::_interval;
...
+      return x;
+    }()
+  } -> std::convertible_to<Abscissa>;
+  {
+    [&p] {
+      auto const [x, y] = p;
```
pleroy: Same.

astronomy/lunar_orbit_test.cpp
eggrobin: It is not, and there is a comment explaining why.

astronomy/lunar_orbit_test.cpp
eggrobin: No, because we add things to it.

astronomy/lunar_orbit_test.cpp
eggrobin: This is to match the notation from the paper.

astronomy/lunar_orbit_test.cpp
pleroy: My question was missing the word "not", sorry.

I find it very strange that, to figure out what this number is, I have to read a comment that is like 300 lines before.  I'd suggest to have a (static, constexpr) constant for this number and then you can attach the comment to that constant.

eggrobin: Note that there is also a comment here, two lines above:
```
    // The relative error on the semimajor axis is the same as the relative
    // error on our LU with respect to the one in the paper: the semimajor axis
    // has the same value in LU.
```

astronomy/lunar_orbit_test.cpp
eggrobin: I don’t know, it was chosen 8 years ago. Maybe we should just iterate over the DiscreteTrajectory (but that will change the goldens; they will have ~5 times more points).

astronomy/lunar_orbit_test.cpp
eggrobin: We do when we mean to rename things (there are already multiple instances of `double const& e` specifically renaming an eccentricity).

pleroy (approved):

